PR #10, opened 2020-07-02
adding discussed schema changes for permissions and renaming joining table

helrick:


flask/app/models.py
```diff
@@ -21,6 +22,18 @@ def check_password(self, password):
 
     def get_id(self):
         return self.user_id
+    
+    def make_admin(self):
+        self.is_admin = True
+
+    def revoke_admin(self):
+        self.is_admin = False
```
kevinlul, line 30: Do these methods serve any purpose or will they be expanded to do something?

flask/app/models.py
```diff
@@ -140,11 +153,22 @@ class DatasetReadType(Enum):
...
+    'users_groups', db.Model.metadata,
+    db.Column('user_id', db.Integer, db.ForeignKey('user.user_id')),
+    db.Column('group_id', db.Integer, db.ForeignKey('group.group_id'))
+)
+
+groups_datasets_table = db.Table(
+    'groups_datasets', db.Model.metadata,
+    db.Column('group_id', db.Integer, db.ForeignKey('group.group_id')),
+    db.Column('dataset_id', db.Integer, db.ForeignKey('dataset.dataset_id'))
+)
```
kevinlul, line 171: Move block above `dataset_analyses_table`.

flask/app/models.py
helrick: I figured we'd need the ability to edit someone's admin status. Now that I'm looking through though, this would probably be better suited for the routes.py file?

kevinlul: If these don't add additional functionality related to the model like `set_password` does then they don't have a purpose.

PR #12, opened 2020-07-07
Analysis material table

dennis-kao:
- Changed the Table in the Analysis page over to material-table
- Added forms and dialogs to run analysis, and view analysis


react/src/pages/analysis/Analysis.tsx
```diff
@@ -87,103 +89,93 @@ function createAnalysis(
...
+                title={`Stop Analysis ${(activeRow as AnalysisRun).analysisID}?`}
+                message={`Do you really want to stop the analysis of samples ${(activeRow as AnalysisRun).participants.join(', ')}? Stopping an analysis will delete all intermediate files and progress. Input files will remain untouched.`}
+                onClose={() => { setCancel(false) }}
+            />}
+            {(activeRow as AnalysisRun) &&
+            <AnalysisInfoDialog
+                open={detail}
+                analysis={activeRow as AnalysisRun}
+                onClose={() => setDetail(false)}
+            />}
```
kevinlul, line 133: The type casts in the above block are unnecessary.

react/src/pages/analysis/CancelAnalysisDialog.tsx
```diff
@@ -0,0 +1,53 @@
...
+        open={open}
+        TransitionComponent={Transition}
+        keepMounted
+        onClose={onClose}
+        aria-labelledby="alert-dialog-slide-title"
+        aria-describedby="alert-dialog-slide-description"
+      >
+        <DialogTitle id="alert-dialog-slide-title">{title}</DialogTitle>
+        <DialogContent>
+          <DialogContentText id="alert-dialog-slide-description">
```
kevinlul, line 38: Since this component is used multiple times but HTML IDs are meant to be unique, it would be best to pass the prefixes for these for WAI-ARIA as a prop.

react/src/pages/analysis/ChipStrip.tsx
```diff
@@ -0,0 +1,33 @@
...
+      flexWrap: 'wrap',
+      '& > *': {
+        margin: theme.spacing(0.5),
+      },
+    },
+  }),
+);
+
+export default function ChipStrip({labels, color} : ChipStripProps) {
+    const chipStyle = makeChipStyle();
```
kevinlul, line 24: I believe we've been doing `useStyles` for every other component?

react/src/pages/participants/AnalysisRunnerDialog.tsx
```diff
@@ -0,0 +1,125 @@
...
+            TransitionComponent={Transition}
+            keepMounted
+            onClose={onClose}
+            aria-labelledby="alert-dialog-slide-title"
+            aria-describedby="alert-dialog-slide-description"
+            maxWidth='md' 
+            fullWidth={true}
+        >
+            <DialogTitle id="alert-dialog-slide-title">
+                {"Run analysis"}
```
kevinlul, line 61: Text nodes can just be written out inline, no need for `{""}`. This happens multiple times in this file.

react/src/pages/participants/AnalysisRunnerDialog.tsx
```diff
@@ -0,0 +1,125 @@
...
+            open={open}
+            TransitionComponent={Transition}
+            keepMounted
+            onClose={onClose}
+            aria-labelledby="alert-dialog-slide-title"
+            aria-describedby="alert-dialog-slide-description"
+            maxWidth='md' 
+            fullWidth={true}
+        >
+            <DialogTitle id="alert-dialog-slide-title">
```
kevinlul, line 60: This ID is the same as one used in the other dialog component.

react/src/pages/analysis/Analysis.tsx
```diff
@@ -65,7 +60,14 @@ const useStyles = makeStyles(theme => ({
 
 }));
 
-interface AnalysisRun {
+export enum PipelineStatus {
+    PENDING = "Pending",
+    RUNNING = "Running",
+    COMPLETED = "Completed",
+    ERROR = "Error",
+}
```
kevinlul, line 68: This is a good change so this can get type-checked too instead of just being a string.

react/src/pages/analysis/AnalysisInfoDialog.tsx
```diff
@@ -0,0 +1,96 @@
...
+      ) : null}
+    </MuiDialogTitle>
+  );
+});
+
+const DialogContent = withStyles((theme: Theme) => ({
+  root: {
+    padding: theme.spacing(2),
+  },
+}))(MuiDialogContent);
```
kevinlul, line 57: `withStyles` is actually really cool, especially if we have the free effort at the end to do changeable themes!

react/src/pages/analysis/DatasetTable.tsx
```diff
@@ -0,0 +1,58 @@
...
+  createData('3000_AA921', "RES", "AF983D_3000_AA921_R1.fq.gz, AF983D_3000_AA921_R2.fq.gz,", "102400 + 102400", "2020-10-01"),
+  createData('3000_AA922', "RES", "AF983D_3000_AA922_R1.fq.gz, AF983D_3000_AA922_R2.fq.gz,", "102400 + 102400", "2020-10-01"),
+];
+
+export default function DatasetTable() {
+  const classes = useStyles();
+
+  return (
+    <TableContainer component={Paper}>
+      <Table className={classes.table} aria-label="simple table">
```
kevinlul, line 32: Contents of the `aria-label` are unhelpful.

react/src/pages/analysis/DatasetTable.tsx
```diff
@@ -0,0 +1,58 @@
...
+          <TableRow>
+            <TableCell>SampleID</TableCell>
+            <TableCell align="right">Dataset Type</TableCell>
+            <TableCell align="right">Input files</TableCell>
+            <TableCell align="right">File size (mb)</TableCell>
+            <TableCell align="right">Date uploaded</TableCell>
+          </TableRow>
+        </TableHead>
+        <TableBody>
+          {rows.map((row) => (
```
kevinlul, line 43: Unnecessary parentheses around the argument to this arrow function.

react/src/pages/participants/ParticipantsTable.tsx
```diff
@@ -45,69 +56,77 @@ const rows = [
 
 export default function ParticipantsTable({ display }: ParticipantsTableProps) {
     const classes = useStyles();
+    const [showRunner, setRunner] = React.useState(false);
+    const [activeParticipants, setActiveParticipants] = React.useState<Participant[]>([]);
```
kevinlul, line 60: Every other file imports `useState` via destructured import instead of `React.useState` so it would be best to stay consistent.

react/src/pages/participants/ParticipantsTable.tsx
```diff
@@ -45,69 +56,77 @@ const rows = [
...
-                        setTimeout(() => {
-                            // const dataDelete = [...data];
-                            // const index = oldData.tableData.id;
-                            // dataDelete.splice(index, 1);
-                            // setData([...dataDelete]);
+        <div>
+            <AnalysisRunnerDialog
+                participants={activeParticipants}
+                open={showRunner}
+                onClose={() => {setRunner(false)}}
```
kevinlul, line 67: Unnecessary curly braces around the body of this arrow function.

react/src/pages/participants/ParticipantsTable.tsx
```diff
@@ -45,69 +56,77 @@ const rows = [
...
+                        onClick: (evt, data) => {
+                            const sampleString = (data as Participant[]).map((participant) => {return participant.participantID}).join(', ')
+                            alert(`Widthdraw all datasets and records associated with the samples: ${sampleString}`)
+                        }
+                    },
+                    {
+                        tooltip: 'Run an analysis with the selected datasets',
+                        icon: PlayArrowIcon,
+                        onClick: (evt, data) => {
+                            setActiveParticipants(data as Participant[])
```
kevinlul, line 124: Apparently Material Table's second parameter passed to onClick can be either a single object or an array so I'm not sure if there's a way around this and how we know it will always be the intended type? I've seen this cast in several places but will have to investigate further.

react/src/pages/analysis/Analysis.tsx
dennis-kao: Type cast is needed, otherwise I get this error since activeRow is of type AnalysisRun | null:

Screen Shot 2020-07-07 at 1 48 41 PM

kevinlul: That's strange, I'm not getting then because with `activeRow &&` it implies that for the rest of the expression `activeRow` cannot be `null`...

react/src/pages/participants/ParticipantsTable.tsx
dennis-kao: I have no idea either ... dealing with this was a pain. I've noticed that the return type always seems to be an array even when 1 element in selected. No idea why a single object is there but yes this will have to be solved later.

react/src/pages/analysis/CancelAnalysisDialog.tsx
dennis-kao: Good idea!

PR #18, opened 2020-09-10
Flask: GET /api/participant, /api/datasets, /api/analyses, /api/pipelines

Madelinehazel:
I created endpoints for samples (participants), analyses, and pipelines. I'm not sure if i've handled returning the tissue_samples from participants correctly.

### DIFF
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -173,11 +173,12 @@ def change_password():
         return 'Server error', 500
 
 @app.route('/api/participant', methods=['GET'], endpoint='partipants_list')
+@login_required
 def participants_list():
-    db_participants = db.session.query(models.Participant).all()
+    db_participants = db.session.query(models.Participant).join(models.Family)
     participants = [
         {
-            'family_codename': participant.family_id,
+            'family_codename': participant.family.family_codename,
             'participant_codename': participant.participant_codename,
             'sex': participant.sex,
             'participant_type': participant.participant_type,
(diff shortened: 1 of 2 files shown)


flask/app/routes.py
```diff
@@ -162,3 +171,61 @@ def change_password():
...
+            'family_codename': participant.family_id,
+            'participant_codename': participant.participant_codename,
+            'sex': participant.sex,
+            'participant_type': participant.participant_type,
+            'affected': participant.affected,
+            'solved': participant.solved,
+            'created': participant.created,
+            'created_by': participant.created_by,
+            'updated': participant.updated_by,
+            'tissue_samples': [tissue for tissue in participant.tissue_samples.all()]
```
helrick, line 189: My only thought here is that maybe we should show the datasets of a participant instead of (or in addition to) tissue_samples. Unless there was a reason it's left out that I'm forgetting?

kevinlul (changes requested):

flask/app/routes.py
Madelinehazel: That would make sense, should I add that into the participant model?

flask/app/routes.py
```diff
@@ -162,3 +171,61 @@ def change_password():
     except:
         db.session.rollback()
         return 'Server error', 500
+
+@app.route('/api/participant', methods=['GET'], endpoint='partipants_list')
```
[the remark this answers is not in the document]
Madelinehazel: For some reason I was getting this error: 'AssertionError: View function mapping is overwriting an existing endpoint function' even though there were no duplicate routes or functions. Adding the endpoint parameter resolves this, but maybe there's a better solution?

flask/app/routes.py
helrick: Would it be possible to do something like `'datasets':` [tissue.datasets.all() for tissue in participant.tissue_samples.all()]`? I've got another branch checked out and have loaded up the legacy database in it so can't test this myself super easily, but it looks like there's a link to the datasets via the TissueSample table.

Madelinehazel: That works! I just need to figure out how to serialize the tissue object.

PR #21, opened 2020-09-18
Rnaseq database

helrick:
Updates on RNASeq specific portions of database, based on convo with Alper


flask/migrations/versions/0d8af5e5d7f4_rnaseq_updates.py
```diff
@@ -0,0 +1,28 @@
+"""RNASeq updates
```
kevinlul, line 1: Changes to tissue sample type enumeration not captured here

flask/app/models.py
```diff
@@ -76,16 +76,63 @@ class Participant(db.Model):
 
 
 class TissueSampleType(Enum):
-    Blood = 'Blood'  # BLO
```
kevinlul, line 79: The _name_ of the enumeration is used by SQLAlchemy in the database and the _value_ only has an effect on the Python side. This means that when serialized to SQL, the names will be Pascal_Snake_Case but when serialized to JSON, the names with be "human readable - with spaces". I think this is confusing and we should just use the same names on both sides.

flask/app/models.py
```diff
@@ -76,16 +76,63 @@ class Participant(db.Model):
...
+    Skin_Not_Sun_Exposed = 'Skin - Not Sun Exposed (Suprapubic)'
+    Skin_Sun_Exposed = 'Skin - Sun Exposed (Lower leg)'
+    Small_Intestine_Terminal_Ileum = 'Small Intestine - Terminal Ileum'
+    Spleen = 'Spleen'
+    Stomach = 'Stomach'
+    Testis = 'Testis'
+    Thyroid = 'Thyroid'
+    Uterus = 'Uterus'
+    Vagina = 'Vagina'
+    Whole_Blood = 'Whole Blood' # BLO
```
kevinlul, line 134: Change dummy data to account for name changes too.

kevinlul (changes requested):

flask/app/models.py
helrick: Hmm, that's a good point. I guess the only problem that arises is when we try to map these to the GTEx samples for the rnaseq pipeline (i.e. should we do a mapping or change the pipeline reference). Since several of them have special characters (parentheses and dashes) they may not be automatically map-able.

This makes me wonder whether it's actually better to just eliminate references to the GTEx database in the Tissue table and make it a RNASeq dataset specific attribute. Or something selected at analysis runtime. Thoughts?

PR #24, opened 2020-09-21
draft PATCH endpoint for participants

delvinso:
```
from datetime import datetime 
import requests 
import json


d = {
        'participant_id': '1',                                 # will be removed from the dictionary as we don't want this field updated
        'participant_codename': 'agent 10',
        'notes': 'this is my super interesting note',
        'created_by' : 'hahahah',                       # will be removed from the dictionary as we don't want this field updated 
        'created' : 'fake_date_123',                    # will be removed from the dictionary as we don't want this field updated 
        'participant_type': 'Father',
        'sex': 'Male', # how to convert this into an enum for ORM? - don't need to. it will fail if the enum is wrong
    }


# ----- before ----

# mysql> select * from participant where participant_id = "1";
# +----------------+-----------+----------------------+--------+------------------+----------+--------+------------------------------+---------------------+------------+---------------------+------------+
# | participant_id | family_id | participant_codename | sex    | participant_type | affected | solved | notes                        | created             | created_by | updated             | updated_by |
# +----------------+-----------+----------------------+--------+------------------+----------+--------+------------------------------+---------------------+------------+---------------------+------------+
# |              1 |         1 | AA001                | Female | Proband          |        1 |   NULL | Extra info about sample here | 2020-09-23 18:40:48 |          1 | 2020-09-23 18:40:48 |          1 |
# +----------------+-----------+----------------------+--------+------------------+----------+--------+------------------------------+---------------------+------------+---------------------+------------+
# 1 row in set (0.00 sec)

url = 'http://0.0.0.0:5000/api/participants/1'
result = requests.patch(url, json = d)
result.json()
# we can see the created and created by fields were not updated
{'participant_id': '1', 'family_id': '1', 'participant_codename': 'agent 10', 'sex': 'Sex.Male', 'participant_type': 'ParticipantType.Father', 'affected': 'True', 'solved': 'None', 'notes': 'this is my super interesting note', 'created': '2020-09-23 18:40:48', 'created_by': '1', 'updated': '2020-09-23 18:52:33', 'updated_by': '1'}

# ----- after -------
# mysql> select * from participant where participant_id = "1";
# +----------------+-----------+----------------------+------+------------------+----------+--------+-----------------------------------+---------------------+------------+---------------------+------------+
# | participant_id | family_id | participant_codename | sex  | participant_type | affected | solved | notes                             | created             | created_by | updated             | updated_by |
# +----------------+-----------+----------------------+------+------------------+----------+--------+-----------------------------------+---------------------+------------+---------------------+------------+
# |              1 |         1 | agent 10             | Male | Father           |        1 |   NULL | this is my super interesting note | 2020-09-23 18:40:48 |          1 | 2020-09-23 18:42:18 |          1 |
# +----------------+-----------+----------------------+------+------------------+----------+--------+-----------------------------------+---------------------+------------+---------------------+------------+
# 1 row in set (0.00 sec)
```

### DIFF
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -6,6 +6,7 @@ from flask import request
 from flask_login import login_user, logout_user, current_user, login_required
 
 from app import app, db, login, models
+from datetime import datetime
 
 
 @login.user_loader
@@ -162,3 +163,42 @@ def change_password():
     except:
         db.session.rollback()
         return 'Server error', 500
+
+
+@app.route('/api/participants/<participant_id>', methods = ['PATCH'])
+# @login_required # check document for how to test endpoints with login signing? 
+def update_participants(participant_id):
+    patch_json = request.get_json()
+
+    # convert to dictionary
+    patch_d = json.loads(patch_json)
+
+    # get the PK
+    id = patch_d['participant_id']
+
+    # check if PK exists, is this necessary ??
+    q = db.session.query(models.Participant.participant_id)\
+        .filter_by(participant_id = id)
+
+    if q.scalar() == 1:
+        # add time of edit
+        patch_d['updated'] = datetime.now()
+
+        q.update(dict(patch_d)) # for more granular control, i was thinking of looping through
+                                # the dictionary and updating so we prevent the primary key from being changed? 
+        try:
+            db.session.commit()
+            # TODO: return the json which will update the front-end
+            return 'success!!!!', 200
+
+        except:
+            db.session.rollback()
+            return 'oh no something has gone terribly wrong', 500
+    else:
+        return 'oh no something has gone terribly wrong', 400
+
+# @app.route('/api/datasets/<dataset_id>', methods = ['PATCH'])
+# @login_required
+
+# @app.route('/api/analyses/<analysis_id>', methods = ['PATCH'])
+# @login_required
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -9,6 +9,17 @@ from app import app, db, login, models
 from datetime import datetime
 
 
+def row2dict(row):
+    """
+    converts an sqlalchemy row into a dictionary, credits to: https://stackoverflow.com/questions/1958219/convert-sqlalchemy-row-object-to-python-dict
+    alternative to '.__dict__' which returns an extra '_sa_instance_state' field
+    """
+    d = {}
+    for column in row.__table__.columns:
+        d[column.name] = str(getattr(row, column.name))
+    return d
+    
+
 @login.user_loader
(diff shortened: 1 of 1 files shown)


flask/app/routes.py
```diff
@@ -162,3 +163,42 @@ def change_password():
     except:
         db.session.rollback()
         return 'Server error', 500
+
+
+@app.route('/api/participants/<participant_id>', methods = ['PATCH'])
+# @login_required # check document for how to test endpoints with login signing?
```
kevinlul, line 169: Keep this on. For your testing you can either comment this out (but don't commit it) or in config.py add `LOGIN_DISABLED = True` to disable it globally.

flask/app/routes.py
```diff
@@ -162,3 +163,42 @@ def change_password():
...
         return 'Server error', 500
+
+
+@app.route('/api/participants/<participant_id>', methods = ['PATCH'])
+# @login_required # check document for how to test endpoints with login signing? 
+def update_participants(participant_id):
+    patch_json = request.get_json()
+
+    # convert to dictionary
+    patch_d = json.loads(patch_json)
```
kevinlul, line 174: `request.get_json()` already returns dict or None. Check for `None` and return a 400 Bad Request.

flask/app/routes.py
```diff
@@ -162,3 +163,42 @@ def change_password():
...
+
+    # get the PK
+    id = patch_d['participant_id']
```
kevinlul, line 177: Look in the admin routes for an example. Python dictionaries raise a KeyError if it's not found so you'll want to handle this case too.

flask/app/routes.py
```diff
@@ -162,3 +163,42 @@ def change_password():
...
+
+    # check if PK exists, is this necessary ??
+    q = db.session.query(models.Participant.participant_id)\
+        .filter_by(participant_id = id)
```
kevinlul, line 181: Yes, you should check if the primary key exists and respond with a 404 Not Found appropriately.

flask/app/routes.py
```diff
@@ -162,3 +163,42 @@ def change_password():
...
+
+    if q.scalar() == 1:
+        # add time of edit
+        patch_d['updated'] = datetime.now()
+
+        q.update(dict(patch_d)) # for more granular control, i was thinking of looping through
+                                # the dictionary and updating so we prevent the primary key from being changed?
```
kevinlul, line 188: Yes, that's the idea. Use the ORM to fetch the corresponding Participant object, update only fields we allow (so no primary key overwrites or changing the dates, for example), commit and then return the object serialized as JSON, or return the error.

PR #26, opened 2020-09-23
React: change participants page to show datasets

kevinlul:
DatasetTable is modified from the old ParticipantTable but the delta is apparently too large in commit for Git to consider it a move-and-edit.


annnag (changes requested):

react/src/pages/datasets/AnalysisRunnerDialog.tsx
```diff
@@ -19,7 +19,7 @@ import { annotations } from '../analysis/AnalysisInfoDialog';
 import ChipStrip from '../analysis/ChipStrip';
 
 interface AnalysisRunnerDialogProp {
-    participants: Participant[],
+    participants: Dataset[],
```
frewmack, line 22: The participants prop should probably be renamed to 'datasets' in this component

react/src/pages/datasets/DatasetTable.tsx
```diff
@@ -0,0 +1,147 @@
...
+                open={showRunner}
+                onClose={() => setRunner(false)}
+            />
+            <MaterialTable
+                columns={[
+                    { title: 'Participant', field: 'participant_codename', editable: 'never' },
+                    { title: 'Family', field: 'family_codename', editable: 'never' },
+                    { title: 'Tissue sample', field: 'tissue_sample_type' },
+                    { title: 'Dataset type', field: 'dataset_type', defaultFilter: datasetType },
+                    { title: 'Condition', field: 'condition' },
```
frewmack, line 81: These 3 columns display values chosen from a pre-defined list (eg. Tissue Sample: Blood, Saliva, Skin, etc.), and so you should edit the column definitions to reflect that. 

MaterialTable has an option for columns called `lookup` to provide said values to select from when filtering or editing rows.

eg. `{ title: 'Condition', field: 'condition', lookup: { 0: 'Control', 1: 'GermLine', 2: 'Somatic' } },` I think the key determines what order they are displayed in, but it otherwise doesn't matter.

frewmack (changes requested):

react/src/pages/datasets/DatasetTable.tsx
kevinlul: The key is the actual value in the table, and the value is displayed in the filter dropdown. If numeric keys are used then the cells for this column are blank, I think because they don't match.

react/src/pages/datasets/DatasetTable.tsx
```diff
@@ -0,0 +1,168 @@
...
+                            }, 1000);
+                        }),
+                }}
+                components={{
+                    Toolbar: props => (
+                        <div>
+                            <MTableToolbar {...props} />
+                            <div className={classes.chipBar}>
+                                {[...new Set(rows.map(e => e.dataset_type))].map(type => (
+                                    <Chip label={type} onClick={() => setDatasetType(type)} clickable className={classes.chip} />
```
annnag, line 136: When I clicked the chip button, the page goes blank. You can solve this problem by making `datasetType` an array and passing an array containing the filter values to `setDatasetType`.
```suggestion
                                    <Chip label={type} onClick={() => setDatasetType([type])} clickable className={classes.chip} />
```

annnag (changes requested):

react/src/pages/datasets/DatasetTable.tsx
kevinlul: Great catch, thanks!

annnag (approved):
Looks good to me 😃

PR #27, opened 2020-09-24
Per-column table filtering 

frewmack:
The material-table component provides built-in per-column filtering options. This change nullifies the need for a general search bar, so it is also disabled.

### DIFF
--- a/react/src/pages/dashboard/SamplesTable.tsx
+++ b/react/src/pages/dashboard/SamplesTable.tsx
@@ -2,6 +2,8 @@ import React, { useState } from 'react';
 import { makeStyles } from '@material-ui/core/styles';
 import Chip from '@material-ui/core/Chip';
 import MaterialTable, { MTableToolbar } from 'material-table';
+import { Cancel } from '@material-ui/icons';
+import { IconButton } from '@material-ui/core';
 
 
 export interface Sample {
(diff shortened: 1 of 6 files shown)


react/src/pages/dashboard/SamplesTable.tsx
```diff
@@ -2,6 +2,8 @@ import React, { useState } from 'react';
 import { makeStyles } from '@material-ui/core/styles';
 import Chip from '@material-ui/core/Chip';
 import MaterialTable, { MTableToolbar } from 'material-table';
+import { Cancel } from '@material-ui/icons';
+import { IconButton } from '@material-ui/core';
```
kevinlul, line 6: Could you combine all the `core` imports as one statement since there's a few now? Bundle size should not be a concern since [material-ui supports tree-shaking](https://material-ui.com/guides/minimizing-bundle-size/).

react/src/pages/participants/ParticipantsTable.tsx
```diff
@@ -5,6 +5,8 @@ import MaterialTable, { MTableToolbar } from 'material-table';
 import PlayArrowIcon from '@material-ui/icons/PlayArrow';
 import DeleteIcon from '@material-ui/icons/Delete';
 import AnalysisRunnerDialog from './AnalysisRunnerDialog';
+import { Cancel } from '@material-ui/icons';
+import { IconButton } from '@material-ui/core';
```
kevinlul, line 9: Ditto for here but also combine the icon imports please.

react/src/pages/participants/ParticipantsTable.tsx
```diff
@@ -70,7 +73,7 @@ export default function ParticipantsTable({ display }: ParticipantsTableProps) {
                 columns={[
                     { title: 'Participant', field: 'participantID' },
                     { title: 'Project', field: 'project' },
-                    { title: 'Uploader', field: 'uploader' },
+                    { title: 'Uploader', field: 'uploader', defaultFilter: centre},
```
kevinlul, line 76: style: space between `centre` and `}`

kevinlul (changes requested):
Looking good! I like that you were able to integrate the chip buttons with the filters too seamlessly.

kevinlul (approved):
LGTM

PR #28, opened 2020-09-24
update page name to match the current page

### LINKED ISSUE #25: Display page name in app bar

annnag:
The page name now matches the page you are on. Closes #25

### DIFF
--- a/react/src/pages/ListItemRouterLink.tsx
+++ b/react/src/pages/ListItemRouterLink.tsx
@@ -8,17 +8,14 @@ export interface ListItemRouterLinkProps {
     primary: React.ReactNode;
     to: string;
     children: React.ReactNode;
-    updatePageName: Function;
 }
 
 // https://material-ui.com/guides/composition/#caveat-with-inlining
-export default function ListItemRouterLink({ primary, to, children, updatePageName }: ListItemRouterLinkProps) {
+export default function ListItemRouterLink({ primary, to, children }: ListItemRouterLinkProps) {
     const ForwardedRouterLink = React.useMemo(
         () =>
             React.forwardRef<HTMLAnchorElement>((linkProps, ref) => (
-                <Link ref={ref} to={to} {...linkProps} onClick={(e: React.MouseEvent) => {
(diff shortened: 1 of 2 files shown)


react/src/pages/ListItemRouterLink.tsx
```diff
@@ -8,14 +8,17 @@ export interface ListItemRouterLinkProps {
...
 // https://material-ui.com/guides/composition/#caveat-with-inlining
-export default function ListItemRouterLink({ primary, to, children }: ListItemRouterLinkProps) {
+export default function ListItemRouterLink({ primary, to, children, updatePageName }: ListItemRouterLinkProps) {
     const ForwardedRouterLink = React.useMemo(
         () =>
             React.forwardRef<HTMLAnchorElement>((linkProps, ref) => (
-                <Link ref={ref} to={to} {...linkProps} />
+                <Link ref={ref} to={to} {...linkProps} onClick={(e: React.MouseEvent) => {
+                    updatePageName((e.target as HTMLInputElement).innerText);
+                }}/>
```
kevinlul, line 21: This actually causes the page state to be set twice if using the sidebar instead of direct links: once when clicking the link above and again when the render function for the route is called. Since the latter occurs in all cases, then the click event isn't necessary.

react/src/pages/Navigation.tsx
```diff
@@ -129,12 +129,16 @@ export interface NavigationProps {
...
+    const [pageName, setPageName] = React.useState("Dashboard");
     const handleDrawerOpen = () => {
         setOpen(true);
     };
     const handleDrawerClose = () => {
         setOpen(false);
     };
+    const updatePageName = (name: string) => {
+        setPageName(name)
+    }
```
kevinlul, line 141: While this function isn't needed because of the reasons I mentioned above, it also really doesn't serve a purpose since it's not doing anything on top of `setPageName`.

kevinlul (changes requested):
The app bar title change works! Just requesting some changes to get rid of redundancy.

kevinlul (approved):
LGTM

PR #34, opened 2020-09-25
Dialog guiding users from Analysis to Participants when creating new analysis

frewmack:


react/src/pages/analysis/Analysis.tsx
```diff
@@ -8,6 +8,8 @@ import MaterialTable from 'material-table';
 import Title from '../Title';
 import CancelAnalysisDialog from './CancelAnalysisDialog';
 import AnalysisInfoDialog from './AnalysisInfoDialog';
+import { useHistory } from 'react-router';
```
kevinlul, line 11: Can you put the external library imports before internal ones?

react/src/pages/analysis/AddAnalysisAlert.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+export default function AddAnalysisAlert(props: AddAnalysisAlertProps) {
+    return (
+        <Dialog
+            open={props.open}
+            onClose={props.onClose}
+            aria-labelledby="alert-dialog-title"
+            aria-describedby="alert-dialog-description"
+        >
+            <DialogTitle id="alert-dialog-title">{"Add new Analysis"}</DialogTitle>
```
kevinlul, line 19: Don't need curly braces around the body of the title. I think `Request analysis` might be a better title.

kevinlul (changes requested):

PR #35, opened 2020-09-28
Add last_login timestamp in database and update on login

### LINKED ISSUE #32: Add last logged in timestamp to User model
This will support the "analyses that have changed since you last signed in" feature on the dashboard.

helrick:
updating schema, endpoints, and flask migration files to enable last_login field for a user. Closes #32


flask/app/routes.py
```diff
@@ -26,6 +30,8 @@ def login():
     if user is None or not user.check_password(body['password']):
         return 'Unauthorized', 401
     login_user(user)
+    user.last_login = datetime.now()
+    db.session.commit()
     return json.dumps({ "username": user.username }), 200
```
kevinlul, line 35: Do we want to try/except the db commits in case something goes awry or is it acceptable to throw an error 500 when trying to sign in if this fails?

helrick: Yeah, that's a good point, I'll update

helrick: I added a new commit, it's just a 500 error to match the other wrapped db.commits in other endpoints but I can update if there's a more specific error code we can show here

flask/app/routes.py
```diff
@@ -16,7 +18,14 @@ def load_user(uid: int):
...
 def login():
     if current_user.is_authenticated:
-        return json.dumps({ "username": current_user.username }), 200
+        # get/update last login
+        last_login = current_user.last_login
+        try:
+            current_user.last_login = datetime.now()
+            db.session.commit()
+        except:
+            app.logger.info('failed to update last_login for %s', current_user.username)
```
kevinlul, line 27: Should be a warning

flask/app/routes.py
```diff
@@ -16,7 +18,14 @@ def load_user(uid: int):
...
+        return jsonify({ "username": current_user.username, "last_login": current_user.last_login }), 200
```
kevinlul, line 28: Should return `last_login` or we're returning `datetime.now()` again! 
Also `jsonify` already returns a response object so we don't need the `, 200` in tuple

flask/app/routes.py
```diff
@@ -25,8 +34,17 @@ def login():
...
+    last_login = current_user.last_login
+    try:
+        user.last_login = datetime.now()
+        db.session.commit()
+    except:
+        app.logger.info('failed to update last_login for %s', current_user.username)
+
     login_user(user)
-    return json.dumps({ "username": user.username }), 200
+    return jsonify({ "username": user.username, "last_login": current_user.last_login}), 200
```
kevinlul, line 47: ditto

flask/app/routes.py
```diff
@@ -2,11 +2,13 @@
 
 from functools import wraps
 
-from flask import request
+from flask import jsonify, request
 from flask_login import login_user, logout_user, current_user, login_required
 
 from app import app, db, login, models
 
+from datetime import datetime
```
kevinlul, line 10: Put this up top with the other Python standard library imports

kevinlul (changes requested):

flask/app/routes.py
helrick: resolved in 7e525b6

flask/app/routes.py
helrick: 7e525b6

flask/app/routes.py
helrick: resolved in 7e525b6

flask/app/routes.py
helrick: 👍resolved in 7e525b6

kevinlul (approved):
LGTM, resolve merge conflicts

PR #36, opened 2020-09-28
React: add collapsible notification panel on dashboard

annnag:
Collapsible notification panel added. 
Since backend API is not implemented yet, fake data is used instead of fetch(). 
Please review.


react/src/pages/analysis/Analysis.tsx
```diff
@@ -107,9 +108,10 @@ const analyses = [
 
 export default function Analysis() {
     const classes = useStyles();
-    const [detail, setDetail] = useState(false);
+    let { analysisID } : any = useParams();
```
kevinlul, line 111: If you use the angle bracket syntax to specify the type like below, `any` can be avoided. Also I think you can use `const` here.

kevinlul (changes requested):
This looks really good! Some high-level things I noticed:
- after the analysis dialog is closed, the address bar remains `/analysis/:id`
- after the analysis dialog is closed, we are now on the analyses page instead of the dashboard, which might be unintuitive
- there's a merge conflict with `master` because a file was deleted on this branch that happened to be changed on `master` that should be resolvable by merging from command-line and confirming the deletion

react/src/pages/analysis/Analysis.tsx
```diff
@@ -136,7 +141,7 @@ export default function Analysis() {
                 <AnalysisInfoDialog
                     open={detail}
                     analysis={activeRow}
-                    onClose={() => setDetail(false)}
+                    onClose={() => {setDetail(false); if(analysisID){history.push("/dashboard")}}}
```
kevinlul, line 144: I think [`history.goBack()`](https://reactrouter.com/web/api/history) would be better so we can have multiple points of entry to this view.

react/src/pages/analysis/Analysis.tsx
```diff
@@ -107,11 +107,16 @@ const analyses = [
...
+  }
+
 export default function Analysis() {
     const classes = useStyles();
-    const [detail, setDetail] = useState(false);
+    const { analysisID }= useParams<ParamTypes>();
+    const [detail, setDetail] = useState(true);
     const [cancel, setCancel] = useState(false);
-    const [activeRow, setActiveRow] = useState<AnalysisRun | null>(null);
+    const [activeRow, setActiveRow] = useState<AnalysisRun | null>(analyses.filter(analysis => analysis.analysisID === analysisID)[0]);
```
kevinlul, line 119: You can just use [`Array.find`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead of `Array.filter()[0]`

react/src/pages/analysis/Analysis.tsx
```diff
@@ -107,11 +107,16 @@ const analyses = [
     createAnalysis(8, 'AN20032', '2020-06-20 7:07 AM', 'User C', '3839', ['CC773', 'CC774', 'CC775'], 'CRE', '22hrs', PipelineStatus.ERROR),
 ];
 
+type ParamTypes = {
+    analysisID: string | undefined
+  }
```
kevinlul, line 112: funky indentation

react/src/pages/dashboard/Notification.tsx
```diff
@@ -0,0 +1,74 @@
+import React from 'react';
+import { Link } from'react-router-dom';
```
kevinlul, line 2: missing space
```suggestion
import { Link } from 'react-router-dom';
```

react/src/pages/dashboard/NotificationPanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { makeStyles } from '@material-ui/core/styles';
+import Paper from '@material-ui/core/Paper';
+import { AnalysisRun } from '../analysis/Analysis';
+import Box from '@material-ui/core/Box';
+import Typography from '@material-ui/core/Typography';
+import Switch from "@material-ui/core/Switch";
+import Collapse from "@material-ui/core/Collapse";
+import FormControlLabel from "@material-ui/core/FormControlLabel";
+import NotificationsActiveIcon from '@material-ui/icons/NotificationsActive';
+import Notification from './Notification';
```
kevinlul, line 11: User imports after library imports please!

react/src/pages/dashboard/NotificationPanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+        padding: theme.spacing(2),
+        display: 'flex',
+        overflow: 'auto',
+        flexDirection: 'column',
+    },
+    alert: {
+        width: '100%',
+        '& > * + *': {
+          marginTop: theme.spacing(2),
+        },
```
kevinlul, line 24: This is pretty cool 👀

react/src/pages/dashboard/NotificationPanel.tsx
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    return (
+        <Paper className={classes.paper}>
+            <Box className={classes.titleBox}>
+                <NotificationsActiveIcon />
+                <Typography component="h1" variant="h6" color="inherit" noWrap className={classes.title} display="inline">
+                    Notifications
+                </Typography>
+                <FormControlLabel control={<Switch checked={checked} onChange={handleChange} />} label=""/>
```
kevinlul, line 55: Since the label is actually the `Typography`, I'm not sure if we need a blank `FormControlLabel` here.

kevinlul (changes requested):

kevinlul (approved):

PR #37, opened 2020-09-28
Flask: PATCH /api/participants/:id, /api/datasets/:id, /api/analyses/:id in one function

delvinso:
A lot of the code was reused between the different PATCH endpoints so I thought it would be better to have one generic view. I'm open to splitting this into individual routes for each table though.

### DIFF
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -180,16 +180,13 @@ def mixin(entity: db.Model, json_mixin: Dict[str, Any], columns: List[str]) -> U
 
 
 @app.route('/api/<model_name>/<int:id>', methods = ['PATCH'])
-# eg. /api/analyses/1 for analysis table, /api/participants/1 for participants table, and so forth
 @login_required
 def update_db(model_name:str, id:int):
     if not request.json:
         return 'Request body must be JSON', 415
-    print(model_name)
-    #return(jsonify(model_name))
+
     editable_columns = []
 
-    # can these if statements be generalized?
     if model_name == 'participants':
         table = models.Participant.query.get(id)
         editable_columns = ['participant_codename', 'sex', 'participant_type',
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -190,7 +190,7 @@ def update_entity(model_name:str, id:int):
         editable_columns = ['participant_codename', 'sex', 'participant_type',
                              'affected', 'solved', 'notes']
     elif model_name == 'datasets':
-        table = models.Datasets.query.get(id)
+        table = models.Dataset.query.get(id)
         editable_columns = ['dataset_type', 'input_hpf_path', 'notes', 'condition',
                             'extraction_protocol', 'capture_kit', 'library_prep_method',
                             'library_prep_date', 'read_length', 'read_type', 'sequencing_id',
@@ -205,7 +205,11 @@ def update_entity(model_name:str, id:int):
     else:
         return 'Not Found', 404
 
+    if not table:
+         return 'Not Found', 404
+
     enum_error = mixin(table, request.json, editable_columns)
+    
     if enum_error:
         return enum_error, 400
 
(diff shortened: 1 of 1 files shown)


flask/app/routes.py
```diff
@@ -179,33 +179,52 @@ def mixin(entity: db.Model, json_mixin: Dict[str, Any], columns: List[str]) -> U
...
-def update_participants(participant_id: int):
+def update_db(model_name:str, id:int):
     if not request.json:
         return 'Request body must be JSON', 415
-
-    participant = models.Participant.query.get(participant_id)
-    if not participant:
+    print(model_name)
+    #return(jsonify(model_name))
+    editable_columns = []
```
kevinlul, line 190: Not needed

flask/app/routes.py
```diff
@@ -179,33 +179,52 @@ def mixin(entity: db.Model, json_mixin: Dict[str, Any], columns: List[str]) -> U
...
+
+    # can these if statements be generalized?
```
kevinlul, line 192: I don't think so. We don't want to allow arbitrary writes to tables.

kevinlul (changes requested):

flask/app/routes.py
delvinso: Oops, forgot to take that out

flask/app/routes.py
```diff
@@ -179,33 +179,49 @@ def mixin(entity: db.Model, json_mixin: Dict[str, Any], columns: List[str]) -> U
             setattr(entity, field, value)
 
 
-@app.route('/api/participants/<int:participant_id>', methods = ['PATCH'])
+@app.route('/api/<model_name>/<int:id>', methods = ['PATCH'])
 @login_required
-def update_participants(participant_id: int):
+def update_db(model_name:str, id:int):
```
kevinlul, line 184:
```suggestion
def update_entity(model_name: str, id: int):
```

flask/app/routes.py
```diff
@@ -179,33 +179,49 @@ def mixin(entity: db.Model, json_mixin: Dict[str, Any], columns: List[str]) -> U
...
     if not request.json:
         return 'Request body must be JSON', 415
 
-    participant = models.Participant.query.get(participant_id)
-    if not participant:
+    editable_columns = []
```
kevinlul, line 188: Don't believe this is necessary either because the `else` branch returns out of the function and we're not doing the same for `table` either

kevinlul (changes requested):

flask/app/routes.py
```diff
@@ -179,33 +179,45 @@ def mixin(entity: db.Model, json_mixin: Dict[str, Any], columns: List[str]) -> U
...
         return 'Request body must be JSON', 415
 
-    participant = models.Participant.query.get(participant_id)
-    if not participant:
+    if model_name == 'participants':
+        table = models.Participant.query.get(id)
+        editable_columns = ['participant_codename', 'sex', 'participant_type',
+                             'affected', 'solved', 'notes']
+    elif model_name == 'datasets':
+        table = models.Datasets.query.get(id)
```
kevinlul, line 193: small typo
```suggestion
        table = models.Dataset.query.get(id)
```

flask/app/routes.py
```diff
@@ -179,33 +179,45 @@ def mixin(entity: db.Model, json_mixin: Dict[str, Any], columns: List[str]) -> U
...
-@app.route('/api/participants/<int:participant_id>', methods = ['PATCH'])
+@app.route('/api/<model_name>/<int:id>', methods = ['PATCH'])
 @login_required
-def update_participants(participant_id: int):
+def update_entity(model_name:str, id:int):
     if not request.json:
         return 'Request body must be JSON', 415
 
-    participant = models.Participant.query.get(participant_id)
-    if not participant:
```
kevinlul, line 189: the `None` check got deleted but it's necessary because we want to 404 if the entity does not exist, not 500 trying to access a property of `None`

kevinlul (changes requested):

kevinlul (approved):
LGTM

Issue #49, opened 2020-09-29
Analysis action button: set assignee


kevinlul:
Set the assignee for multiple analyses to a particular user

Requires #37 for mutations

Issue #47, opened 2020-09-29
Analysis action button: change status


kevinlul:
Set the state of multiple analyses to started/finished/cancelled

Requires #37 for mutations

PR #56, opened 2020-09-30
Analysis page layout changes

### LINKED ISSUE #46: Analysis page layout updates
Columns to display: state, pipeline, assignee, notes, updated date, requester, result_hpf_path
Editable fields: notes, result_hpf_path
Add chip buttons for filtering by running status

Requires #37 for mutations

frewmack:
All the changes in issue #46 are handled here, however I made some choices regarding the format that I might need someone to clarify or suggest changes:

- The row data is changed based on the database schema, and the Assignee, Requester and Pipeline columns all display their database IDs instead of a username string. I'm pretty sure displaying the name is better but I just want to be 100% sure.
- I setup the 'Notes' column so that it can be displayed one of two ways (currently using the 1st one):
  1. Display a block of text with fixed width, which expands the height of the row as it gets longer. This option requires that the column be somewhat wide.
  2. Display an Icon that reveals the text notes in an interactive Tooltip when the user hovers over it. No column width/row height changes required.

Closes #46

EDIT: I'll probably re-use this branch for issues #47, #49


kevinlul (commented):
Definitely prefer displaying the actual usernames over their database IDs. I think this can be retrieved by the same API that fetches the user list but if exposing the internal database ID is too leaky in the case of usernames then maybe we can change the backend API to never reveal that, only usernames.

I like the tooltip approach to the notes better since it seems like it can still be copied.

EDIT: Forgot to say this but there's a merge conflict with `master` as #36 was merged.

frewmack:
I changed the analysis rows to use text instead of number IDs, and the notes column now uses the text display as we discussed prior. Should be ready for review now.

react/src/pages/analysis/Analysis.tsx
```diff
@@ -152,48 +198,92 @@ export default function Analysis() {
...
                         rowData => ({
-                            icon: CancelIcon,
+                            icon: Cancel,
                             tooltip: 'Cancel analysis',
                             onClick: () => {
                                 setActiveRow(rowData)
                                 setCancel(true)
                             },
-                            disabled: rowData.status !== PipelineStatus.RUNNING,
+                            disabled: rowData.state !== PipelineStatus.RUNNING,
```
kevinlul, line 245: Should be able to cancel if pending too.

react/src/pages/analysis/Analysis.tsx
```diff
@@ -152,48 +198,92 @@ export default function Analysis() {
...
+                                const dataUpdate = [...rows];
+                                // find the row; assume analysis_id is unique
+                                const index = dataUpdate.findIndex((row, index, obj) => {
+                                    return row.analysis_id === oldData?.analysis_id
+                                });
+
+                                // only update the rows that are allowed
+                                dataUpdate[index].notes = newData.notes;
+                                dataUpdate[index].result_hpf_path = newData.result_hpf_path;
+                                setRows([...dataUpdate]);
```
kevinlul, line 266: Don't think making a copy here again is necessary.

react/src/pages/analysis/Analysis.tsx
```diff
@@ -152,48 +198,92 @@ export default function Analysis() {
...
+                            new Promise((resolve, reject) => {
+                                const dataUpdate = [...rows];
+                                // find the row; assume analysis_id is unique
+                                const index = dataUpdate.findIndex((row, index, obj) => {
+                                    return row.analysis_id === oldData?.analysis_id
+                                });
+
+                                // only update the rows that are allowed
+                                dataUpdate[index].notes = newData.notes;
+                                dataUpdate[index].result_hpf_path = newData.result_hpf_path;
```
kevinlul, line 265: This still mutates the previous state because this is not a deep copy, which might have unexpected consequences. I think you should update the properties on a copy of `dataUpdate[index]` and then replace the array entry, if that makes sense.

react/src/pages/analysis/Analysis.tsx
```diff
@@ -152,48 +198,92 @@ export default function Analysis() {
...
-                        { title: 'Submitted', field: 'dateSubmitted', type: 'string' },
-                        { title: 'Submitted By', field: 'submittedBy', type: 'string' },
-                        { title: 'Project', field: 'project', type: 'string' },
-                        { title: 'Participants', field: 'participants', type: 'string', render: rowData => rowData.participants.join(', ') },
-                        { title: 'Pipeline', field: 'pipeline', type: 'string' },
-                        { title: 'Status', field: 'status', type: 'string' }
+                        { title: 'Analysis ID', field: 'analysis_id', type: 'string', editable: 'never', width: '8%' },
+                        { title: 'Pipeline ID', field: 'pipeline_id', type: 'string', editable: 'never', width: '8%' },
+                        { title: 'Assignee ID', field: 'assignee', type: 'string', editable: 'never', width: '8%' },
+                        { title: 'Requester ID', field: 'requester', type: 'string', editable: 'never', width: '8%' },
```
kevinlul, line 204: Get rid of `ID` in the display name for the latter three to save some display space.

react/src/pages/analysis/Analysis.tsx
```diff
@@ -152,48 +198,92 @@ export default function Analysis() {
...
                     }}
                     actions={[
                         {
-                            icon: VisibilityIcon,
+                            icon: Visibility,
                             tooltip: 'Analysis details',
                             onClick: (event, rowData) => {
-                                setActiveRow((rowData as AnalysisRun))
+                                setActiveRow((rowData as AnalysisRow))
                                 setDetail(true)
```
kevinlul, line 235: For consistency with how the notification panel changes the address bar, this should also be adjusted to use the history API.

kevinlul (changes requested):

frewmack:
I went down a little rabbit hole while trying to get react-router to open the analysis details dialog when I modify the URL through the browser. It turns out that it's more complicated than I originally thought, but I found this helpful link regarding the issue: https://stackoverflow.com/questions/27928372/react-router-urls-dont-work-when-refreshing-or-writing-manually?rq=1

TL;DR sending a new URL from your browser will send a server request instead of a client one that react-router can work with, or it will send back a locally cached page. I can see about setting up a workaround tomorrow if there's interest for it.

react/src/pages/analysis/Analysis.tsx
```diff
@@ -69,60 +71,104 @@ export enum PipelineStatus {
...
+    createAnalysis("A0006", "P02", '/bar/baz/', "User 1", "User 5", PipelineStatus.PENDING, '2020-06-19 10:00 AM', loremIpsum[2]),
+    createAnalysis("A0007", "P03", '/example/path/', "User 2", "User 5", PipelineStatus.RUNNING, '2020-06-19 9:09 AM', loremIpsum[0]),
+    createAnalysis("A0008", "P03", '/example/path/', "User 5", "User 4", PipelineStatus.COMPLETED, '2020-06-20 7:07 AM', loremIpsum[1])
 ];
 
+// Displays notes as an interactive tooltip
+function renderNotes(rowData: AnalysisRow) {
+    return (
+    <Tooltip title={
+        <>
```
annnag, line 136: What's <></> for?

react/src/pages/analysis/Analysis.tsx
```diff
@@ -152,48 +199,96 @@ export default function Analysis() {
...
                         pageSize: 10,
                         filtering: true,
-                        search: false
+                        search: false,
+                        padding: 'dense'
                     }}
                     actions={[
                         {
-                            icon: VisibilityIcon,
+                            icon: Visibility,
```
annnag, line 232: Could add some padding to the left of this icon, it is pretty close to the edge now.

annnag (changes requested):

react/src/pages/analysis/Analysis.tsx
frewmack: I borrowed code from here: https://material-ui.com/components/tooltips/#CustomizedTooltips.tsx
The `<> ... </>` is a [React Fragment](https://reactjs.org/docs/fragments.html). React render functions usually require that you provide a single JSX element to render, so Fragments work like a single parent for multiple child elements that also doesn't affect the style or anything. I technically didn't need it here since there's only 1 child element, but it lets us add more in the future if needed.

annnag: Okay, I see.

react/src/pages/analysis/Analysis.tsx
```diff
@@ -224,8 +227,7 @@ export default function Analysis() {
                     options={{
                         pageSize: 10,
                         filtering: true,
-                        search: false,
-                        padding: 'dense'
```
annnag: I think we want to keep `padding: dense,` beacause we are able to see more rows with this parameter. For the participant table on dashboard, I added `copyIcon: {
        marginLeft: theme.spacing(1),
    }`to style the left most action button so the table looks more symmetrical with the extra margin. Maybe you can also add this to the visibility icon.

annnag (changes requested):

annnag (approved):

kevinlul (approved):
LGTM

PR #58, opened 2020-10-01
React: switch dashboard to show participant table with copy buttons for combined codenames

annnag:
Fixed the participant table and added copy button for copy-pasting codenames.


react/package.json
```diff
@@ -16,6 +16,7 @@
     "@types/react-router": "^5.1.8",
     "@types/react-router-dom": "^5.1.3",
     "@types/react-table": "^7.0.23",
+    "copy-to-clipboard": "^3.3.1",
```
kevinlul, line 19: What's the reasoning behind introducing another dependency instead of directly using the native [synchronous](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) or [asynchronous](https://developer.mozilla.org/en-US/docs/Web/API/Clipboard/writeText) APIs?

react/src/pages/dashboard/ParticipantTable.tsx
```diff
@@ -0,0 +1,98 @@
+import React, { useState } from 'react';
+import { makeStyles } from '@material-ui/core/styles';
+import { Chip, IconButton } from '@material-ui/core';
+import MaterialTable, { MTableToolbar } from 'material-table';
+import { Cancel } from '@material-ui/icons';
+import FileCopyIcon from '@material-ui/icons/FileCopy';
```
kevinlul, line 6: combine import into the destructured import above and put it below the other `@material-ui` import

react/src/pages/dashboard/ParticipantTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+        }else{
+            newTypes[type] = 1;
+        }
+    })
+    return newTypes
+}
+
+export default function ParticipantTable() {
+    const classes = useStyles();
+    const [centre, setCentre] = useState("");
```
kevinlul, line 38: Better variable name would be `participantType` or just `filter` now.

react/src/pages/dashboard/Note.tsx
```diff
@@ -0,0 +1,16 @@
...
+}
+
+export default function Note({ rowData }: NoteProps) {
+    
+    return (
+        <Typography>
+            {rowData.note}
+        </Typography>
+    )
+}
```
kevinlul, line 16: Not quite sure why a whole component in a separate file was created for this?

react/src/pages/dashboard/ParticipantTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+                                <Chip label="Sibling" clickable className={classes.chip} onClick={() => setCentre("Sibling")} />
+                                <IconButton className={classes.chip} onClick={() => setCentre("")}> <Cancel /> </IconButton>
+                            </div>
+                        </div>
+                    ),
+                }}
+                actions={[
+                    {
+                        icon: () => <FileCopyIcon className={classes.copyIcon}/>,
+                        tooltip: 'Copy Participant Codename_Family Codename',
```
kevinlul, line 86: "Copy combined codename"

react/src/pages/dashboard/ParticipantTable.tsx
```diff
@@ -0,0 +1,98 @@
...
+                        onClick: CopyToClipboard,
+                    },
+                ]}
+                localization={{
+                    header: {
+                        actions: '',
```
kevinlul, line 92: Maybe add a comment that this hides the default "Action" label for the leftmost column?

kevinlul (changes requested):
Visually this looks really good! I just have some questions about the implementation.

react/package.json
annnag: With this package, I only need to write one line of code: `copy(toCopy)`.
Without it, I will need to create a DOM object in order to select and copy something, so there would be more code: 
`const dummy = document.createElement("input");`
           ` document.body.appendChild(dummy);`
       `     dummy.setAttribute('value', toCopy);`
          `  dummy.select();`
         `   document.execCommand("copy");`
          `  document.body.removeChild(dummy);`
They both works, so I can switch to use the native API if you think it's better.

kevinlul: That is pretty awful but because we don't care about IE at all, I think you can use the [newer async API](https://caniuse.com/mdn-api_clipboard_writetext) that's well-supported and also one line:
```js
await navigator.clipboard.writeText("foo");
```

annnag:
I noticed that after adding filter by predefined values, the chip buttons no longer work. I will merge master after I fix this problem.

react/src/pages/dashboard/ParticipantTable.tsx
```diff
@@ -0,0 +1,106 @@
...
+    const newTypes: {[key: string]: number} = {};
+    types.map(type => {
+        if(newTypes.hasOwnProperty(type)){
+            newTypes[type] += 1;
+        }else{
+            newTypes[type] = 1;
+        }
+    })
+    return newTypes
+}
```
kevinlul, line 29: There was a type warning because the return value of map is meant to be used (as opposed to forEach) so this should be a forEach loop or using Array.reduce:
```suggestion
const organizeDatasetTypes = (types: string[]) => types.reduce<Hash>(
    (newTypes, type) => {
        if (newTypes[type]) {
            newTypes[type] += 1;
        } else {
            newTypes[type] = 1;
        }
        return newTypes;
    }, Object.create(null));
```

kevinlul (changes requested):

PR #59, opened 2020-10-02
Retrieve usernames for all analyses in one query with three joins

kevinlul:


Madelinehazel (approved):

delvinso (approved):
nice, so this only performs one query! :thumbsup:

Issue #45, opened 2020-09-29
POST /api/_bulk


kevinlul:
Adds each dataset described in each row of the CSV or element of the JSON array, creating missing tissue samples, participants, and families if missing

Input:
- CSV ~~or JSON array~~ (distinguish based on Content-Type)

Output:
- Array of all datasets added in the same format as GET /api/datasets

Example:
- CSV row: participant_codename,sex,participant_type,affected,tissue_sample_type,family_codename,notes
- Create the family if it does not exist already
- Create the participant under the family if it does not exist already
- Create a tissue sample of that type if it does not exist already
- Add a dataset derived from that tissue sample
- Set the created and updated fields as relevant
- Fail if input has ANY inconsistencies (disagreement over participant_type and similar errors from denormalization)

PR #60, opened 2020-10-02
React: add CSV uploads via react-file-drop library

frewmack:
The general layout is in place for these upload forms. 
TODO: 
- ~~[ ] The fields need to align with those in the design document~~ Move to new branch
- [x] Actual upload depends via #45


kevinlul (commented):
Noting some tasks for this branch

- [x] merge or rebase onto latest master
- [ ] get rid of the link to existing dataset in the modals
- [ ] enumerations can be fetched from the backend now but that only relates to the manual portion

more to come

frewmack:
The file upload should work as intended. The datasets page pulls uploaded records when you upload.

Currently there is no feedback on success or error. I'm thinking of adding a reusable Snackbar component to utils and adding action feedback in a separate branch for all user controls.

frewmack:
Moving the manual form entry issue to a separate branch. After the meeting with CHEO, I think it's better that we take extra care in making the manual form. The CSV upload works as-is.

kevinlul (commented):
Is it intended that the drop area expands when dragged over?

frewmack:
> 
> 
> Is it intended that the drop area expands when dragged over?

Yes. The button needs to 'catch' the file, so it expands to make that easier.

kevinlul (approved):

PR #63, opened 2020-10-06
Flask: GET analysis by updated date

### LINKED ISSUE #53: Add parameter to GET analysis endpoint to query by updated date
Requires #18

Madelinehazel:
I'm not sure that Flask has a built-in method for verifying a datetime in the URL (i.e. @app.route('/api/analyses/<datetime:updated_date>', methods=['GET'], endpoint='analyses_list') returns an error, raise LookupError("the converter %r does not exist" % converter_name) LookupError: the converter 'datetime' does not exist). May need a custom convert? E.g. https://stackoverflow.com/questions/31669864/date-in-flask-url

Testing: curl http://localhost:3000/api/analyses/2020-08-05

Closes #53


delvinso:
yeah, I think a custom converter is the best bet: from the flask documentation only strings, int, float, or paths are automatically converted: https://exploreflask.com/en/latest/views.html

kevinlul:
Echoing what Delvin said, you'll need to provide the converting logic from strings to datetime (like `strptime`) for the time argument, or only accept Unix timestamps. The parameter should be in the query string (e.g. `/api/analyses?since=2020-05-01`) instead of part of the path, to avoid suggesting that the date labels an object in the collection.

flask/app/__init__.py
```diff
@@ -21,3 +21,4 @@
 from app.manage import *
 from app.routes import *
 from app.buckets import *
+from app.util import *
```
kevinlul, line 24: This no longer exists!

flask/app/manage.py
```diff
@@ -1,11 +1,13 @@
...
+from datetime import datetime
 
 @app.cli.command('add-default-admin')
 def add_default_admin():
     if len(db.session.query(models.User).all()) == 0:
         default_admin = models.User(
             username=app.config.get('DEFAULT_ADMIN'),
-            email=app.config.get('DEFAULT_ADMIN_EMAIL')
+            email=app.config.get('DEFAULT_ADMIN_EMAIL'),
+            last_login= datetime.now()
```
kevinlul, line 10: Good catch, but drop the space after the `=`

flask/app/routes.py
```diff
@@ -12,6 +12,8 @@
 
 from app import app, db, login, models
 
+app.url_map.converters['date'] = util.DateConverter
+
```
kevinlul, line 16: This no longer exists!

flask/app/routes.py
```diff
@@ -277,10 +279,17 @@ def participants_list():
 @app.route('/api/analyses', methods=['GET'], endpoint='analyses_list')
 @login_required
 def analyses_list():
+    since_date = request.args.get('since', default='0001-01-01')
+    try:
+        since_date = datetime.strptime(since_date, "%Y-%m-%d").date()
+    except:
+        return 'Malformed query date', 401
```
kevinlul, line 286: Status code should be 400

flask/app/routes.py
```diff
@@ -277,10 +279,17 @@ def participants_list():
 @app.route('/api/analyses', methods=['GET'], endpoint='analyses_list')
 @login_required
 def analyses_list():
+    since_date = request.args.get('since', default='0001-01-01')
+    try:
+        since_date = datetime.strptime(since_date, "%Y-%m-%d").date()
```
kevinlul, line 284: This works great! However, we want to query with more granularity than just the date (sorry if the requirement was unclear). Can the time be included as well?

kevinlul (changes requested):

Madelinehazel:
Query string must now contain time, ex:
`curl http://localhost:3000/api/analyses?since=2020-08-03T01:00:00`

@kevinlul does this comply with conventions for representing datetimes? Should the timezone be included?

kevinlul:
Timezone should be included so the timestamp is an ISO standard string.

Madelinehazel:
Added timezone handling, example query: 
`curl http://localhost:3000/api/analyses?since=2020-08-03T00:00:00-04:00`

where -04:00 represents the UTC offset for time in Toronto with daylight savings

kevinlul:
Sorry, I looked into this a little more to see how this might interoperate with the frontend, and I think using the standard library [ISO formatter](https://docs.python.org/3/library/datetime.html#datetime.datetime.fromisoformat) will achieve better results.

kevinlul (approved):
LGTM

PR #64, opened 2020-10-06
Adding month of birth to participant table

helrick:
Using datetime as suggested by @kevinlul to take advantage of date checking functionality in MySQL


flask/app/models.py
```diff
@@ -69,6 +69,7 @@ class Participant(db.Model):
     sex: Sex = db.Column(db.Enum(Sex), nullable=False)
     # Sample.SampleType
     participant_type: ParticipantType = db.Column(db.Enum(ParticipantType), nullable=False)
+    month_of_birth: str = db.Column(db.DateTime)
```
kevinlul, line 72: Just use [`db.Date`](https://docs.sqlalchemy.org/en/13/core/type_basics.html#sqlalchemy.types.Date) since the time is irrelevant. The Python typing should be `datetime.date`

kevinlul (commented):
Is it possible to define a check constraint limiting the day of the month to 1 as discussed? I think in SQL it would be like `CHECK(DAY(month_of_birth) = 1)` but I'm not confident.

- [Check constraints in SQLAlchemy](https://docs.sqlalchemy.org/en/13/core/constraints.html#sqlalchemy.schema.CheckConstraint)
- [Check constraints in MySQL](https://dev.mysql.com/doc/refman/8.0/en/create-table-check-constraints.html)
- https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_day

flask/app/models.py
helrick: resolved in d86688a

helrick:
> Is it possible to define a check constraint limiting the day of the month to 1 as discussed? I think in SQL it would be like `CHECK(DAY(month_of_birth) = 1)` but I'm not confident.
> 
> * [Check constraints in SQLAlchemy](https://docs.sqlalchemy.org/en/13/core/constraints.html#sqlalchemy.schema.CheckConstraint)
> * [Check constraints in MySQL](https://dev.mysql.com/doc/refman/8.0/en/create-table-check-constraints.html)
> * https://dev.mysql.com/doc/refman/8.0/en/date-and-time-functions.html#function_day

This isn't possible through SQLAlchemy as far as I can tell (https://docs.sqlalchemy.org/en/13/core/constraints.html#sqlalchemy.schema.CheckConstraint says that "some databases do not actively support check constraints such as MySQL."). However, I was able to enable this by manually editing the flask migration file and adding the CHECK constraint MySQL command in d86688a

kevinlul:
You can specify literal SQL. I believe the talk about MySQL not supporting check constraints is for versions before 8.0.16, but we are using the latest.
image

helrick:
> You can specify literal SQL. I believe the talk about MySQL not supporting check constraints is for versions before 8.0.16, but we are using the latest.
> image

I tried this by making the models.py definition `month_of_birth: datetime.date = db.Column(db.Date, CheckContraint("(DAY(month_of_birth) = 1)"))` and also tried wrapping the DAY in a CHECK() call. Neither of those added check to the MySQL database though. The only way I could figure out how enable it was through pure MySQL in the migration file.

kevinlul (approved):

PR #66, opened 2020-10-07
Add utils.tsx

### LINKED ISSUE #65: Refactor some duplicate code introduced with #26 and #58
1. SlideUpTransition
2. Hash type
3. Conversion function from array to Hash

kevinlul:
Refactor SlideUpTransition from AnalysisInfoDialog and CancelAnalysisDialog for reuse
Refactor countArray and toKeyValue from ParticipantTable and DatasetTable for reuse

Closes #65

### DIFF
--- a/react/src/pages/utils.tsx
+++ b/react/src/pages/utils.tsx
@@ -1,0 +1,29 @@
+import React, { forwardRef, ReactElement, Ref } from "react";
+import { Slide } from "@material-ui/core";
+import { TransitionProps } from "@material-ui/core/transitions";
+
+export const SlideUpTransition = forwardRef((
+    props: TransitionProps & { children?: ReactElement },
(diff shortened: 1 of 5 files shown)


react/src/pages/utils.tsx
```diff
@@ -0,0 +1,29 @@
+import React, { forwardRef, ReactElement, Ref } from "react";
+import { Slide } from "@material-ui/core";
+import { TransitionProps } from "@material-ui/core/transitions";
+
+export const SlideUpTransition = forwardRef((
```
annnag, line 5: Not sure if this transition is necessary, because all the other pages use the default transition.

kevinlul: Yeah, it's only used in the two modals on the analysis page, probably because of when they were written/copy-pasted.

PR #67, opened 2020-10-08
Fix hardcoded database name in migration

kevinlul:
Accidentally introduced in #64 resulting in an OperationalError indicating missing permissions for the table


helrick:
lgtm! good, catch 👌

helrick (approved):

PR #69, opened 2020-10-09
React: add participant detail modal on clicking dashboard rows

### LINKED ISSUE #42: Dashboard modal: view details wireframe
Can be triggered by clicking a row or an action button involving the row

Show in modal:
- all columns (created, updated)
- owned samples
- owned datasets
- analyses involved
- other family members

Requires #40

annnag:
Added participant detail panel. (Only participant AA0001 has some data)
Closes #42


react/src/pages/dashboard/DatasetAccordion.tsx
```diff
@@ -0,0 +1,67 @@
...
+                        {getInfo("Created By", dataset.createdBy.toString())}
+                        {getInfo("Update Time", dataset.updated)}
+                        {getInfo("Updated By", dataset.updatedBy.toString())}
+                        {getInfo("Discriminator", dataset.discriminator)}
+                    </Grid>
+                </Grid>
+            </AccordionDetails>
+        </Accordion>
+    )
+}
```
kevinlul, line 50: These already almost qualify as React functional components so it would be nice to name them like a component instead of `getX`, but it means that all parameters will have to be destructured from a single props arguments. Then you would be able to use JSX syntax.

react/src/pages/dashboard/DatasetTypes.tsx
```diff
@@ -1,14 +1,26 @@
...
 interface DatasetTypesProps {
     datasetTypes: {[key: string]: number},
 }
 
 export default function DatasetTypes({ datasetTypes }: DatasetTypesProps) {
+    const classes = useStyles();
+
+    const getChip = (type: string, number: number) => {
+        return <Chip className={classes.chip} label={type} avatar={<Avatar>{number}</Avatar>} />
+    }
```
kevinlul, line 19: Same thing here, although this is only used in one place so it might not warrant the separation at all.

kevinlul (approved):

PR #85, opened 2020-10-14
Flask: add GET /api/datasets/:id

### LINKED ISSUE #76: GET /api/datasets/:id
Retrieve just the dataset with the specified primary key, but also include an array of analyses it is involved in and data for the owning tissue sample.

Madelinehazel:
closes #76

### DIFF
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -319,16 +319,10 @@ def pipelines_list():
     return jsonify(db_pipelines)
 
 
-@app.route('/api/datasets', defaults={'id': None}, methods=['GET'], endpoint='datasets_list')
-@app.route('/api/datasets/<int:id>', methods=['GET'], endpoint='datasets_list')
+@app.route('/api/datasets', methods=['GET'], endpoint='datasets_list')
 @login_required
-def datasets_list(id:int):
-    if id:
-        db_datasets = db.session.query(models.Dataset).filter_by(dataset_id=id)
-    else:
-        db_datasets = db.session.query(models.Dataset)
-        
-    db_datasets = db_datasets.options(
+def datasets_list():
+    db_datasets = db.session.query(models.Dataset).options(
             joinedload(models.Dataset.tissue_sample).
             joinedload(models.TissueSample.participant).
             joinedload(models.Participant.family)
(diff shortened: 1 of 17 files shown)


kevinlul (changes requested):
For GET `/api/datasets/:id`, it shouldn't just return the same thing as the main one. Since we're fetching an individual object, it should just return that object rather than a single-element array. There were also some additional requirements in the issue that make this endpoint slightly different -- joining with analyses and all metadata for the tissue sample.

kevinlul:
Can I take over this branch to incorporate changes from #91 ?

Madelinehazel:
> Can I take over this branch to incorporate changes from #91 ?

Go for it

flask/app/routes.py
```diff
@@ -356,6 +370,34 @@ def datasets_list():
...
+@app.route('/api/datasets/<int:id>', methods=['GET'])
+@login_required
+def get_dataset(id: int):
+    dataset = db.session.query(models.Dataset).filter_by(dataset_id=id).options(
+        joinedload(models.Dataset.analyses),
+        joinedload(models.Dataset.tissue_sample).
+        joinedload(models.TissueSample.participant).
+        joinedload(models.Participant.family)
+     ).one()
+    if not dataset:
```
delvinso, line 382: What are the conditions for this? Requesting an ID not found in the table doesn't trigger this as it throws an error on the query (`sqlalchemy.orm.exc.NoResultFound: No row was found for one()`)

kevinlul: I was using the wrong method -- should be fixed now.

flask/app/routes.py
```diff
@@ -356,6 +370,34 @@ def datasets_list():
...
+@login_required
+def get_dataset(id: int):
+    dataset = models.Dataset.query.filter_by(dataset_id=id).options(
+        joinedload(models.Dataset.analyses),
+        joinedload(models.Dataset.tissue_sample).
+        joinedload(models.TissueSample.participant).
+        joinedload(models.Participant.family)
+     ).one_or_none()
+    if not dataset:
+        return 'Not Found', 404
```
delvinso, line 383: Could this return the ID that wasn't found or would it be exposing too much information?

kevinlul: The ID is in the URL parameter `/api/datasets/123413`

delvinso: That's true, was thinking it would be more for confirmation that the ID requested was the one actually queried

delvinso (approved):

PR #91, opened 2020-10-19
Flask: fix PATCH /api/analyses/:id to accept string usernames

kevinlul:
The general serialization problem will be a separate issue


frewmack (approved):
Looks good to me!

delvinso (approved):
looks good - a clean solution!

Issue #54, opened 2020-09-29
Fetch analysis endpoint by last signed in time to populate notification panel


kevinlul:
Endpoint implemented: `GET /api/analyses?since=ISO_TIMESTAMP`

Requires #35, #36, #53

kevinlul:
Closed by #92 in c5a1929e150fa448c031975af6b4ff666d8af3b5

PR #92, opened 2020-10-19
React: new notification panel & connect it to backend

### LINKED ISSUE #83: Notification panel: reduce size of see detail buttons, try align right
Experiment with ways to fill whitespace. May be blocked by #54, #78

annnag:
Removed the notification panel on the dashboard and replaced it with a popover on the top right corner. 

Closes #83, #54


react/src/pages/NotificationPopover.tsx
```diff
@@ -0,0 +1,99 @@
...
+        setAnchorEl(event.currentTarget);
+    };
+    const handlePopoverClose = () => {
+        setAnchorEl(null);
+    };
+    useEffect(() => {
+        const lastLoginISO = new Date(lastLoginTime).toISOString().slice(0, -1);
+        fetch(`/api/analyses?since=${lastLoginISO}`).then(async response => {
+            if (response.ok) {
+                setAnalyses((await response.json()) as AnalysisRow[]);
```
kevinlul, line 51: There's another step after fetching from https://github.com/ccmbioinfo/ST2020/blob/master/react/src/pages/analysis/Analysis.tsx#L177 to fit the `AnalysisRow` type.

react/src/pages/NotificationPopover.tsx
```diff
@@ -0,0 +1,99 @@
...
+            } else {
+                console.error(`GET /api/analyses?since=ISO_TIMESTAMP failed with ${response.status}: ${response.statusText}`);
+            }
+        });
+    }, []);
```
kevinlul, line 56: Add `lastLoginTime` to the dependency array. Between this and the other issue, it's causing a freeze and crash when actual data is returned from the endpoint. To test with a login timestamp, you can try something like this in your shell:
```bash
docker exec -it st2020_app_1 python

from datetime import datetime
from app import models, db
user = models.User.query.get(1)
user.last_login = datetime(2020, 1, 1)
db.session.commit()
```

kevinlul (changes requested):

kevinlul (approved):
LGTM

PR #93, opened 2020-10-19
React: simplify date format in dataset & analysis table 

### LINKED ISSUE #82: Reduce updated date in tables to just YYYY-MM-DD
Do this as part of any work on these pages
- [ ] dashboard (participants)
- [x] datasets
- [x] analysis

annnag:
Reduced datetime format to MM/DD/YYYY in dataset table and analysis table. (there is no update date column in the participant table)
MM/DD/YYYY is used instead of YYYY-MM-DD because there is a built-in method that can do this.

Closes #82


kevinlul:
Probably superseded by #94 for now

PR #94, opened 2020-10-20
ISO Date formatting (#82) and editable cells (#79)

### LINKED ISSUE #79: Editing data in tables: try cell edits vs row edits

### LINKED ISSUE #82: Reduce updated date in tables to just YYYY-MM-DD

frewmack:
* Change rendering of datetime strings to ISO format (closes #82)
* Add editable cells to Analysis table (closes #79) (note: changes are local; mutations are on different branch)
* Enable dense padding for dataset, dashboard, analysis tables

### DIFF
--- a/react/src/pages/datasets/DatasetTable.tsx
+++ b/react/src/pages/datasets/DatasetTable.tsx
@@ -2,7 +2,7 @@ import React, { useEffect, useState } from 'react';
 import { makeStyles, Chip, IconButton, TextField } from '@material-ui/core';
 import { PlayArrow, Delete, Cancel } from '@material-ui/icons';
 import MaterialTable, { MTableToolbar } from 'material-table';
-import { toKeyValue, KeyValue } from "../utils";
+import { toKeyValue, KeyValue, formatDateString } from "../utils";
 import AnalysisRunnerDialog, { Pipeline } from './AnalysisRunnerDialog';
 
 export interface Dataset {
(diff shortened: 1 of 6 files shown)


react/src/pages/datasets/DatasetTable.tsx
```diff
@@ -131,6 +133,36 @@ export default function DatasetTable() {
                         }
                     }
                 }}
+                cellEditable={{
```
annnag, line 136: Did you forget to add `emptyValue: '<empty>'` to the editable columns or you did that on purpose?

frewmack: My mistake. I'll fix that shortly.

kevinlul (changes requested):
This looks great! I had some trouble editing this field. I think it's because no other notes were as long, so when the edit UI came up, the width of the column was reduced to the width of the other much shorter notes.

Before click
Screen Shot 2020-10-20 at 15 35 02

After click
Screen Shot 2020-10-20 at 15 34 53

frewmack:
> 
> 
> This looks great! I had some trouble editing this field. I think it's because no other notes were as long, so when the edit UI came up, the width of the column was reduced to the width of the other much shorter notes.
> 
> Before click
> Screen Shot 2020-10-20 at 15 35 02
> 
> After click
> Screen Shot 2020-10-20 at 15 34 53

I tried playing around with the column width and other props, but it doesn't look like there is a workaround for this. I also noticed that word-wrapping just doesn't happen if the column is cell-editable.

kevinlul (approved):

PR #95, opened 2020-10-20
React: hook up analysis table rows, set assignee

### LINKED ISSUE #80: Hook up analysis page mutations to PATCH endpoints
Frontend for mutations from #46, #47, #49 were merged in #62
Backend endpoint merged in #37

- [x] Table rows
- [x] Set assignee
- [ ] Cancel analysis
- [ ] Run/finish analysis

frewmack:
For #80, however we may need a new PATCH endpoint for patching multiple analyses in one request (for set assignee, start analysis, cancel analysis). Only option as of now is to send 1 PATCH request for every row selected, which isn't great.

### DIFF
--- a/react/src/pages/analysis/Analysis.tsx
+++ b/react/src/pages/analysis/Analysis.tsx
@@ -446,7 +446,6 @@ export default function Analysis() {
                         onRowUpdate: (newData, oldData) =>
                             new Promise((resolve, reject) => {
                                 // TODO: Send PATCH here for editing notes or path
-
                                 const dataUpdate = [...rows];
                                 // find the row; assume analysis_id is unique
                                 const index = dataUpdate.findIndex((row, index, obj) => {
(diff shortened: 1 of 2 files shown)


react/src/pages/analysis/Analysis.tsx
```diff
@@ -446,7 +446,6 @@ export default function Analysis() {
                         onRowUpdate: (newData, oldData) =>
                             new Promise((resolve, reject) => {
                                 // TODO: Send PATCH here for editing notes or path
-
                                 const dataUpdate = [...rows];
                                 // find the row; assume analysis_id is unique
                                 const index = dataUpdate.findIndex((row, index, obj) => {
```
kevinlul, line 451: I didn't think of this before with duplicating the state block but this part should be much more clear if we go all in on the functional style like here https://github.com/ccmbioinfo/ST2020/blob/master/react/src/pages/datasets/DatasetTable.tsx#L124. Use `Array.map` to replace just the row that changed but return a different array.

kevinlul (changes requested):
Looks to be working great! Since the entire `onRowUpdate` returns a promise, it might be more clear to rewrite it as an async function

kevinlul:
We talked about the 1 PATCH per row thing in a standup so we'll have to see if it becomes a performance problem down the road. I just noticed that this PR only affects the table rows and not the action buttons?

kevinlul (approved):

kevinlul (changes requested):
async doesn't work like you expect with Array.forEach, so to fire off the requests in sequence and then update the frontend state, you'll have to use a for..of loop.

Either we update state in one shot at the end (and we should adopt Array.map here too) or we update each time we get a response, but this seems to do both.

kevinlul (approved):

PR #97, opened 2020-10-21
React: provide user feedback via Snackbars via notistack library

frewmack:
Uses a library called `notistack` to add Snackbar messages to provide user feedback by simply calling a function.


react/src/App.tsx
```diff
@@ -1,9 +1,17 @@
...
+import { SnackbarKey, SnackbarProvider } from 'notistack';
 import React, { useState, useEffect } from 'react';
 import { createMuiTheme, ThemeProvider } from '@material-ui/core';
 
 import LoginForm from './pages/Login';
 import Navigation from './pages/Navigation';
 
+const notistackRef = React.createRef<any>();
+const onClickDismiss = (key: SnackbarKey) => () => {
+    notistackRef.current.closeSnackbar(key);
```
kevinlul, line 12: I believe the typing to `createRef` is `SnackbarProvider` and then you can do `current!.closeSnackbar` to assert not null since this is only called from the IconButton click?

kevinlul (changes requested):

react/src/App.tsx
```diff
@@ -1,9 +1,17 @@
+import { IconButton } from '@material-ui/core';
+import { Close } from '@material-ui/icons';
+import { SnackbarKey, SnackbarProvider } from 'notistack';
```
kevinlul, line 3: These should come after the React import!

react/src/pages/analysis/Analysis.tsx
```diff
@@ -9,6 +9,7 @@ import CancelAnalysisDialog from './CancelAnalysisDialog';
 import AnalysisInfoDialog from './AnalysisInfoDialog';
 import AddAnalysisAlert from './AddAnalysisAlert';
 import SetAssigneeDialog from './SetAssigneeDialog';
+import { useSnackbar } from 'notistack';
```
kevinlul, line 12: Move the notistack imports above our own

kevinlul (changes requested):
Just a few more things. This looks great!

react/src/App.tsx
kevinlul: There's already a `@material-ui/core` import in this file. Could you combine them please?

kevinlul (approved):
LGTM

PR #101, opened 2020-10-21
React: tidy imports, fix compiler warnings

frewmack:
Removed some unused dummy-data functions now that we're getting data remotely.


react/src/pages/upload/FilesTable.tsx
```diff
@@ -1,9 +1,7 @@
 import React, { useState, useEffect }  from 'react';
 import { makeStyles } from '@material-ui/core/styles';
-import Chip from '@material-ui/core/Chip';
+import { Chip } from '@material-ui/core';
```
kevinlul, line 3: You can combine the makeStyles import into this too.

kevinlul (approved):

PR #105, opened 2020-10-22
React: move up analysis info dialog with notifications; remove /analysis/:id route

frewmack:
The NotificationPopover GETs analyses that have changed since last login. Instead of loading the analysis page and requesting all of the analyses, Notifications can now create the info dialog themselves on click using the information they already have.


react/src/pages/Notification.tsx
```diff
@@ -50,36 +51,38 @@ function createNotification(
...
 }
 
-export default function Notification({ analysis } : NotificationProps) {
+export default function Notification({ analysis, onClick } : NotificationProps) {
     const classes = useStyles();
     const [hover, setHover] = useState(false);
     const {analysis_id, msg, severity} = getNotificationInfo(analysis)!;
 
     return (
-        <Button className={classes.button} component={Link} to={`/analysis/${analysis_id}`} onMouseEnter={() => setHover(true)} onMouseLeave={() => setHover(false)}>
```
annnag, line 73: `import { Link } from 'react-router-dom';` can be removed

annnag (changes requested):
Maybe the fake analyses can be removed from Analyses.tsx

kevinlul (approved):
LGTM

PR #106, opened 2020-10-23
Flask: add GET /api/analyses/:id

### LINKED ISSUE #87: GET /api/analyses/:id
Retrieve just the analysis with the specified primary key, but also include an array of datasets that it is running on, and optionally join to the pipeline table too. This will be used to populate the analysis detail popup.

kevinlul:
Closes #87


delvinso (approved):
Looks good to me

PR #109, opened 2020-10-26
React: Datasets modal; view details

### LINKED ISSUE #52: Datasets modal: view details wireframe
Show all columns, trigger via action button

Requires #26, #76 to be fully functional

frewmack:
Closes #52. Layout is similar to #69

### DIFF
--- a/react/src/pages/datasets/DatasetInfoDialog.tsx
+++ b/react/src/pages/datasets/DatasetInfoDialog.tsx
@@ -143,7 +143,7 @@ export default function DatasetInfoDialog({ dataset_id, open, onClose }: DialogP
                     </Grid>
                     </Collapse>
                     <Button onClick={() => {setMoreDetails(!moreDetails)}}>
-                        Show {moreDetails ? 'fewer' : 'more'} details
+                        {moreDetails ? "Hide" : "Show"} more details
                     </Button>
                     </>
                     }
@@ -151,7 +151,21 @@ export default function DatasetInfoDialog({ dataset_id, open, onClose }: DialogP
 
                 {sample &&
                 <Paper className={classes.paper} elevation={2}>
-                    Associated Sample goes here
+                   <Grid container spacing={2} justify="space-evenly" className={classes.grid}>
+                        <Grid item xs={6}>
+                            {getInfo('Sample ID', sample.sampleID)}
+                            {getInfo('Sample Type', sample.sampleType)}
+                            {getInfo('Extraction Date', sample.extractionDate)}
+                            {getInfo('Tissue Processing Protocol', sample.tissueProcessing)}
+                        </Grid>
+                        <Grid item xs={6}>
+                            {getInfo('Notes', sample.notes)}
+                            {getInfo('Created', sample.created)}
+                            {getInfo('Created By', sample.createBy)}
+                            {getInfo('Updated', sample.updated)}
+                            {getInfo('Updated By', sample.updatedBy)}
+                        </Grid>
+                    </Grid>
                 </Paper>
                 }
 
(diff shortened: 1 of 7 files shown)


react/src/pages/datasets/DatasetInfoDialog.tsx
```diff
@@ -0,0 +1,235 @@
...
+        val = value.toString();
+    else if (typeof value === 'string')
+        val = value;
+    else
+        val = "";
+
+    return (
+        <Typography variant="body1" gutterBottom><b>{subtitle}:</b> {val}</Typography>
+    );
+}
```
kevinlul, line 72: Like DialogHeader above this would be more idiomatic as a functional component.

react/src/pages/datasets/DatasetInfoDialog.tsx
```diff
@@ -0,0 +1,235 @@
...
+    function clickAnalysis(index: number) {
+        try {
+            // toggle
+            setShowAnalysis(
+                showAnalysis.map((val, i) => index === i ? !val : val)
+            );
+        } catch (error) {
+            console.error(error);
+        }
+    }
```
kevinlul, line 99: What is the purpose of the try-catch here?

react/src/pages/datasets/DatasetInfoDialog.tsx
```diff
@@ -0,0 +1,235 @@
...
+
+    useEffect(() => {
+        fetch('/api/datasets/'+dataset_id)
+        .then(response => response.json())
+        .then(data => {
+            let temp = Object.assign({}, { ...data, analyses: undefined, tissue_sample: undefined });
```
kevinlul, line 105: Why is Object.assign used here?

kevinlul (changes requested):
I really like how this looks and making the fields that might not always be present collapsible is a great idea!

react/src/pages/datasets/DatasetInfoDialog.tsx
frewmack: I think it was to catch an IndexError, but I changed the code inside the try block since then, so it's no longer necessary.

react/src/pages/datasets/DatasetInfoDialog.tsx
```diff
@@ -0,0 +1,235 @@
...
+            {onClose ? (
+                <IconButton aria-label="close" className={classes.closeButton} onClick={onClose}>
+                    <Close />
+                </IconButton>
+            ) : null}
+        </DialogTitle>
+    );
+});
+
+const getInfo = (subtitle: string, value: string | number | null) => {
```
annnag, line 60: This function is also used in ParticipantDetailDialog component in the dashboard page, so you could move it to utils.tsx

react/src/pages/datasets/DatasetInfoDialog.tsx
```diff
@@ -0,0 +1,235 @@
...
+                            {getInfo('Notes', dataset.notes)}
+                            {getInfo('Created', dataset.created)}
+                            {getInfo('Created By', dataset.created_by)}
+                            {getInfo('Updated', dataset.updated)}
+                            {getInfo('Updated By', dataset.updated_by)}
+                        </Grid>
+                    </Grid>
+                    <Collapse in={moreDetails}>
+                    <Grid container spacing={2} justify="space-evenly">
+                        <Grid item xs={6}>
```
annnag, line 143: Maybe you could make this a separate component? It is used a few times in this file and also in the participant dialog.

annnag (changes requested):

kevinlul (approved):
LGTM

PR #110, opened 2020-10-26
Add Prettier, Black, Pylint

### LINKED ISSUE #75: Add formatting plugins
Prettier for JavaScript/TypeScript
Black for Python

Include instructions for how to automatically integrate with editors. This should save a bit of time in code reviews.

kevinlul:
Formatting:
- install dev dependencies in a virtualenv in `flask/` and run `black app migrations tests`
- run `yarn format` in `react/`

[Ignoring `git blame` on the format commit](https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revltrevgt) (to do after merge if squashed)

Closes #75


react/src/pages/analysis/Analyses.tsx
```diff
@@ -448,25 +623,52 @@ export default function Analyses() {
...
+                                    />
+                                    <Chip
+                                        label="Cancelled"
+                                        clickable
+                                        className={classes.chip}
+                                        onClick={() => setChipFilter(PipelineStatus.CANCELLED)}
+                                    />
+                                    <IconButton onClick={() => setChipFilter("")}>
+                                        {" "}
+                                        <Cancel />{" "}
```
frewmack, line 664: Only weirdness I could find on the frontend. Not really an issue to address right now, but we should double-check our code after formatting on our own branches.

frewmack (approved):
Looks good to me!

react/src/pages/analysis/Analyses.tsx
kevinlul: It looks like this happened because there was a space before and after when this was one line, and sometimes that's significant when rendered. I'll get rid of it.

helrick (approved):
LGTM!

annnag (approved):
Looks great!

Issue #119, opened 2020-10-27
Standardize user interface across participants, datasets, analyses


kevinlul:
Come up with a design first before implementing

- [x] do we need family and tissue sample views?
- [x] cell vs row edits
- [x] action column label
- [x] view details button vs clicking the row
- [x] some fonts
- [x] material-ui tables vs material-table in dialogs

PR #127, opened 2020-10-28
React: hook up analysis details to backend

### LINKED ISSUE #88: Hook up analysis details modal to backend
Requires #87

Remove the irrelevant labels and just show the name and version of the pipeline with the other metadata.

Replace samples table with a table of datasets.

frewmack:
Closes #88 
Layout is very similar to dialog for #109, and so this should be revisited for #119

### DIFF
--- a/react/src/pages/AnalysisInfoDialog.tsx
+++ b/react/src/pages/AnalysisInfoDialog.tsx
@@ -1,14 +1,20 @@
 import React, { useEffect, useState } from 'react';
-import { createStyles, makeStyles, Theme, withStyles, WithStyles } from '@material-ui/core/styles';
-import Dialog from '@material-ui/core/Dialog';
+import {
+    List,
+    ListItemIcon,
+    ListItem,
+    Paper,
+    ListItemText,
+    Collapse,
+    Dialog,
+    IconButton,
+    Typography
+} from '@material-ui/core';
 import MuiDialogTitle from '@material-ui/core/DialogTitle';
 import MuiDialogContent from '@material-ui/core/DialogContent';
-import IconButton from '@material-ui/core/IconButton';
-import CloseIcon from '@material-ui/icons/Close';
-import Typography from '@material-ui/core/Typography';
+import { createStyles, makeStyles, Theme, withStyles, WithStyles } from '@material-ui/core/styles';
+import { Dns, ExpandLess, ExpandMore, Close } from '@material-ui/icons';
 import { Analysis, Dataset, FieldDisplay, Pipeline } from './utils';
-import { List, ListItemIcon, ListItem, Paper, ListItemText, Collapse } from '@material-ui/core';
-import { Dns, ExpandLess, ExpandMore } from '@material-ui/icons';
 import { DatasetDetailSection } from './datasets/DialogSections';
 
 interface AlertInfoDialogProp {
@@ -44,7 +50,7 @@ const DialogTitle = withStyles(styles)((props: DialogTitleProps) => {
             <Typography variant="h6">{children}</Typography>
             {onClose ? (
                 <IconButton aria-label="close" className={classes.closeButton} onClick={onClose}>
-                    <CloseIcon />
+                    <Close />
                 </IconButton>
             ) : null}
         </MuiDialogTitle>
@@ -70,10 +76,6 @@ const useStyles = makeStyles(theme => ({
         margin: theme.spacing(1)
     }
 }));
-
-const pipelineParams: string[] = ["gnomAD_AF <= 0.01", "trio", "joint_genotyping", "denovo", "IMPACT_SEVERITY=HIGH",]
-const analyses: string[] = ["SNP", "INDELs",]
-export const annotations: string[] = ["OMIM 2020-07-01", "HGMD 2019-02-03", "snpEff 4.3T", "gnomAD v2.1.1"]
 
 export default function AnalysisInfoDialog({ analysis, open, onClose }: AlertInfoDialogProp) {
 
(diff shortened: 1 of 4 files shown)


annnag (commented):
Looks really good! One suggestion: space here could be reduced
Screen Shot 2020-10-29 at 10 46 11 AM

react/src/pages/AnalysisInfoDialog.tsx
```diff
@@ -56,46 +57,94 @@ const DialogContent = withStyles((theme: Theme) => ({
...
+    },
+    box: {
+        padding: theme.spacing(2),
+        margin: theme.spacing(1)
+    }
+}));
+
 const pipelineParams: string[] = ["gnomAD_AF <= 0.01", "trio", "joint_genotyping", "denovo", "IMPACT_SEVERITY=HIGH",]
 const analyses: string[] = ["SNP", "INDELs",]
 export const annotations: string[] = ["OMIM 2020-07-01", "HGMD 2019-02-03", "snpEff 4.3T", "gnomAD v2.1.1"]
```
kevinlul, line 80: These placeholders are no longer needed!

react/src/pages/utils.tsx
```diff
@@ -173,24 +173,56 @@ export function jsonToAnalyses(data: Array<any>): Analysis[] {
...
+        console.error(error);
+        return null;
+    }
+}
+
+/**
+ * Return whether this material-table row is checked / selected.
+ * If it is not a material-table row, return null.
+ */
+export function getIsRowSelected(row: any): boolean | null {
```
kevinlul, line 199: You can just call this `isRowSelected`

react/src/pages/utils.tsx
```diff
@@ -173,24 +173,56 @@ export function jsonToAnalyses(data: Array<any>): Analysis[] {
...
+    try {
+        let isChecked: boolean = row.tableData.checked;
```
kevinlul, line 201: You can use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) here (`?.` operator) and avoid the try-catch. If it's just a boolean then you can really just `return `row?.tableData?.checked`. This and `getRowIndex` might not need to be separate functions out here then if it boils down to that. I'm guessing `material-table` doesn't export `tableData` in its typings?

react/src/pages/utils.tsx
```diff
@@ -173,24 +173,56 @@ export function jsonToAnalyses(data: Array<any>): Analysis[] {
...
  * Returns a simple Typography JSX element for displaying "title: value".
  */
-export function FieldDisplay(props: { title: string, value: string | number | null }) {
+export function FieldDisplay(props: { title: string, value?: string[] | string | number | null, variant?: Variant }) {
     let val: string;
     if (typeof props.value === 'number')
         val = props.value.toString();
     else if (typeof props.value === 'string')
         val = props.value;
+    else if (typeof props.value === 'object' && props.value?.length !== undefined)
```
kevinlul, line 222: Use [Array.isArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray).

I think given the context that this value is used in to avoid `null` or `undefined` being displayed, you only need to care about the `null` and array cases since JSX will do the number to string for you below.

react/src/pages/utils.tsx
```diff
@@ -22,6 +22,7 @@ const months = [
 /*****   TYPINGS   *****/
 export type Counts = { [key: string]: number };
 export type KeyValue = { [key: string]: string };
+export type Variant = "button" | "caption" | "h1" | "h2" | "h3" | "h4" | "h5" | "h6" | "inherit" | "subtitle1" | "subtitle2" | "body1" | "body2" | "overline" | "srOnly";
```
kevinlul, line 25: I believe you can just use `Typography["variant"]` to get the type instead of duplicating it here.

react/src/pages/analysis/Analyses.tsx
```diff
@@ -346,7 +327,7 @@ export default function Analyses() {
                             tooltip: 'Add New Analysis',
                             position: 'toolbar',
                             isFreeAction: true,
-                            onClick: (event) => setDirect(true)
+                            onClick: (event, rowData) => setDirect(true)
```
kevinlul, line 330: Neither argument is used here

react/src/pages/AnalysisInfoDialog.tsx
```diff
@@ -1,14 +1,15 @@
...
 import MuiDialogContent from '@material-ui/core/DialogContent';
 import IconButton from '@material-ui/core/IconButton';
 import CloseIcon from '@material-ui/icons/Close';
 import Typography from '@material-ui/core/Typography';
-import DatasetTable from './analysis/DatasetTable';
-import { Analysis } from './utils';
-import ChipStrip from './analysis/ChipStrip';
+import { Analysis, Dataset, FieldDisplay, Pipeline } from './utils';
+import { List, ListItemIcon, ListItem, Paper, ListItemText, Collapse } from '@material-ui/core';
+import { Dns, ExpandLess, ExpandMore } from '@material-ui/icons';
```
kevinlul, line 11: Combine imports with above

kevinlul (changes requested):
Looking good!

react/src/pages/utils.tsx
frewmack: Yeah tableData is not exported. Also strangely if the row is unchecked, then the `checked` field is just undefined instead of false.

kevinlul: You can use `!!row?.tableData?.checked` to coerce to boolean then

kevinlul (approved):

PR #143, opened 2020-11-05
Flask:  DELETE /api/datasets/:id

### LINKED ISSUE #137: DELETE /api/datasets/:id
Admin-only. Delete the dataset if it is not being used in any analyses or fail with 422.

Create a new `datasets.py` and extract all datasets APIs here except the mixed PATCH function unless you're up for that.

Madelinehazel:
closes #137


delvinso:
Not sure if this is an issue but when testing the endpoint - a successful deletion is silent (doesn't return 'Updated') but the 204 does get sent to the logs. The other errors are thrown properly though so I'm not sure what's going on
```
 curl -X DELETE localhost:5000/api/datasets/4
# nothing returned here but returns 204 indicating success in logs
```
subsequent tries throw an error (as expected since it was successfully deleted)
```
 curl -X DELETE localhost:5000/api/datasets/4
Not Found%                  
```
```
curl -X DELETE localhost:5000/api/datasets/1
Dataset has analyses, cannot delete%           
```

Otherwise looks good to me!

delvinso (approved):

kevinlul:
204 means No Content meaning you shouldn't expect to see a response body.

delvinso:
Okay that makes sense, I was using error code 200  before so maybe that's why I was expecting the 'Updated' message on success

kevinlul (changes requested):
`datasets.py` isn't hooked up in `__init__.py`!

Madelinehazel:
> `datasets.py` isn't hooked up in `__init__.py`!

Forgot to add the file, fixed now!

kevinlul (approved):

PR #145, opened 2020-11-06
React: fix rendering failure when there are no datasets for a participant or tissue sample

kevinlul:
Instead of being set to `[]`, `dataset_types` would be undefined


annnag (approved):
👍

PR #146, opened 2020-11-06
Flask: DELETE /api/participants/:id

### LINKED ISSUE #135: DELETE /api/participants/:id
Admin-only. Delete the participant if it does not contain any tissue samples or fail with 422.

Create a new `participants.py` and extract all participants APIs here except the mixed PATCH function unless you're up for that.

Madelinehazel:
closes #135 

Added participant with participant_id=4  in manage.py for testing deletion of a participant with no associated tissue samples.

### DIFF
--- a/flask/app/__init__.py
+++ b/flask/app/__init__.py
@@ -39,5 +39,6 @@ def create_app(config):
         from . import analyses
         from . import families
         from . import datasets
+        from . import participants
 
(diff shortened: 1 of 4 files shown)


delvinso (approved):

kevinlul (approved):
LGTM

Issue #122, opened 2020-10-27
Add CI via GitHub Actions


kevinlul:
Some stuff needs research

- [x] Python lint and test
- [x] React lint and test
- [x] Build Docker images
- [x] [pull request labeller](https://github.com/actions/labeler)
- [ ] ~~Automatic deployment via SSH or Docker?~~
- [ ] Fix Dependabot-exclusive labels

Task points: 2

kevinlul:
Done for now, next step would be CD

PR #172, opened 2020-11-11
Flask: add MinIO administration module

### LINKED ISSUE #115: Programmatically create MinIO identities for permissions groups
Read [MinIO Go](https://github.com/minio/minio/blob/master/pkg/madmin/user-commands.go) code to figure out if we can programmatically create identities for Sample Tracker permission groups

kevinlul:
Closes #115

### DIFF
--- a/flask/tests/test_madmin.py
+++ b/flask/tests/test_madmin.py
@@ -1,0 +1,123 @@
+import pytest
+from app.madmin import MinioAdmin
+from conftest import TestConfig
+
+
+@pytest.fixture
+def mc():
+    return MinioAdmin(
+        endpoint=TestConfig.MINIO_ENDPOINT,
+        access_key=TestConfig.MINIO_ACCESS_KEY,
+        secret_key=TestConfig.MINIO_SECRET_KEY,
+    )
+
+
+def test_add_remove_users(mc: MinioAdmin):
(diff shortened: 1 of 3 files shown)


flask/tests/test_madmin.py
```diff
@@ -0,0 +1,136 @@
+import pytest
+from app.madmin import MinioAdmin, readonly_buckets_policy, readwrite_buckets_policy
```
SamarthVP, line 2: I can't seem to run these tests because `app` module is not found. How are you running this?

kevinlul: You have to do `python -m pytest` or `./utils/run.sh test`, assuming you installed dev dependencies in a virtualenv. For some reason, running pytest as a module works for module resolution but not `pytest`. If you're just using Docker then I'm working on getting tests up inside that on a separate branch for #122. Make sure that when you run these tests, you haven't added any custom users, groups, or policies to the MinIO instance.

SamarthVP: "For some reason, running pytest as a module works for module resolution but not `pytest`"
What on earth?!?
Well I ran it with `python -m pytest` and the tests passed so looks good

SamarthVP (approved):

PR #173, opened 2020-11-11
React Add participants: Checkbox cell for boolean headers

### LINKED ISSUE #157: Add participants: Checkbox cell for boolean headers (solved, affected)
Task points: 3

frewmack:
Closes #157

### DIFF
--- a/react/src/pages/utils/upload/DataEntryTable.tsx
+++ b/react/src/pages/utils/upload/DataEntryTable.tsx
@@ -1,6 +1,8 @@
 import React, { ReactNode, useEffect, useState } from "react";
 import {
     Box,
+    Button,
+    Checkbox,
     IconButton,
     makeStyles,
     Menu,
@@ -22,8 +24,15 @@ import {
 import { Autocomplete, createFilterOptions } from "@material-ui/lab";
 import { AddBoxOutlined, CloudUpload, Delete, LibraryAdd, ViewColumn } from "@material-ui/icons";
 import { DataEntryHeader, DataEntryRow } from "../typings";
-import { Option, toOption, getOptions as _getOptions, getColumns } from "./UploadUtils";
+import {
+    Option,
+    toOption,
+    getOptions as _getOptions,
+    getColumns,
+    booleanColumns,
+} from "./UploadUtils";
 import UploadDialog from "./UploadDialog";
+import { setProp } from "../functions";
 
 export interface DataEntryTableProps {
     data?: DataEntryRow[];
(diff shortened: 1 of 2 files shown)


react/src/pages/utils/upload/DataEntryTable.tsx
```diff
@@ -239,14 +249,51 @@ export default function DataEntryTable(props: DataEntryTableProps) {
...
+                value={toOption(props.row[props.col.field])}
+                options={props.getOptions(props.rowIndex, props.col)}
+                onEdit={props.onEdit}
+                aria-label={`enter ${props.col.title} row ${props.rowIndex}`}
+                disabled={props.disabled}
+            />
+        );
+    }
+
+    return cell;
```
kevinlul, line 286: I think you can just return out of each branch

react/src/pages/utils/upload/DataEntryTable.tsx
```diff
@@ -239,14 +249,51 @@ export default function DataEntryTable(props: DataEntryTableProps) {
...
+    getOptions: (rowIndex: number, col: DataEntryHeader) => Option[];
+    onEdit: (newValue: string | boolean) => void;
+    disabled?: boolean;
+}) {
+    let cell;
+
+    if (booleanColumns.includes(props.col.field)) {
+        cell = (
+            <CheckboxCell
+                value={props.row[props.col.field] as boolean}
```
kevinlul, line 269:
```suggestion
                value={!!props.row[props.col.field]}
```

react/src/pages/utils/upload/DataEntryTable.tsx
```diff
@@ -313,6 +360,25 @@ function DataEntryCell(
...
+    onEdit: (newValue: boolean) => void;
+    disabled?: boolean;
+}) {
+    function handleChange(event: React.ChangeEvent<HTMLInputElement>) {
+        props.onEdit(event.target.checked);
+    }
+
+    return (
+        <TableCell padding="checkbox">
+            <Checkbox checked={!!props.value} onChange={handleChange} />
```
kevinlul, line 377:
```suggestion
            <Checkbox checked={props.value} onChange={event => props.onEdit(event.target.checked)} />
```
This should shorten this component by a lot. `props.disabled` isn't used at all though.

kevinlul (changes requested):

react/src/pages/utils/upload/DataEntryTable.tsx
```diff
@@ -1,6 +1,8 @@
 import React, { ReactNode, useEffect, useState } from "react";
 import {
     Box,
+    Button,
```
kevinlul, line 4: This doesn't seem to be used anywhere. If it's needed for later then it's fine.

kevinlul (approved):
LGTM

annnag (commented):
Looks good! One thing I noticed is that there isn't a state/icon to represent null, so I wonder if it is necessary to have one? Or having true and false is enough for the users?

kevinlul:
There is an `indeterminate` state for mui checkboxes but I don't know how usable it is from a UX perspective.

frewmack:
Right now the state just assumes that `null`/`undefined` is the same as `false`. The `indeterminate` state for checkboxes is usually used on nested checklist when only some of the 'child' checkboxes are checked, so I'm a little worried that users may find it confusing. Example: 
image

helrick:
Would it be possible to convert the nullable boolean fields to dropdowns with a "Blank" option? In that case we could give them the three options that the database has and don't have to decide on a default value for these fields

frewmack:
> Would it be possible to convert the nullable boolean fields to dropdowns with a "Blank" option? In that case we could give them the three options that the database has and don't have to decide on a default value for these fields

I can add this as an additional component so that we can compare in the next standup. I'll also make a copy of the checkbox component where `indeterminate` state is used to see how it looks.

annnag (approved):

PR #175, opened 2020-11-11
Flask: Create MinIO unlinked files endpoint

### LINKED ISSUE #155: Backend API to query MinIO for unlinked files that the current user is allowed to access
Task points: 5

May be blocked by #115

SamarthVP:
closes #155

### DIFF
--- a/flask/app/buckets.py
+++ b/flask/app/buckets.py
@@ -42,30 +42,43 @@ def get_object_info():
 
 
 @app.route("/api/unlinked", methods=["GET"])
+@login_required
 def get_unlinked_files():
 
-    bucket_names = [bucket.name for bucket in minioClient.list_buckets()]
-    # match against buckets/groups the user should be seeing, remove those from the list
-    # note that bucket names are lower case to confrom with s3
+    # Get all minio bucket names
+    all_bucket_names = [bucket.name for bucket in minioClient.list_buckets()]
+
+    # Remove buckets the current user does not have access to assuming group_code.lower() == bucket name
+    user = (
+        db.session.query(models.User)
+        .filter_by(user_id=current_user.user_id)
+        .options(joinedload(models.User.groups))
+        .first()
+    )
+
+    valid_bucket_names = []
+    for group in user.groups:
+        code = group.group_code.lower()
+        if code in all_bucket_names:
+            valid_bucket_names.append(code)
 
     # Get all files in valid minio buckets
     all_files = []
-    for bucket in bucket_names:
+    for bucket in valid_bucket_names:
         objs = minioClient.list_objects(bucket)
         for obj in objs:
-            all_files.append(obj.__dict__['object_name'])
+            all_files.append(obj.__dict__["object_name"])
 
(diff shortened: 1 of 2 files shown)


flask/app/models.py
```diff
@@ -9,26 +9,6 @@
...
-        self.password_hash = generate_password_hash(
-            password, method="pbkdf2:sha256:50000"
-        )
-
-    def check_password(self, password):
-        return check_password_hash(self.password_hash, password)
-
-    def get_id(self):
-        return self.user_id
-
```
kevinlul, line 31: Could you move the `user_groups` table up instead of moving the `User` model down?

flask/app/buckets.py
```diff
@@ -33,3 +39,46 @@ def get_object_info():
...
+        code = group.group_code.lower()
+        if code in all_bucket_names:
+            valid_bucket_names.append(code)
+
+    # Get all files in valid minio buckets
+    all_files = []
+    for bucket in valid_bucket_names:
+        objs = minioClient.list_objects(bucket)
+        for obj in objs:
+            all_files.append(obj.__dict__["object_name"])
```
kevinlul, line 70: Does `obj.object_name` not work?

kevinlul (changes requested):
How did you go about testing this endpoint?

flask/app/buckets.py
SamarthVP: Ah yes it does!

kevinlul (changes requested):
The returned file names should be prefixed with `bucketname/` since object names are only unique within a bucket.

The runtime of this is `O(datasets * objects)` but if we use a dictionary (implemented as a hash table) to store either the known linked files or the unlinked files, then the runtime should be just `O(datasets + objects)` amortized due to constant search instead of linear search time. The same applies for finding valid bucket names but this isn't a big concern there since we don't expect to have many of those.

SamarthVP:
Great idea, with those changes the run time is now effectively O(datasets + total objects) assuming the number of groups and buckets is small.

flask/app/buckets.py
```diff
@@ -33,3 +39,46 @@ def get_object_info():
...
+    # Get all linked files
+    linked_files = {}
+    for dataset in models.Dataset.query.all():
+        if dataset.input_hpf_path is not None:
+            linked_files[dataset.input_hpf_path] = ":)"
+
+    # Put all unlinked files in new list
+    unlinked_files = []
+    for file_name in all_files:
+        if file_name[file_name.index('/')+1:] not in linked_files:
```
kevinlul, line 81: This should just be `file_name` because we should be storing the combined bucket and object name that is unique

kevinlul (changes requested):

flask/app/buckets.py
SamarthVP: oh so `input_hpf_path` will have the bucket prefix as well, got it

kevinlul (approved):
LGTM!

PR #176, opened 2020-11-11
CI: add pull request labeller workflow

kevinlul:
#122

### DIFF
--- a/.github/labels.yml
+++ b/.github/labels.yml
@@ -1,0 +1,25 @@
+dependencies:
+- flask/requirements*.txt
+- flask/requirements*.in
+- react/yarn.lock
+database:
+- flask/app/manage.py
+- flask/app/models.py
+- flask/migrations/**/*
+backend:
+- any: ["flask/**/*"]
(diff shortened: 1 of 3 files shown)


.github/labels.yml
```diff
@@ -0,0 +1,25 @@
...
+- flask/requirements*.txt
+- flask/requirements*.in
+- react/yarn.lock
+database:
+- flask/app/manage.py
+- flask/app/models.py
+- flask/migrations/**/*
+backend:
+- any: ["flask/**/*"]
+  all: ["!flask/app/manage.py", "!flask/app/models.py", "!flask/migrations/**/*"]
```
SamarthVP, line 11: If flask/app/manage.py and flask/something are changed, I don't think this will apply the backend label. Something like 
```
# Add 'source' label to any change to src files within the source dir EXCEPT for the docs sub-folder
source:
- any: ['src/**/*', '!src/docs/*']
```
might be more applicable

kevinlul: Good catch! Should be fixed

SamarthVP (approved):

PR #178, opened 2020-11-12
Flask: POST /api/tissue_samples

### LINKED ISSUE #168: POST /api/tissue_samples
Request body is a JSON object containing required fields to add a tissue sample.
Create a new tissue sample in the database. If the participant does not exist, fail with 404.
Return a JSON object of the created sample (make sure the primary key is included) with status 201. Set the Location header to /api/tissue_samples/:id, the endpoint you would query to find the created tissue sample.

Task points: 1

Madelinehazel:
closes #168

testing: 
```
from datetime import datetime
import requests
import json

tissue = {
            "participant_id": 1,
            "extraction_date": "0001-01-01T00:00:00-04:00",
            "tissue_processing": "FreshFrozen",
            "notes": "blahblah"
        }

url = 'http://0.0.0.0:5000/api/tissue_samples'
result = requests.post(url, json = tissue)
r = result.json()
print(r)
```

### DIFF
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -524,10 +524,6 @@ def post_tissue():
     if enum_error:
         return enum_error, 400
 
-    now = datetime.now()
-    created = now
-    updated = now
-
     try:
         created_by = updated_by = current_user.user_id
     except:  # LOGIN DISABLED
@@ -535,27 +531,21 @@ def post_tissue():
 
     tissue_sample = models.TissueSample(
         **{
-            "participant_id": request.json["participant_id"],
-            "extraction_date": request.json["extraction_date"],
+            "participant_id": request.json.get("participant_id"),
+            "extraction_date": request.json.get("extraction_date"),
             "tissue_sample_type": request.json["tissue_sample_type"],
-            "tissue_processing": request.json["tissue_processing"],
-            "notes": request.json["notes"],
-            "created": created,
+            "tissue_processing": request.json.get("tissue_processing"),
+            "notes": request.json.get('notes'),
             "created_by": created_by,
-            "updated": updated,
             "updated_by": updated_by
         }
     )
     try:
         db.session.add(tissue_sample)
-        db.session.flush()
+        db.session.commit()
         ts_id = tissue_sample.tissue_sample_id
-        tissue_sample_new = {
-             **asdict(tissue_sample),
-             "tissue_sample_id": ts_id
-         }
         location_header = "/api/tissue_samples/{}".format(ts_id)
-        return jsonify(tissue_sample_new), 201, {"location": location_header}
+        return jsonify(tissue_sample), 201, {"location": location_header}
     except:
         db.session.rollback()
         return "Server error", 500
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -507,6 +507,7 @@ def delete_tissue(id: int):
     else:
         return "Not Found", 404
 
+
 @app.route("/api/tissue_samples", methods=["POST"])
 @login_required
 def post_tissue():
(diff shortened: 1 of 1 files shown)


flask/app/routes.py
```diff
@@ -506,3 +506,56 @@ def delete_tissue(id: int):
...
+    except:  # LOGIN DISABLED
+        created_by = updated_by = 1
+
+    tissue_sample = models.TissueSample(
+        **{
+            "participant_id": request.json["participant_id"],
+            "extraction_date": request.json["extraction_date"],
+            "tissue_sample_type": request.json["tissue_sample_type"],
+            "tissue_processing": request.json["tissue_processing"],
+            "notes": request.json["notes"],
```
delvinso, line 542: you can use the get method (`request.json.get('notes')`) so `None` is returned if the value for a field isn't supplied in the json. right now if there is no value say for notes then `JSONDecodeError` is raised

flask/app/routes.py
```diff
@@ -506,3 +506,56 @@ def delete_tissue(id: int):
...
+        request.json,
+        ["tissue_sample_type","tissue_processing"]
+        )
+
+    if enum_error:
+        return enum_error, 400
+
+    now = datetime.now()
+    created = now
+    updated = now
```
kevinlul, line 529: These are automatically set by SQLAlchemy because of how we wrote `models.py` so you don't need to set them.

flask/app/routes.py
```diff
@@ -506,3 +506,56 @@ def delete_tissue(id: int):
...
+            "notes": request.json["notes"],
+            "created": created,
+            "created_by": created_by,
+            "updated": updated,
+            "updated_by": updated_by
+        }
+    )
+    try:
+        db.session.add(tissue_sample)
+        db.session.flush()
```
kevinlul, line 551:
```suggestion
        db.session.commit()
```
or the transaction doesn't actually happen and you can check by fetching `/api/tissue_samples/:id` (the Location you return)

flask/app/routes.py
kevinlul, line 542: Aside from tissue_sample_type, none of these are mandatory fields.

flask/app/routes.py
```diff
@@ -506,3 +506,56 @@ def delete_tissue(id: int):
...
+        ts_id = tissue_sample.tissue_sample_id
+        tissue_sample_new = {
+             **asdict(tissue_sample),
+             "tissue_sample_id": ts_id
```
kevinlul, line 555: tissue_sample_id is already serialized so you can just return `jsonify(tissue_sample)`

kevinlul (changes requested):

flask/app/routes.py
Madelinehazel: neat, thanks!

Madelinehazel: I changed these to request.json.get(<fieldname>) as @delvinso  suggested for notes- is this correct?

flask/app/routes.py
```diff
@@ -506,3 +506,46 @@ def delete_tissue(id: int):
...
     else:
         return "Not Found", 404
+
+@app.route("/api/tissue_samples", methods=["POST"])
+@login_required
+def post_tissue():
+    if not request.json:
+        return "Request body must be JSON", 400
+
+    models.Participant.query.filter_by(participant_id=request.json["participant_id"]).first_or_404()
```
kevinlul, line 516: KeyError here if participant_id is missing

flask/app/routes.py
```diff
@@ -506,3 +506,46 @@ def delete_tissue(id: int):
...
+    try:
+        created_by = updated_by = current_user.user_id
+    except:  # LOGIN DISABLED
+        created_by = updated_by = 1
+
+    tissue_sample = models.TissueSample(
+        **{
+            "participant_id": request.json.get("participant_id"),
+            "extraction_date": request.json.get("extraction_date"),
+            "tissue_sample_type": request.json["tissue_sample_type"],
```
kevinlul, line 536: KeyError here if tissue_sample_type is missing

kevinlul (changes requested):

kevinlul (approved):
LGTM

PR #181, opened 2020-11-12
React: Add link to tables from info dialogs

### LINKED ISSUE #163: Add a link from dialog table rows to "go to dataset"
Requires some research

Task points: 8

frewmack:
Closes #163


kevinlul (commented):
What's the expected behaviour here?

frewmack:
> What's the expected behaviour here?

When the user opens a detail dialog about any given element in a table (datasets, analyses, participants), and they open one of the listed items from a different table (eg. datasets from a participant, or analyses from a dataset), clicking the 'Open in table' button will take them to that other table, and change the ID filter so that the item they clicked is the only one present.

This functionality is mostly handled by react-router. Any links to `/<tablename>/:id` will direct them to the specified table, with the specified ID in the table's ID filter.

react/src/pages/participants/ParticipantInfoDialog.tsx
```diff
@@ -49,38 +49,44 @@ const getParticipantValues = (participant: Participant) => {
...
+    let analysisList: Analysis[] = [];
+    for (const dataset of datasets) {
+        const response = await fetch("/api/datasets/" + dataset.dataset_id);
+        if (response.ok) {
+            const data = await response.json();
+            console.log(data);
+            analysisList = analysisList.concat(data.analyses as Analysis[]);
+        } else {
+            console.error(response);
+            throw new Error(`GET /api/datasets/${dataset.dataset_id} failed. Reason: ${response.status} - ${response.statusText}`);
```
kevinlul, line 74: To be consistent with the rest, this should also be in the useEffect and enqueue a snackbar message if something goes wrong with the request.

react/src/pages/utils/components/InfoList.tsx
```diff
@@ -65,6 +71,16 @@ export default function InfoList(props: { infoList: Info[]; title?: string; icon
                                     collapsibleTitles={info.collapsibleTitles}
                                     collapsibleValues={info.collapsibleValues}
                                 />
+                                {props.linkPath && info.identifier &&
+                                    <Button
+                                    href={`${props.linkPath}/${info.identifier}`}
```
kevinlul, line 76: You should use the react-router's history API to push a state here, as a regular hyperlink like this causes a full page reload.

kevinlul (changes requested):

kevinlul (approved):
LGTM

react/src/pages/utils/components/InfoList.tsx
```diff
@@ -65,6 +73,16 @@ export default function InfoList(props: { infoList: Info[]; title?: string; icon
                                     collapsibleTitles={info.collapsibleTitles}
                                     collapsibleValues={info.collapsibleValues}
                                 />
+                                {props.linkPath && info.identifier &&
```
annnag, line 76: Could add some space between these two buttons
Screen Shot 2020-11-13 at 10 21 57 AM

annnag (changes requested):
Looks good! One minor problem: when I get redirected to a table from a dialog and I clear the ID filter, the ID parameter is not removed from the URL. So if I refresh the page, the table gets filtered again. I think it would be better if the URL changes after the user clear the filter.

Issue #117, opened 2020-10-27
Epic: permission gating


kevinlul:
- [x] Admins can always perform every action
- [x] Regular users can only retrieve and mutate datasets associated with permission groups they are in (@kevinlul, 5 task points)
- [x] Permissions for families, tissue samples, participants, and analyses are inferred, so users can only see those containing datasets they can see (@SamarthVP, 5 task points, probably depends on the above item)
- [ ] Users can only request analyses on datasets they have access (@Madelinehazel, 1 task point assuming the above are done)
- [ ] Users can only cancel analyses; the other state changes are admin-only (@Madelinehazel, 1 task point for backend) (@frewmack, 1 task point for frontend)

PR #184, opened 2020-11-13
Flask: Permission Gating

SamarthVP:
For #117

The endpoints are a pain to test, but one can set `LOGIN_DISABLED=True` in `config.py` and then pass `?user=x` to the endpoints to emulate users

Then mess with the db and see what comes out.

Testing PATCH `/api/datasets/:id`
```
curl -X PATCH \
    --data '{
            "input_hpf_path": "woohoo"
    }' \
    localhost:5000/api/datasets/3\?user=1
```

Testing PATCH `/api/participants/:id`
```
curl -X PATCH \
    --data '{
            "notes": "k"
    }' \
    localhost:5000/api/participants/3\?user=1
```

Testing PATCH `/api/families/:id`
```
curl -X PATCH \
    --data '{
            "family_codename": 5
    }' \
    localhost:5000/api/families/1\?user=1
```

Testing PATCH `/api/analyses/:id`
```
curl -X PATCH \
    --data '{
            "result_hpf_path": "nice one 2"
    }' \
    localhost:5000/api/analyses/2\?user=1
```


flask/app/participants.py
```diff
@@ -15,19 +15,51 @@
...
 
     participants = [
         {
             **asdict(participant),
             "family_codename": participant.family.family_codename,
             "tissue_samples": [
-                {**asdict(tissue_sample), "datasets": tissue_sample.datasets}
+                {
+                    **asdict(tissue_sample),
+                    "datasets": (
```
SamarthVP, line 45: Multiple db queries works but runs the risk of being slow. I couldn't really think of a better way though...
Thoughts @kevinlul ?

flask/app/families.py
```diff
@@ -11,14 +11,46 @@
 @login_required
 def families_list():
 
-    db_families = models.Family.query.options(
-        joinedload(models.Family.participants)
-    ).all()
+    db_families = (
+        models.Family.query.join(models.Participant)
```
kevinlul, line 15: Before the first join, add `.options(contains_eager(models.Family.participants))` and make sure to import `contains_eager` from `sqlalchemy.orm` and see if this works for the rest.

flask/app/datasets.py
```diff
@@ -17,6 +17,13 @@
 def datasets_list():
     db_datasets = (
         db.session.query(models.Dataset)
+        .join(models.groups_datasets_table)
+        .join(
+            models.users_groups_table,
+            models.groups_datasets_table.columns.group_id
+            == models.users_groups_table.columns.group_id,
+        )
+        .filter(models.users_groups_table.columns.user_id == current_user.user_id)
```
kevinlul, line 26: We need to remember to ignore this if the current user is an admin

SamarthVP:
Now that we've done the general `/api/*thing*` endpoints, I believe we need to extend this to the `/api/*thing*/:id` endpoints. Or is that not a concern because `/api/*thing*/:id` would only ever be called on `*things*` coming from the `/api/*thing*` endpoints?

kevinlul:
It should be extended and the logic would be to 404 if the user doesn't have permission (maybe 403 if assuming a user identity as admin).

SamarthVP:
I think 404 in all cases where permission is blocked makes the most sense

flask/app/participants.py
```diff
@@ -76,15 +133,13 @@ def update_participant(id: int):
...
         return enum_error, 400
 
-    try:
-        table.updated_by = current_user.user_id
-    except:
-        pass  # LOGIN_DISABLED
+    if user_id:
+        participant.updated_by = user_id
 
     routes.transaction_or_abort(db.session.commit)
```
SamarthVP, line 145: Should return jsonify(participant) since table got renamed

flask/app/analyses.py
```diff
@@ -24,65 +29,116 @@ def analyses_list():
...
-    return jsonify(analyses)
+    return jsonify(
+        [
+            {
+                **asdict(analysis),
+                "requester": requester and requester.username,
+                "updated_by": updated_by and updated_by.username,
+                "assignee": assignee and assignee.username,
+            }
+            for analysis, requester, updated_by, assignee in db_analyses
```
SamarthVP, line 72:
```suggestion
            for analysis, requester, updated_by, assignee in analyses
```
Another one

helrick:
lgtm! (based on currently defined permissions)

small item: non-admins are still able to see the admin page. not sure if that's in the scope of this issue though

helrick (approved):

PR #187, opened 2020-11-16
Flask: POST /api/datasets

### LINKED ISSUE #169: POST /api/datasets
Request body is a JSON object containing required fields to add a dataset.
Create a new dataset in the database. If the owning tissue sample does not exist, fail with 404.
Return a JSON object of the created dataset (make sure the primary key is included) with status 201. Set the Location header to /api/datasets/:id, the endpoint you would query to find the created tissue sample.

Task points: 1

Madelinehazel:
Closes #169

```
curl -X POST \
    --data '{
            "tissue_sample_id": "1",
            "dataset_type": "CES",
            "input_hpf_path": "/hpf/CES",
            "notes": "blahblah",
            "condition": "GermLine",
            "extraction_protocol": "Something",
            "capture_kit": "capture kit x",
            "read_length": "150",
            "read_type": "PairedEnd",
            "discriminator": "dataset"
        }' \
    localhost:3000/api/datasets
```

returns
```
{
  "batch_id": null, 
  "capture_kit": "capture kit x", 
  "condition": "GermLine", 
  "created": "Mon, 16 Nov 2020 21:37:06 GMT", 
  "created_by": 1, 
  "dataset_id": 5, 
  "dataset_type": "CES", 
  "extraction_protocol": "Something", 
  "input_hpf_path": "/hpf/CES", 
  "library_prep_date": null, 
  "library_prep_method": null, 
  "notes": "blahblah", 
  "read_length": 150, 
  "read_type": "PairedEnd", 
  "sequencing_centre": null, 
  "sequencing_date": null, 
  "sequencing_id": null, 
  "tissue_sample_id": 1, 
  "updated": "Mon, 16 Nov 2020 21:37:06 GMT", 
  "updated_by": 1
}
```

### DIFF
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -11,6 +11,24 @@ from sqlalchemy import exc
 from sqlalchemy.orm import aliased, joinedload
 from werkzeug.exceptions import HTTPException
 
+editable_columns = [
+    "dataset_type",
+    "input_hpf_path",
+    "notes",
+    "condition",
+    "extraction_protocol",
+    "capture_kit",
+    "library_prep_method",
+    "library_prep_date",
+    "read_length",
+    "read_type",
+    "sequencing_id",
+    "sequencing_date",
+    "sequencing_centre",
+    "batch_id",
+    "discriminator",
+]
+
 
 @app.route("/api/datasets", methods=["GET"])
 @login_required
@@ -107,24 +125,6 @@ def update_dataset(id: int):
 
     table = models.Dataset.query.get_or_404(id)
 
-    editable_columns = [
-        "dataset_type",
-        "input_hpf_path",
-        "notes",
-        "condition",
-        "extraction_protocol",
-        "capture_kit",
-        "library_prep_method",
-        "library_prep_date",
-        "read_length",
-        "read_type",
-        "sequencing_id",
-        "sequencing_date",
-        "sequencing_centre",
-        "batch_id",
-        "discriminator",
-    ]
-
     enum_error = routes.mixin(table, request.json, editable_columns)
 
     if enum_error:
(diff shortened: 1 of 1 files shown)


flask/app/datasets.py
```diff
@@ -138,3 +138,65 @@ def update_dataset(id: int):
...
+            "sequencing_id": request.json.get("sequencing_id"),
+            "sequencing_date": request.json.get("sequencing_date"),
+            "sequencing_centre": request.json.get("sequencing_centre"),
+            "batch_id": request.json.get("batch_id"),
+            "created_by": created_by,
+            "updated_by": updated_by,
+            "discriminator": request.json.get("discriminator"),
+        }
+    )
+    try:
```
delvinso, line 194: minor - you can wrap the try/except statement into a single line using `routes.transaction_or_abort(db.session.commit)`

flask/app/datasets.py
```diff
@@ -138,3 +138,65 @@ def update_dataset(id: int):
...
+
+    tissue_sample_id = request.json.get("tissue_sample_id")
+    if not tissue_sample_id:
+        return "A tissue sample id must be provided", 400
+
+    models.TissueSample.query.filter_by(
+        tissue_sample_id=tissue_sample_id
+    ).first_or_404()
+
+    col_names = [key for key in request.json]
```
delvinso, line 161: This will check all keys in the json regardless of whether they are valid and can be added under a dataset. I think these should be checking a pre-defined list of keys similar to the other POST/PATCH endpoints - what do you think @kevinlul

flask/app/datasets.py
Madelinehazel: nice, that's clean!

flask/app/datasets.py
Madelinehazel: good point! i think that makes sense- hadn't considered that it would be checking invalid keys

delvinso (approved):
Looks good!

PR #190, opened 2020-11-19
React: Edit functionality in dialogs

### LINKED ISSUE #118: Epic: edit functionality in dialogs
Not all database columns can fit in the table, so some edit functionality has to exist in the popups too. Draft a design for each first and make sure they are consistent.

- [x] Participant
- [x] Dataset

Task points: 5

annnag:
Closes #118


react/src/pages/datasets/DatasetInfoDialog.tsx
```diff
@@ -66,7 +51,8 @@ export default function DatasetInfoDialog({ dataset_id, open, onClose }: DialogP
     const [dataset, setDataset] = useState<Dataset>();
     const [analyses, setAnalyses] = useState<Analysis[]>([]);
     const [sample, setSample] = useState<Sample>();
-
+    //for updating the dialog content when re-open the dialog
+    const [num, reRender] = useState(0);
```
frewmack, line 55: You don't need to force a re-render. Change this dialog to accept the dataset object as a prop, and add it to the `useEffect` dependencies below, same as in ParticipantInfoDialog. `useEffect` is not fetching the dataset again because `dataset_id` does not change.

react/src/pages/datasets/DatasetTable.tsx
```diff
@@ -87,7 +89,10 @@ export default function DatasetTable() {
                 <DatasetInfoDialog
                     dataset_id={infoDataset.dataset_id}
                     open={showInfo}
-                    onClose={() => setShowInfo(false)}
+                    onClose={() => {
+                        setShowInfo(false);
+                        reRender(n => n + 1);
```
frewmack, line 94: The table should only be re-rendered if the user changed something in the dialog. You should also look into re-using material-table onRowUpdate for updating/redrawing the table, as material-table probably has a more optimized means of redrawing new rows.

react/src/pages/participants/ParticipantTable.tsx
```diff
@@ -69,15 +71,18 @@ export default function ParticipantTable() {
...
     return (
         <div>
             {activeRow && (
                 <ParticipantInfoDialog
                     open={detail}
                     participant={activeRow}
-                    onClose={() => setDetail(false)}
+                    onClose={() => {
+                        setDetail(false);
+                        reRender(n => n + 1);
```
frewmack, line 84: Same as the dataset table, a re-render should only occur if the user changed something.

react/src/pages/utils/components/DetailSection.tsx
```diff
@@ -1,119 +1,441 @@
...
+            };
+            const newData = [...fieldsOnEdit];
+            newData[idx] = newField;
+            setFields(newData);
+        }
+    }
+
+    async function updateData() {
+        let url;
+        let identifier;
```
frewmack, line 331: From my testing, this identifier always ends up as `undefined`.

frewmack (changes requested):

react/src/pages/datasets/DatasetTable.tsx
```diff
@@ -85,9 +85,23 @@ export default function DatasetTable() {
...
+                        const updatedDatasets = [...datasets];
+                        const idx = datasets.findIndex(
+                            dataset => dataset.dataset_id === dataset_id
+                        );
+                        const updatedDataset = { ...datasets[idx] } as { [key: string]: any };
+                        Object.keys(newDataset).forEach(key => {
+                            updatedDataset[key] = newDataset[key];
+                        });
+                        updatedDatasets[idx] = updatedDataset as Dataset;
+                        setDatasets(updatedDatasets);
```
kevinlul, line 103: This can be done with a single `map` call over `datasets` and passing the result to `setDatasets`. The only element that should be replaced by map is the one with matching `dataset_id`, where we make a copy of the old and mixin properties of the new.

react/src/pages/participants/ParticipantTable.tsx
```diff
@@ -77,7 +77,24 @@ export default function ParticipantTable() {
...
+                        );
+                        const updatedParticipant = { ...participants[idx] } as {
+                            [key: string]: any;
+                        };
+                        Object.keys(newParticipant).forEach(key => {
+                            updatedParticipant[key] = newParticipant[key];
+                        });
+                        updatedParticipants[idx] = updatedParticipant as Participant;
+
+                        setParticipants(updatedParticipants);
```
kevinlul, line 93: ditto

react/src/pages/utils/components/DetailSection.tsx
```diff
@@ -1,119 +1,451 @@
...
+                                    </Box>
+                                </Fade>
+                            ) : null}
+                            {!editMode ? (
+                                <Fade in={!editMode} mountOnEnter unmountOnExit>
+                                    <Box className={classes.box}>
+                                        <FieldDisplay title={field.title} value={field.value} />
+                                    </Box>
+                                </Fade>
+                            ) : null}
```
kevinlul, line 274: Shouldn't this be one ternary instead of two? Or just if/else, possibly even as part of the existing branching in this function?

kevinlul (changes requested):
In the participant dialog, do not allow editing the family codename. The backend won't actually change it anyway.

In the dataset dialog, do not allow editing the family or participant codenames as this will affect other datasets or other parts of the hierarchy. This is consistent with the behaviour in table rows.

kevinlul (approved):

frewmack (approved):

PR #197, opened 2020-11-23
adding constraints to groupname to match minio rules

### LINKED ISSUE #174: Constrain Group.group_code to only valid domain names
Allowed characters: lowercase letters, digits, and hyphens. Length: minimum 3, maximum 63.

The group_code will be used to name the corresponding MinIO bucket. 

Task points: 1

helrick:
closes #174


flask/app/models.py
```diff
@@ -44,8 +44,11 @@ def get_id(self):
 class Group(db.Model):
     group_id: int = db.Column(db.Integer, primary_key=True)
     group_code: str = db.Column(
-        db.String(50), nullable=False, unique=True
-    )  # i.e. CHEO, SK, AB
+        db.String(50),
```
kevinlul, line 47: The maximum length can be checked by the width of this field instead of the constraint since `LENGTH(group_code) < 64` will always be true

kevinlul (changes requested):

flask/app/models.py
helrick: resolved in 42b09e1

kevinlul (approved):

PR #199, opened 2020-11-24
Flask: add unit tests for all endpoints

### LINKED ISSUE #165: Epic: write tests for backend
Modules
- [x] families.py
- [x] participants.py
- [x] datasets.py
- [x] analyses.py
- [x] miscellaneous ones in routes.py, excluding authentication

Task points: 21

kevinlul:
Implement check_admin decorator
Add Config.SQLALCHEMY_LOG to disable SQL query logs in unit tests

Closes #165

### DIFF
--- a/flask/app/__init__.py
+++ b/flask/app/__init__.py
@@ -4,10 +4,6 @@ from flask_login import LoginManager
 from flask_migrate import Migrate
 from flask_sqlalchemy import SQLAlchemy
 
-logging.basicConfig()
-logging.getLogger("sqlalchemy.engine").setLevel(logging.INFO)
-logging.getLogger("sqlalchemy.pool").setLevel(logging.INFO)
-
 db = SQLAlchemy()
 login = LoginManager()
 migrate = Migrate()
@@ -22,6 +18,11 @@ def create_app(config):
     app = Flask(__name__)
 
     app.config.from_object(config)
+
+    if config.SQLALCHEMY_LOG:
+        logging.basicConfig()
+        logging.getLogger("sqlalchemy.engine").setLevel(logging.INFO)
+        logging.getLogger("sqlalchemy.pool").setLevel(logging.INFO)
 
     flask_logging.create_logger(app)
     login.session_protection = "strong"
--- a/flask/app/analyses.py
+++ b/flask/app/analyses.py
@@ -142,9 +142,8 @@ def get_analysis(id: int):
 @app.route("/api/analyses", methods=["POST"])
 @login_required
 def create_analysis():
-
     if not request.json:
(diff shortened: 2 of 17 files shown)


helrick (approved):
Only question is about the warnings summary. Unclear why/how warning messages are selected.

```
tests/test_families.py: 2 warnings
tests/test_misc.py: 85 warnings
  /usr/local/lib/python3.7/site-packages/sqlalchemy/orm/query.py:223: SADeprecationWarning: Plain string expression passed to Query() should be explicitly declared using literal_column(); automatic coercion of this value will be removed in SQLAlchemy 1.4
    entity_wrapper(self, ent)
```

Here for example, why are there no warnings displayed for `test_families.py` despite it having 2? Additionally, are the 85 warnings for `test_misc.py` duplicates? Or is only the first shown?

Everything else looks good though so ✅

delvinso:
Are these warnings suppose to occur?

```
tests/test_analyses.py::test_no_analyses
  /usr/local/lib/python3.7/site-packages/minio/copy_conditions.py:29: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    from collections import MutableMapping

tests/test_datasets.py::test_create_dataset
  /usr/local/lib/python3.7/site-packages/sqlalchemy/orm/mapper.py:1691: SAWarning: Flushing object <Dataset at 0x7f8ee5875810> with incompatible polymorphic identity None; the object may not refresh and/or load correctly (this warning may be suppressed after 10 occurrences)
    (state_str(state), dict_[polymorphic_key]),

tests/test_datasets.py::test_create_dataset
  /usr/local/lib/python3.7/site-packages/sqlalchemy/orm/mapper.py:1691: SAWarning: Flushing object <Dataset at 0x7f8ee58f1810> with incompatible polymorphic identity None; the object may not refresh and/or load correctly (this warning may be suppressed after 10 occurrences)
    (state_str(state), dict_[polymorphic_key]),
```

SamarthVP:
The first seems to be a warning from minio, not much we can/should do there. The second and third appear to be because the dataset that is created has some fields set to `None` instead of leaving them out entirely. However, seeing as the endpoint updates the db properly, the only risk is if the dataset was queried from the session before the endpoint committed. But there is no reason we would want to do that, so even though this could be fixed in the endpoint we really have no reason to.

delvinso (approved):
Test logic looks sound to me (have to say I enjoyed some of the test json fields/values) - no issues asides from what I brought up earlier which Samarth cleared up.  Only (minor) comment is that the formatting for the endpoints could be more consistent.

Issue #128, opened 2020-10-30
Add query parameters to limit the number of objects returned and the columns returned per object


helrick:
This will reduce overfetch and aid in dynamic autocomplete on the add participant page

- [x] /api/families (3 task points)
- [x] /api/participants (3 task points)

PR #195, opened 2020-11-23
Flask: limit number of families returned

Madelinehazel:
Related to #128 

`curl "http://localhost:3000/api/families?order=family_codename&starts_with=1&max_rows=1"`
returns:

```
[
  {
    "created": "Wed, 18 Nov 2020 21:12:38 GMT", 
    "created_by": 1, 
    "family_codename": "1000", 
    "family_id": 1, 
    "participants": [
      {
        "affected": true, 
        "created": "Wed, 18 Nov 2020 21:12:38 GMT", 
        "created_by": 1, 
        "month_of_birth": "Sat, 01 Jan 2000 00:00:00 GMT", 
        "notes": "Extra info about sample here", 
        "participant_codename": "AA001", 
        "participant_id": 1, 
        "participant_type": "Proband", 
        "sex": "Female", 
        "solved": null, 
        "updated": "Wed, 18 Nov 2020 21:12:38 GMT", 
        "updated_by": 1
      }, 
      {
        "affected": false, 
        "created": "Wed, 18 Nov 2020 21:12:38 GMT", 
        "created_by": 1, 
        "month_of_birth": "Thu, 01 Jan 1970 00:00:00 GMT", 
        "notes": "", 
        "participant_codename": "AA002", 
        "participant_id": 2, 
        "participant_type": "Parent", 
        "sex": "Male", 
        "solved": null, 
        "updated": "Wed, 18 Nov 2020 21:12:38 GMT", 
        "updated_by": 1
      }, 
      {
        "affected": false, 
        "created": "Wed, 18 Nov 2020 21:12:38 GMT", 
        "created_by": 1, 
        "month_of_birth": "Sun, 01 Feb 1970 00:00:00 GMT", 
        "notes": "", 
        "participant_codename": "AA003", 
        "participant_id": 3, 
        "participant_type": "Parent", 
        "sex": "Female", 
        "solved": null, 
        "updated": "Wed, 18 Nov 2020 21:12:38 GMT", 
        "updated_by": 1
      }, 
      {
        "affected": false, 
        "created": "Wed, 18 Nov 2020 21:12:38 GMT", 
        "created_by": 1, 
        "month_of_birth": "Sun, 01 Mar 1970 00:00:00 GMT", 
        "notes": "", 
        "participant_codename": "AA004", 
        "participant_id": 4, 
        "participant_type": "Parent", 
        "sex": "Female", 
        "solved": null, 
        "updated": "Wed, 18 Nov 2020 21:12:38 GMT", 
[...]

PR #200, opened 2020-11-25
Flask: add query parameters to GET api/participants

delvinso:
For `order_by()`, we need to add a default field, the alternative is to have an if else statement checking for the existence of the order query parameter and having one query with the order_by method, and one without. AFAIK there is no 'default' parameter we can pass into order_by to get the default sql ordering scheme (if there is we can just assign that as default and pass it into order_by) 

No user groups exist for testing permissions so we need to exec into the MySQL container and `insert into users_groups values (1,1); insert into groups_datasets values (1,1);` (thanks to Samarth)

some examples:

```
curl  "localhost:5000/api/participants?max_rows=1&starts_with=AA&order=sex" 
curl "localhost:5000/api/participants?max_rows=1&starts_with=AA&order=sex&user=1
curl "localhost:5000/api/participants?max_rows=1&starts_with=AA&order=sFx&user=1"
```

For #128

### DIFF
--- a/flask/app/participants.py
+++ b/flask/app/participants.py
@@ -28,11 +28,16 @@ editable_columns = [
 def list_participants():
 
     # parsing query parameters
-    max_rows = request.args.get("max_rows", default=100, type=int)
+    max_rows = request.args.get("max_rows", default=100)
     starts_with = request.args.get("starts_with", default="", type=str)
     starts_with = f"{starts_with}%"  # sql syntax
     order_by_col = request.args.get("order", default="participant_id", type=str)
     # need some default or we need an ifelse statement, one with order_by method and one without. AFAIK there is no 'default' parameter we can pass into order_by to get the default sql ordering scheme
+
+    try:
+        int(max_rows)
+    except:
+        return "Max rows must be a valid integer", 400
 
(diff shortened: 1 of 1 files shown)


SamarthVP (approved):
Looks great! Permissions aren't broken and still passes all tests in #199

flask/app/participants.py
```diff
@@ -26,6 +26,20 @@
 @app.route("/api/participants", methods=["GET"])
 @login_required
 def list_participants():
+
+    # parsing query parameters
+    max_rows = request.args.get("max_rows", default=100, type=int)
```
Madelinehazel, line 31: It looks like the behaviour of using type=int here means that if max_rows is passed a string, it instead uses the default value for max_rows rather than passing an error. Not sure if that's a problem or not?

delvinso: ooh good point like you said it still works even if the query is something like max_rows='DFSDFDSF', should I hold off on this until we figure out the input validation mentioned in the retrospective?

kevinlul: As long as this and #195 roughly match up it should be okay

kevinlul (approved):

PR #213, opened 2020-11-26
React: Group management component for admin page

frewmack:


react/src/pages/utils/components/ChipSelect.tsx
```diff
@@ -0,0 +1,107 @@
...
+                    disabled={disableAdd}
+                >
+                    <AddCircle />
+                </IconButton>
+            </Paper>
+            <Menu anchorEl={anchorEl} open={!!anchorEl} onClose={() => setAnchorEl(null)}>
+                {chips
+                    .filter(chip => !chip.selected)
+                    .map(chip => (
+                        <MenuItem
```
kevinlul, line 95: missing key={chip.key}

react/src/pages/utils/components/ChipSelect.tsx
```diff
@@ -0,0 +1,107 @@
...
+        if (Array.isArray(props.selected)) {
+            setChips(chips =>
+                chips.map(chip => ({
+                    ...chip,
+                    selected: !!props.selected!.find(label => label === chip.label),
+                }))
+            );
+            setDisableAdd(props.selected.length === 0);
+        }
+    }, [props.selected]);
```
kevinlul, line 54: This is all determined from the props, so is there a reason to run this in `useEffect` instead of just initializing the states from the props at the top?

kevinlul (changes requested):

react/src/pages/utils/components/ChipSelect.tsx
frewmack: `useEffect` handles the case where the user clicks Cancel and reverts their changes

kevinlul: Ah, I see

kevinlul (approved):
LGTM

PR #236, opened 2020-11-27
React: Hook up add participants submit button

### LINKED ISSUE #222: Add participants: hook up submit button

frewmack:
Closes #222.
The `condition` field was also changed to be required on frontend with default value `GermLine`, since it's set as not nullable.

### DIFF
--- a/react/src/pages/addParticipants/AddParticipants.tsx
+++ b/react/src/pages/addParticipants/AddParticipants.tsx
@@ -1,7 +1,8 @@
-import React, { useEffect } from "react";
+import React, { useEffect, useState } from "react";
 import { makeStyles, Container, Button } from "@material-ui/core";
 import { CloudUpload } from "@material-ui/icons";
 import DataEntryTable from "./DataEntryTable";
+import { DataEntryRow } from "../utils/typings";
 
 const useStyles = makeStyles(theme => ({
     appBarSpacer: theme.mixins.toolbar,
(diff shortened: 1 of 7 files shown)


react/src/pages/addParticipants/AddParticipants.tsx
```diff
@@ -27,16 +32,43 @@ const useStyles = makeStyles(theme => ({
...
+            method: "POST",
+            headers: {
+                "Content-Type": "application/json",
+            },
+            body: JSON.stringify(data),
+        });
+
+        if (response.ok) {
+            const responseData: Array<any> = await response.json();
+            enqueueSnackbar(`${responseData.length} participants successfully added.`, {
```
kevinlul, line 55: Should be datasets that are added

react/src/pages/utils/functions.tsx
```diff
@@ -251,3 +252,20 @@ export function createFieldObj(
...
+
+    var arr = [];
+    for (let i = 0; i < amount; i++) {
+        arr.push({
+            family_codename: "",
+            participant_codename: "",
+            participant_type: "",
+            tissue_sample_type: "",
+            dataset_type: "",
+            condition: "GermLine",
```
kevinlul, line 267: I'll wait until standup but I'm not sure if this is the best default choice, if there even should be one.

kevinlul (changes requested):
Can we verify that the required fields are filled out before trying to submit?

react/src/pages/addParticipants/AddParticipants.tsx
```diff
@@ -64,24 +64,58 @@ export default function AddParticipants() {
...
+                    if (problemRows.get(i)) problemRows.set(i, problemRows.get(i)!.concat(field));
+                    else problemRows.set(i, [field]);
+                }
+            }
+        }
+
+        if (problemRows.size > 0) {
+            let errorMessage = "Cannot submit. Required fields missing for rows:";
+            problemRows.forEach((fields, key) => {
+                const fieldStr = fields.toString().replaceAll(/\[\]/g, "");
```
kevinlul, line 88: Does this show something different in your browser? I expect `Array.toString()` to behave the same as `Array.join()` so we shouldn't have any square brackets in it to begin with.

frewmack: Actually I didn't consider using `.join()` for some reason. I'll do that instead.

kevinlul (approved):
LGTM

PR #237, opened 2020-11-27
React: Fix notes changing table row height

### LINKED ISSUE #231: Investigate handling longer notes in tables
Variable row height can be jarring when flipping between pages

frewmack:
Closes #231 
The component handles overflowing text in `notes` columns, and expands on mouse hover to show the entire entry.

### DIFF
--- a/react/src/pages/utils/components/Note.tsx
+++ b/react/src/pages/utils/components/Note.tsx
@@ -1,0 +1,25 @@
+import React from "react";
+import { makeStyles } from "@material-ui/core";
+
+const useStyles = makeStyles(theme => ({
+    notes: {
+        // TODO: Find a better way to set maxWidth using container width
+        maxWidth: "16em",
+        whiteSpace: "nowrap",
+        overflow: "hidden",
+        textOverflow: "ellipsis",
+        "&:hover": {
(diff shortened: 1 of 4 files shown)


react/src/pages/utils/components/Note.tsx
```diff
@@ -0,0 +1,25 @@
...
+const useStyles = makeStyles(theme => ({
+    notes: {
+        // TODO: Find a better way to set maxWidth using container width
+        maxWidth: "16em",
+        whiteSpace: "nowrap",
+        overflow: "hidden",
+        textOverflow: "ellipsis",
+        "&:hover": {
+            whiteSpace: "normal",
+            textOverflow: "default",
```
kevinlul, line 13: default isn't a style. You can use initial or inherit

kevinlul (changes requested):
This can cause some weird reflows on hover. I believe it's because there's no minimum width set but having one might just change this weird reflow point instead of fixing the problem.

Screen Shot 2020-11-30 at 09 25 39

Screen Shot 2020-11-30 at 09 25 52

kevinlul (approved):
LGTM

PR #238, opened 2020-11-30
React: dark mode

### LINKED ISSUE #205: Dark mode

annnag:
Closes #205

### DIFF
--- a/react/src/App.tsx
+++ b/react/src/App.tsx
@@ -77,7 +77,6 @@ function BaseApp(props: { darkMode: boolean; toggleDarkMode: () => void }) {
 
 export default function App() {
     const [darkMode, setDarkMode] = useState(localStorage.getItem("darkMode") === "true");
-    // const styles =
     const globalTheme = createMuiTheme({
         typography: {
             fontSize: 12,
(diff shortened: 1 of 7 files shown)


kevinlul (changes requested):
Looking good! Reading the [docs](https://material-ui.com/customization/palette/#user-preference) it seems like a good idea to wrap `createMuiTheme` and the Navigation `useStyles` in a `useMemo` hook so they aren't recomputed when the switch is flipped back and forth. We can also use `useMediaQuery` to set the default dark/light mode preference if we don't have a preference in localStorage yet.

react/src/App.tsx
```diff
@@ -81,10 +76,69 @@ function BaseApp() {
 }
 
 export default function App() {
+    const prefersDarkMode = useMediaQuery("(prefers-color-scheme: dark)");
+    console.log(prefersDarkMode);
```
kevinlul, line 80: remove

react/src/App.tsx
```diff
@@ -81,10 +76,69 @@ function BaseApp() {
...
+                          }
+                        : {},
+                },
+            }),
+        [darkMode]
+    );
+
+    useEffect(() => {
+        setDarkMode(prefersDarkMode);
+    }, [prefersDarkMode]);
```
kevinlul, line 130: This will always override the user's saved preference with the system preference. You already take care of this logic when initializing the state above so this isn't needed either.

kevinlul (changes requested):

kevinlul (approved):
LGTM

PR #239, opened 2020-11-30
React: Save default columns for Add Participants page

### LINKED ISSUE #229: Add participants: default columns
- [x] save the user's selected columns to localStorage but allow resetting to default
- [x] optionally configure the default selected columns in `.env` or similar so it can be changed per deployed instance

frewmack:
Closes #229


react/src/pages/addParticipants/DataEntryTable.tsx
```diff
@@ -294,6 +341,11 @@ function DataEntryToolbar(props: {
                     columns={props.columns}
                     onClick={props.handleColumnAction}
                 />
+                <Tooltip title="Reset column defaults">
+                    <IconButton onClick={() => props.handleResetAction()}>
```
kevinlul, line 345:
```suggestion
                    <IconButton onClick={props.handleResetAction}>
```

react/src/pages/addParticipants/DataEntryTable.tsx
```diff
@@ -60,19 +60,55 @@ const useTableStyles = makeStyles(theme => ({
...
+        const validFields = getDataEntryHeaders().optional;
+        const envFields = envCols.split(",");
+        const usableFields = envFields.filter(
+            field => !!validFields.find(valid => valid === field)
+        );
+        return usableFields as Array<keyof DataEntryRowOptional>;
+    } else {
+        return null;
+    }
+}
```
kevinlul, line 78: I appreciate that we're verifying this configuration value. However, this code is running in the browser rather than at compile-time. We can't really afford to eject Create React App to add a webpack plugin that checks the environment variable. I think as a compromise we could leave this check to `NODE_ENV === "development"`. Also, returning an empty array here instead of `null` will probably make things either since there isn't a semantic difference between having a blank value and the absence of the envvar.

react/src/pages/addParticipants/DataEntryTable.tsx
```diff
@@ -60,19 +60,55 @@ const useTableStyles = makeStyles(theme => ({
...
 
-    const [optionals, setOptionals] = useState<DataEntryHeader[]>(
-        getColumns("optional").map(header => {
-            return { ...header, hidden: !defaultOptionals.includes(header.field) };
-        })
-    );
+    function getOptionalHeaders() {
+        return getColumns("optional").map(header => ({
+            ...header,
+            hidden: !getDefaultColumns(fallbackColumns).includes(header.field),
```
kevinlul, line 107: `getDefaultColumns` gets called every time, which means you are reparsing the same values and saving the same thing to localStorage for every optional column.

kevinlul (changes requested):

react/src/pages/addParticipants/DataEntryTable.tsx
```diff
@@ -45,19 +45,59 @@ const useTableStyles = makeStyles(theme => ({
...
+
+function getEnvColumns(): Array<keyof DataEntryRowOptional> {
+    if (process.env.NODE_ENV === "development") {
+        const envCols = process.env.REACT_APP_DEFAULT_OPTIONAL_COLUMNS;
+        if (envCols !== undefined) {
+            const validFields = getDataEntryHeaders().optional;
+            const envFields = envCols.split(",");
+            const usableFields = envFields.filter(
+                field => !!validFields.find(valid => valid === field)
+            );
```
kevinlul, line 58: The idea was that just this check is exclusive to dev to catch mistakes. Now production builds will always have `[]` and ignore `REACT_APP_DEFAULT_OPTIONAL_COLUMNS`.

frewmack: Ah my bad, I misunderstood.

kevinlul (approved):
LGTM

Issue #209, opened 2020-11-26
Fix all endpoints that return user PKs to return usernames


kevinlul:
e.g. https://github.com/ccmbioinfo/ST2020/blob/master/flask/app/analyses.py#L68

Update unit tests accordingly

delvinso:
- [x] rename user foreign keys and add user relationship to each table (#244)
- [x] fix unit tests and in turn, fix all endpoints previously referencing created_by and updated_by  (#244)
     - [x ] datasets, families, manage, models, participants, routes
- [x] fix current endpoints returning usernames (analyses and datasets)  (#244)
- [ ] fix endpoints returning pk to return usernames

PR #242, opened 2020-12-02
Flask: return usernames instead of primary keys

delvinso:
### Example for GET API/families

references #209 

currently does not return the username for children 

would adding the following relationships to the other tables (participants, datasets, tissue_samples) similar to what analyses has simplify this task? (https://github.com/ccmbioinfo/ST2020/blob/8616b549139d9f68946e051a9811d5472bfe8014/flask/app/analyses.py#L124)

```
assignee_user = db.relationship("User", foreign_keys=[assignee]) # not this one though
requester_user = db.relationship("User", foreign_keys=[requester])
updated_by_user = db.relationship("User", foreign_keys=[updated_by])
```

Right now, the query uses aliases for the Users table to join onto `created_by` and `updated_by` within families similar to https://github.com/ccmbioinfo/ST2020/blob/master/flask/app/analyses.py#L68 but this doesn't seem to be efficient with more children being added 

curl "localhost:5000/api/families?max_rows=1"

```
[
  {
    "created": "Tue, 24 Nov 2020 20:52:47 GMT", 
    "created_by": "admin", 
    "family_codename": "1000", 
    "family_id": 1, 
    "participants": [
      {
        "affected": true, 
        "created": "Tue, 24 Nov 2020 20:52:47 GMT", 
        "created_by": 1, 
        "month_of_birth": "Sat, 01 Jan 2000 00:00:00 GMT", 
        "notes": "Extra info about sample here", 
        "participant_codename": "AA001", 
        "participant_id": 1, 
        "participant_type": "Proband", 
        "sex": "Female", 
        "solved": null, 
        "updated": "Tue, 24 Nov 2020 20:52:47 GMT", 
        "updated_by": 1
      }
    ], 
    "updated": "Tue, 24 Nov 2020 20:52:47 GMT", 
    "updated_by": "admin"
  }
]
```


SamarthVP:
Hmmm, I see the problem. I'm not a fan of how the Analysis schema has `updated_by` and `updated_by_user` when using a simple one-to-one relationship should be enough: https://docs.sqlalchemy.org/en/13/orm/basic_relationships.html#one-to-one
There's probably a reason I don't know though...

The options to me seem to be
1) Replace foreign key columns for all `updated_by` and `created_by` with one-to-one relationships so we can use the ORM to easily access any properties we want to return. @kevinlul are there any downsides to setting up a relationship vs standard foreign key columns?
2) Switch from joining on user_id to username (which is also unique) for all `updated_by` and `created_by`. If we ever wanted to return other information we run into the same problem.
3) Go for the easy workaround of adding a separate query for the user from user_id each time we want to return username. Obviously not ideal, but easiest.

Regardless this means a lot of things need to change in a lot of places..... sooooo many changes...

kevinlul:
The `_user` relationships exist because the foreign keys were added to the schema first and they don't conform to the SQLAlchemy expectation of `relationship` `relationship_id`. It is still an ORM relationship though so it can be used like the rest. The name could certainly be changed as long as the migration from old Sample Tracker is also updated. We can't use foreign key on username because that's a user-entered value and subject to change.

SamarthVP:
Good point, in that case if we add the actual relationships to the User objects for the rest of the models, is there any reason to keep the id foreign key column? Because we could just get that from updated_by.user_id once we have the relationship set up.

kevinlul:
As long as @helrick is aware for ccmbioinfo/ST2020db#3 then there's no problem with changing the names so `name` is the relationship and the actual stored foreign key is `name_id`

kevinlul:
You'll need to merge master back in and redo the migration because one was added for the user-minio credentials in #243

delvinso:
Would it be easier to create a new branch from master since the draft changes to families.py won't be merged in anyways?

kevinlul:
Whatever works best for you!

Issue #193, opened 2020-11-20
Epic: /api/users


kevinlul:
Existing endpoints in `routes.py` should be moved out to a new `users.py`

- [x] Update **GET /api/users** (list_users)
  Admin-only. Returns an array of JSON objects of the username, email, admin status, last login, activation status, and an array of group codes.

- [x] **GET /api/users/:username** (get_user)
   Regular users may query themselves. Admins can query all users. Returns a JSON of the username, email, admin status, an array of group codes, last login, activation status, MinIO access key, and MinIO secret key. This is the same as list_users but also includes the MinIO credentials, which need to be added as fields to the User model.

- [x] Update **POST /api/users** (create_user)
  Admin-only. Creates a new user with the specified credentials, email, admin status, groups. Also creates a MinIO identity and assigns it to those groups. Return result matches GET /api/users/:username. Return the new URL in the Location header.

- [x] **PATCH /api/users/:username** (update_user)
  Replaces PUT /api/users and POST /api/password. If the requesting user is an admin, this updates the affected fields (excluding MinIO credentials). If the username is changed, return the new URL in the Location header. Deactivating users deletes the MinIO identity. Activating users recreates it. Fail the request with 404 if any of the new group_codes do not exist. The response should have the same format as get_user.
  If the requesting user is a regular user, the behaviour of this endpoint should match the old /api/password: they are only allowed to change their own password and must provide their old password as well.

- [x] **POST /api/users/:username** (reset_minio_user)
  Resets the MinIO credentials for this user (i.e. delete the MinIO user and create a new one with membership in the same groups) to a new randomly-generated pair and returns them. Regular users may do this for themselves and admins may do this for any user.

- [x] **DELETE /api/users/:username** (delete_user)
  Replaces DELETE /api/users. Admin-only. Deletes the user, including their MinIO identity. If anything in the database other than group membership depends on the user existing, fail the request with 422.

Issue #192, opened 2020-11-20
Epic: /api/groups


kevinlul:
Create a new file `groups.py` for these endpoints

- [x] **GET /api/groups** (list_groups)
  Requires authentication. Returns a list of all group codes and display names, e.g. `[{ "group_code": "ACH", "group_name": "Alberta" }]`

- [x] **GET /api/groups/:group_code** (get_group)
  Returns `{ "group_code": "", "group_name": "",  "users": [] }` where each field is filled out accordingly. The `users` array should be an array of usernames of the users in the group and can be obtained from the `users` relation on the Group model as a backref from User. Admins can use this endpoint for all groups. Regulars can only use this for groups they belong to or they get a 403 Forbidden.

- [x] **POST /api/groups** (create_group)
  Admin-only. Creates a new group with the specified codename, display name, and users (same format as the get_group endpoint return result, but the users array is optional). This also creates a corresponding MinIO group (via madmin.MinioAdmin) and bucket (via MinioClient). The group should have an access policy that gives it complete access to its bucket. The corresponding MinIO users should be added to this group. Fail with 422 if the bucket already exists (do not create a group in the database in this case) or the code/name are already in use. Fail with 404 if a user to be added doesn't exist. The return result should be of the same form as `get_group` and the URL to the created resource should be in the Location header.

- [x] **PATCH /api/groups** (update_groups)
  Admin-only. Updates requested fields accordingly. The group code cannot be changed but the display name can be. Users should be added and removed accordingly and the error codes are the same as create_group.

- [x] **DELETE /api/groups/:group_code** (delete_group)
  Admin-only. Deletes the group and the MinIO group if they are empty but retains the MinIO bucket, otherwise fail with 422.

PR #243, opened 2020-12-02
db: add MinIO credentials to User table

kevinlul:
Supports #192, #193


SamarthVP (approved):
Idk about the migration, but I like the columns

helrick (approved):

PR #244, opened 2020-12-03
Flask: migrations for *_by schema changes

delvinso:
related to #209


kevinlul (changes requested):
`assignee` means who the analysis is _assigned to_, not who assigned it, so preferably `assignee` and `requester` are left as-is for the relationship name and the foreign key is `assignee_id` and similar

kevinlul (commented):
Since the relationship and foreign key naming is what SQLAlchemy expects for a many-to-one relationship, I believe explicitly specifying the foreign key for each isn't necessary. Otherwise blocked by #199

delvinso:
> Since the relationship and foreign key naming is what SQLAlchemy expects for a many-to-one relationship, I believe explicitly specifying the foreign key for each isn't necessary. Otherwise blocked by #199

Error if the foreign key is omitted:

```
sqlalchemy.exc.AmbiguousForeignKeysError: Could not determine join condition between parent/child tables on relationship Family.updated_by - there are multiple foreign key paths linking the tables.  Specify the 'foreign_keys' argument, providing a list of those columns which should be counted as containing a foreign key reference to the parent table.

```

kevinlul:
Oops, I forgot that there are multiple user relationships...

delvinso:
@kevinlul @helrick  I think this is good to review now since all tests pass. Can this PR be merged first and the actual fixes to the endpoints be a separate PR? 

Tests pass:

```
Creating st2020test_app_run ... done
wait-for-it.sh: waiting for mysql:3306 without a timeout
wait-for-it.sh: mysql:3306 is available after 0 seconds
============================= test session starts ==============================
platform linux -- Python 3.7.9, pytest-6.1.2, py-1.9.0, pluggy-0.13.1
rootdir: /var/www/sample-tracker/flask
collected 47 items

tests/test_analyses.py ........                                          [ 17%]
tests/test_datasets.py ..........                                        [ 38%]
tests/test_families.py ........                                          [ 55%]
tests/test_login.py ..                                                   [ 59%]
tests/test_madmin.py ......                                              [ 72%]
tests/test_misc.py ...                                                   [ 78%]
tests/test_participants.py .......                                       [ 93%]
tests/test_samples.py ...                                                [100%]

=============================== warnings summary ===============================
tests/test_analyses.py::test_no_analyses
  /usr/local/lib/python3.7/site-packages/minio/copy_conditions.py:29: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3,and in 3.9 it will stop working
    from collections import MutableMapping

tests/test_datasets.py::test_create_dataset
  /usr/local/lib/python3.7/site-packages/sqlalchemy/orm/mapper.py:1691: SAWarning: Flushing object <Dataset at 0x7f283a8e0090> with incompatible polymorphic identity None; the object may not refresh and/or load correctly (this warning may be suppressed after 10 occurrences)
    (state_str(state), dict_[polymorphic_key]),

tests/test_datasets.py::test_create_dataset
  /usr/local/lib/python3.7/site-packages/sqlalchemy/orm/mapper.py:1691: SAWarning: Flushing object <Dataset at 0x7f283a73fcd0> with incompatible polymorphic identity None; the object may not refresh and/or load correctly (this warning may be suppressed after 10 occurrences)
    (state_str(state), dict_[polymorphic_key]),

tests/test_datasets.py::test_create_dataset
  /usr/local/lib/python3.7/site-packages/sqlalchemy/orm/mapper.py:1691: SAWarning: Flushing object <Dataset at 0x7f283a8c2b10> with incompatible polymorphic identity None; the object may not refresh and/or load correctly (this warning may be suppressed after 10 occurrences)
    (state_str(state), dict_[polymorphic_key]),

tests/test_families.py: 2 warnings
tests/test_misc.py: 85 warnings
  /usr/local/lib/python3.7/site-packages/sqlalchemy/orm/query.py:223: SADeprecationWarning: Plain string expression passed to Query() should be explicitly declared using literal_column(); automatic coercion of this value will be removed in SQLAlchemy 1.4
    entity_wrapper(self, ent)

-- Docs: https://docs.pytest.org/en/stable/warnings.html
======================= 47 passed, 91 warnings in 56.02s =======================
```

flask/app/config.py
```diff
@@ -23,3 +23,4 @@ class Config(object):
     )
     MINIO_ACCESS_KEY = os.getenv("MINIO_ACCESS_KEY", "AKIAIOSFODNN7EXAMPLE")
     TESTING = False
+    LOGIN_DISABLED = True
```
kevinlul, line 26: oops

kevinlul (changes requested):
Seems okay! The only thing I see is that I could try editing the Alembic migration to alter columns instead of dropping and adding new ones but since we're in development the gain of that is marginal at best.

kevinlul (approved):

SamarthVP (approved):
This look like it was maddening... Great work!!

PR #245, opened 2020-12-03
React: Fix 'Pipeline: undefined undefined' issue in analysis dialogs

### LINKED ISSUE #215: Post-migration, analysis dialog shows Pipeline: undefined undefined

frewmack:
Closes #215

Also fixes a bug where DetailSection fields can overlap the Edit Mode switch if the field is long enough.

### DIFF
--- a/react/src/pages/datasets/DatasetTable.tsx
+++ b/react/src/pages/datasets/DatasetTable.tsx
@@ -103,11 +103,15 @@ export default function DatasetTable() {
                     }}
                     onUpdate={(dataset_id: string, newDataset: { [key: string]: any }) => {
                         setDatasets(
-                            datasets.map(dataset =>
-                                dataset.dataset_id === dataset_id
-                                    ? { ...dataset, ...newDataset }
-                                    : dataset
-                            )
+                            datasets.map(dataset => {
+                                if (dataset.dataset_id === dataset_id) {
+                                    const updatedDataset = { ...dataset, ...newDataset };
+                                    setInfoDataset(updatedDataset);
+                                    return updatedDataset;
+                                } else {
+                                    return dataset;
+                                }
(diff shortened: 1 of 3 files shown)


react/src/pages/datasets/DatasetTable.tsx
```diff
@@ -103,11 +103,15 @@ export default function DatasetTable() {
...
                         setDatasets(
-                            datasets.map(dataset =>
-                                dataset.dataset_id === dataset_id
-                                    ? { ...dataset, ...newDataset }
-                                    : dataset
-                            )
+                            datasets.map(dataset => {
+                                if (dataset.dataset_id === dataset_id) {
+                                    const updatedDataset = { ...dataset, ...newDataset };
+                                    setInfoDataset(updatedDataset);
```
kevinlul, line 109: If I'm understanding right, this is because the dialog edits call onUpdate and this state is passed as a prop all the way down?

kevinlul (commented):
The gap to fix the overlapping bug looks pretty bad but given how the components are structured I don't know if we can move the switch into the dialog titlebar.

react/src/pages/datasets/DatasetTable.tsx
frewmack: Yep. The 'info dataset' has to be updated or else the contents of the dialog will not update post-edit.

frewmack:
> 
> 
> The gap to fix the overlapping bug looks pretty bad but given how the components are structured I don't know if we can move the switch into the dialog titlebar.

Another option could be to keep the switch in its original position, but place it inside a Grid that extends to the bottom of the section. I'll try that instead to see how it looks.

kevinlul (approved):
LGTM

PR #246, opened 2020-12-04
Flask: Groups endpoints

### LINKED ISSUE #192: Epic: /api/groups

SamarthVP:
closes #192 

Sample tests, will not work as expected unless User.minio_access_key and User.minio_secret_key are manually set to a manually created minio user:
```
curl -D - -X POST \
        --data '{"group_name" : "bruh20", "group_code" : "whoknows20",
                "users": [
                "admin"
                ]
                }' \
        localhost:5000/api/groups

curl -D - -X PATCH \
        --data '{"group_name" : "changed20.1",
                "users": [
                
                ]
                }' \
        localhost:5000/api/groups/whoknows20

curl -D - -X DELETE \
        localhost:5000/api/groups/whoknows20
```

### DIFF
--- a/flask/app/groups.py
+++ b/flask/app/groups.py
@@ -1,0 +1,64 @@
+from flask import abort, jsonify, request, Response, current_app as app
+from flask_login import login_user, logout_user, current_user, login_required
+from app import db, login, models
+from sqlalchemy.orm import contains_eager, joinedload
+from .routes import check_admin, transaction_or_abort
+
+
+@app.route("/api/groups", methods=["GET"])
+@login_required
+def list_groups():
+    if app.config.get("LOGIN_DISABLED") or current_user.is_admin:
+        user_id = request.args.get("user")
+    else:
+        user_id = current_user.user_id
+
+    if user_id:
+        groups = (
+            models.Group.query.join(models.Group.users)
+            .filter(models.User.user_id == user_id)
+            .all()
+        )
+    else:
+        groups = models.Group.query.all()
+
+    return jsonify(
+        [
+            {"group_code": group.group_code, "group_name": group.group_name}
+            for group in groups
+        ]
+    )
+
+
+@app.route("/api/groups/<string:group_code>", methods=["GET"])
+@login_required
+def get_group(group_code):
+    if app.config.get("LOGIN_DISABLED") or current_user.is_admin:
+        user_id = request.args.get("user")
+    else:
+        user_id = current_user.user_id
+
+    if user_id:
+        group = (
+            models.Group.query.filter(models.Group.group_code == group_code)
+            .join(models.Group.users)
+            .filter(models.User.user_id == user_id)
+            .one_or_none()
+        )
+    else:
+        group = models.Group.query.filter(
+            models.Group.group_code == group_code
+        ).one_or_none()
+
+    if group is None:
+        return "Forbidden", 403
+
+    return jsonify(
+        {
+            "group_code": group.group_code,
+            "group_name": group.group_name,
+            "users": [
+                user.username for user in group.users if user.deactivated == False
+            ],
+        }
+    )
--- a/flask/app/groups.py
+++ b/flask/app/groups.py
@@ -4,7 +4,6 @@ from app import db, login, models
 from sqlalchemy.orm import contains_eager, joinedload
 from .routes import check_admin, transaction_or_abort, mixin
 from minio import Minio
-from os import environ
 from .madmin import MinioAdmin, readwrite_buckets_policy
 
 
@@ -88,9 +87,9 @@ def update_group(group_code):
             group.group_name = request.json["group_name"]
 
     minioAdmin = MinioAdmin(
-        endpoint=environ["MINIO_ENDPOINT"],
-        access_key=environ["MINIO_ACCESS_KEY"],
-        secret_key=environ["MINIO_SECRET_KEY"],
+        endpoint=app.config["MINIO_ENDPOINT"],
+        access_key=app.config["MINIO_ACCESS_KEY"],
+        secret_key=app.config["MINIO_SECRET_KEY"],
     )
 
     # Check if a user to be added doesn't exist, 404
@@ -160,16 +159,16 @@ def create_group():
             group_obj.users.append(user)
 
     minioClient = Minio(
-        environ["MINIO_ENDPOINT"],
-        access_key=environ["MINIO_ACCESS_KEY"],
-        secret_key=environ["MINIO_SECRET_KEY"],
+        app.config["MINIO_ENDPOINT"],
+        access_key=app.config["MINIO_ACCESS_KEY"],
+        secret_key=app.config["MINIO_SECRET_KEY"],
         secure=False,
     )
 
     minioAdmin = MinioAdmin(
-        endpoint=environ["MINIO_ENDPOINT"],
-        access_key=environ["MINIO_ACCESS_KEY"],
-        secret_key=environ["MINIO_SECRET_KEY"],
+        endpoint=app.config["MINIO_ENDPOINT"],
+        access_key=app.config["MINIO_ACCESS_KEY"],
+        secret_key=app.config["MINIO_SECRET_KEY"],
     )
 
     # Create minio bucket via mc, 422 if it already exists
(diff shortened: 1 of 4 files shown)


flask/app/groups.py
```diff
@@ -0,0 +1,230 @@
...
+        if request.json["group_name"]:
+            # Check if display name is in use, 422
+            conflicting_group = models.Group.query.filter_by(
+                group_name=request.json["group_name"]
+            ).one_or_none()
+            if (conflicting_group is not None) and (conflicting_group != group):
+                return "Group name in use", 422
+            group.group_name = request.json["group_name"]
+
+    minioAdmin = MinioAdmin(
```
kevinlul, line 89: Please use snake_case since this is in Python

flask/app/groups.py
```diff
@@ -0,0 +1,230 @@
...
+        ).all()
+        # Make sure all users are valid
+        if len(users) != len(request.json["users"]):
+            return "Invalid username provided", 404
+
+        # Clear users from db and minio group
+        if len(group.users) != 0:
+            group_info = minioAdmin.get_group(group_code)
+            for old_user in group_info["members"]:
+                minioAdmin.group_remove(group_code, old_user)
```
kevinlul, line 108: This function supports passing the entire list of users in at once if you use the spread operator, i.e. `*group_info["members"]`

flask/app/groups.py
```diff
@@ -0,0 +1,230 @@
...
+            group.users[:] = []
```
kevinlul, line 109: What's the use of the slice operator here?

flask/app/groups.py
```diff
@@ -0,0 +1,230 @@
...
+            minioAdmin.group_add(group_code, user.minio_access_key)
+        # Reset policy for group in case the group did not exist
+        minioAdmin.set_policy(group_code, group=group_code)
+
+    try:
+        db.session.commit()
+        return jsonify(group)
+    except:
+        db.session.rollback()
+        return "Server error", 500
```
kevinlul, line 124: Hope this never happens because we can't roll back the MinIO side in the same way 😓

flask/app/groups.py
```diff
@@ -0,0 +1,230 @@
...
+
+
+@app.route("/api/groups", methods=["POST"])
+@login_required
+@check_admin
+def create_group():
+    if not request.json:
+        return "Request body must be JSON", 400
```
kevinlul, line 132: Should be 415

flask/app/groups.py
```diff
@@ -0,0 +1,230 @@
...
+
+    if not group_name:
+        return "A group display name must be provided", 400
+    if not group_code:
+        return "A group codename must be provided", 400
+
+    if models.Group.query.filter(
+        (models.Group.group_code == group_code)
+        | (models.Group.group_name == group_name)
+    ).value("group_id"):
```
kevinlul, line 145: This is interesting that you can use the pipe operator like this!

flask/app/groups.py
```diff
@@ -0,0 +1,230 @@
...
+    # Add users to the db if they were given
+    if request.json.get("users"):
+        users = models.User.query.filter(
+            models.User.username.in_(request.json["users"])
+        ).all()
+        # Make sure all users are valid
+        if len(users) != len(request.json["users"]):
+            return "Invalid username provided", 404
+        for user in users:
+            group_obj.users.append(user)
```
kevinlul, line 159: as an optional thing, could you check if Python list concatenation works with SQLAlchemy collections?

flask/app/groups.py
```diff
@@ -0,0 +1,230 @@
...
+        endpoint=app.config["MINIO_ENDPOINT"],
+        access_key=app.config["MINIO_ACCESS_KEY"],
+        secret_key=app.config["MINIO_SECRET_KEY"],
+    )
+
+    # Create minio bucket via mc, 422 if it already exists
+    if minioClient.bucket_exists(group_code):
+        return "Minio bucket already exists", 422
+
+    minioClient.make_bucket(group_code)
```
kevinlul, line 178: What happens when the `group_code` contains invalid characters or is too short to be a bucket name? Does the SDK raise an exception?

kevinlul (changes requested):

flask/app/groups.py
SamarthVP: Old habits lol

flask/app/groups.py
SamarthVP: :O i didn't know this was a thing

flask/app/groups.py
SamarthVP: It's unnecessary, good catch

flask/app/groups.py
SamarthVP: Fixed

flask/app/groups.py
SamarthVP: Wow it actually does, that's much cleaner

flask/app/groups.py
SamarthVP: I believe filter() just takes any conditional statement so | is treated as the or operator

flask/app/groups.py
SamarthVP: True, both `bucket_exists()` and `make_bucket()` would throw errors which we should definitely be checking for. I've moved those commands closer to the beginning and if they error I return "Invalid bucket name", 422. I've also added a test case to test for this.

flask/tests/test_groups.py
```diff
@@ -0,0 +1,264 @@
+import pytest
+from app import db, models
+from sqlalchemy.orm import joinedload
+from conftest import TestConfig
+from app.madmin import MinioAdmin, readwrite_buckets_policy
+from minio import Minio
+
+
+@pytest.fixture
+def minioAdmin():
```
kevinlul, line 10: snake_case

flask/tests/test_groups.py
```diff
@@ -0,0 +1,264 @@
...
+    )
+    # Test group_code already in use
+    assert (
+        client.post(
+            "/api/groups",
+            json={"group_code": "ach", "group_name": "Nothing suspicious"},
+        ).status_code
+        == 422
+    )
+    # Test invlaid bucket name
```
kevinlul, line 203: typo

flask/tests/test_groups.py
```diff
@@ -0,0 +1,264 @@
...
+    # Test existence
+    assert (
+        client.patch("/api/groups/code", json={"group_name": "yes"}).status_code == 404
+    )
+
+    # Test success and check db and minio
+    ad = models.User.query.filter(models.User.username == "admin").one_or_none()
+    ad.minio_access_key = "admin"
+    user = models.User.query.filter(models.User.username == "user").one_or_none()
+    user.minio_access_key = "user"
```
kevinlul, line 155: This might be better for the test_database fixture instead of making use of the shared SQLAlchemy session here.

kevinlul (changes requested):
Overall, this looks really good!

flask/tests/test_groups.py
SamarthVP: Good point, i thought it might break other tests but it should not and does not.

kevinlul (approved):
Solid!

PR #247, opened 2020-12-04
nginx: forward MinIO on port 9000, other production fixes

kevinlul:
nginx: [forward MinIO](https://docs.min.io/docs/setup-nginx-proxy-with-minio.html) on port 9000 again
nginx: fix public certificate path to use bundle with full certificate chain
compose: use latest version 3.8
compose: use correct prod app Dockerfile and fix CMD syntax


helrick (approved):
lgtm!

PR #249, opened 2020-12-08
React: Components for handling boolean/null in material-table

### LINKED ISSUE #232: Participants: null and false display the same for affected/solved
This effect is most apparent with the migrated database where you can group up participants by either column but you can visualize it yourself by adding some more examples to your test database.

The solution is _probably_ to use a dropdown and display text instead because a four-state boolean for the heading filter seems weird.

frewmack:
Closes #232. 
Boolean values are now handled like strings on the front-end, since material-table doesn't let you use custom render components for falsy values. Incoming boolean values are converted to `PseudoBoolean` strings, and outgoing `PseudoBoolean` strings are converted to boolean.

Also fixes an issue where `ParticipantTable` and `DatasetTable` send the entire row when editing from the table. Now it only sends the diff compared to the row pre-edit.

EDIT: Changes made to `DetailSection` are temporary. I intend to refactor that collection of components later.


react/src/pages/utils/components/BooleanDisplay.tsx
```diff
@@ -0,0 +1,32 @@
...
+        } else {
+            switchValue = props.value;
+        }
+
+        setValue(PseudoBooleanReadableMap[switchValue]);
+    }, [props.type, props.value, props.fieldName]);
+
+    return (
+        <>
+            <b>{value}</b>
```
annnag, line 29: I think the boolean columns don't need to be bold

react/src/pages/utils/components/BooleanFilter.tsx
```diff
@@ -0,0 +1,37 @@
...
+                value={value}
+                onChange={e => {
+                    setValue(e.target.value as PseudoBoolean);
+                    props.onFilterChanged((props.columnDef as any).tableData.id, e.target.value);
+                }}
+            >
+                <MenuItem value="" key="blank">
+                    (Blank)
+                </MenuItem>
+                {options.map(key => (
```
annnag, line 29: Not sure why we have "Not specified" in the filter. It's not one of the options that the user can select so it's a bit confusing.

react/src/pages/utils/components/BooleanDisplay.tsx
```diff
@@ -0,0 +1,32 @@
...
+ */
+export default function BooleanDisplay(props: {
+    value: any | PseudoBoolean;
+    fieldName: any;
+    type: "row" | "group";
+}) {
+    const [value, setValue] = useState<string>("");
+
+    useEffect(() => {
+        let switchValue: PseudoBoolean;
```
annnag, line 16: Maybe this is better?
```suggestion
       const switchValue: PseudoBoolean = props.type === "row" ? (props.value as any)[props.fieldName] : props.value
```

annnag (changes requested):
Looks good! Everything seems to be working!

react/src/pages/utils/components/BooleanFilter.tsx
frewmack: It depends if it's possible for the backend to send `undefined` as a value for these columns. Any editing on the frontend only allows `true`, `false`, and `null`, but we allow filtering for `undefined` as well.

annnag: So to the users, is there a difference between "not specified" and "unknown"?

kevinlul: There is because the default state of the filter is to not filter at all, as opposed to filtering specifically for the unspecified null state.

annnag: Do you mean "not specified" = no filter applied? I thought the "(blank)" option represents the no filter state.

frewmack: Just to clear things up, these are the possible filter values:
- "(Blank)": no filter applied, show all rows
- "Yes": filter for `true` values
- "No": filter for `false` values
- "Unknown": filter for `null` values
- "Not specified": filter for `undefined` values

As far as semantic differences, "Unknown" indicates that the user filled out the field, and specifically said that they don't know that the field is "Yes" or "No". "Not specified" indicates that the field is not filled out (maybe leftover after migrations from old sample tracker or something).

annnag: Okay, that makes sense. Thanks for the clarification!

annnag (approved):

react/src/pages/utils/components/BooleanFilter.tsx
kevinlul: A null field in the database is None in Python and should be null in JSON. `undefined` isn't a value in JSON, only JavaScript. I thought you were using `undefined` to indicate there was no filter applied but this actually has one more state than required then 👀

frewmack: Ah gotcha. I can remove it then.

kevinlul:
Merge conflict, otherwise good to go

PR #252, opened 2020-12-08
Flask & React: Display pipeline name, version instead of id on analyses page

### LINKED ISSUE #234: Display pipeline name-version instead of meaningless id on analysis page

frewmack:
Closes #234

### DIFF
--- a/react/src/pages/utils/typings.tsx
+++ b/react/src/pages/utils/typings.tsx
@@ -93,7 +93,6 @@ export interface Analysis {
     finished: string;
     updated_by: number;
     pipeline?: Pipeline;
-    pipeline_name?: string;
 }
 export interface Pipeline {
     pipeline_id: number;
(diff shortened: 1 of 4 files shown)


react/src/pages/utils/typings.tsx
```diff
@@ -92,6 +92,8 @@ export interface Analysis {
     started: string;
     finished: string;
     updated_by: number;
+    pipeline?: Pipeline;
+    pipeline_name?: string;
```
kevinlul, line 96: I see this is because the former is returned by GET /api/analyses/:id and the second now by /api/analyses. Could it be worth it to just join to Pipeline in the list endpoint and return it like in the individual endpoint and leave it up to the frontend to format the display name? Can material-table do that with its render props?

frewmack: I can do that. I think that's how I originally had it in the first commit.

kevinlul (approved):
LGTM

Issue #230, opened 2020-11-26
Support multiple files per dataset


kevinlul:
- [x] #248 
- [x] #253 
- [ ] Render multiple files in datasets on frontend
- [ ] Hook up modified frontend to backend (data entry and editing)

PR #253, opened 2020-12-09
Flask: support linking multiple files per dataset

kevinlul:
Add DatasetFile table and convert `input_hpf_path` to files relationship. Render to clients as `linked_files` array.
/api/_bulk now accepts pipe-delimited `|` lists for `linked_files` (renamed from `input_hpf_path`) for CSVs and arrays for JSONs.
Analysis.result_hpf_path -> Analysis.result_path

TODO
- [x] add /api/_bulk tests with multiple files
- [x] rename result_hpf_path to result_path on frontend

#230


kevinlul:
Had to fix up /api/_bulk a bit because it wasn't accepting multiple datasets for a participant for various reasons but it should be together now.

flask/app/routes.py
```diff
@@ -341,116 +340,99 @@ def bulk_update():
...
-
-        db.session.add(ptp_objs)
-        transaction_or_abort(db.session.flush)
-
-        # tissue logic
-
-        tis_query = models.TissueSample.query.filter(
-            models.TissueSample.participant_id == ptp_query.value("participant_id")
-        )
-
```
delvinso, line 400: So the check (398) for an existing tissue sample under a participant was causing additional tissue samples to fail?

kevinlul: What happened here was that if the participant already had a tissue sample then it would not proceed further to add any data to do with this row

delvinso: Good catch!

delvinso:
Looks great!

One thing I noticed while playing around with the test json - sending a request that fails (returns 400 bad request) results in the subsequent dataset_id to increase despite the previous request failing. Is this expected behaviour? I can provide the steps to reproduce this if needed

kevinlul:
That's kind of something that happens with a transaction in that it reserves the primary keys and if the transaction is rolled back, the autoincrement key is already ahead. So yes, it's expected behaviour with autoinc PKs

delvinso (approved):

PR #254, opened 2020-12-09
React: autopopulate if selecting existing participants on add dataset page

### LINKED ISSUE #224: Add participants: autopopulate if selecting existing participants to avoid errors
Maybe we could even disable the cells that already have known values?

annnag:
Closes #224 
Expected behavior: if the user selects an existing participant, the participant associated columns in the row will be auto-populated and disabled.


react/src/pages/addParticipants/DataEntryTable.tsx
```diff
@@ -101,6 +119,10 @@ export default function DataEntryTable(props: DataEntryTableProps) {
     const [files, setFiles] = useState<string[]>([]);
 
     const [showRNA, setShowRNA] = useState<boolean>(false);
+    // rows where participant associated columns are disabled
+    const [disabledRows, setDisabledRows] = useState<number[]>([]);
```
frewmack, line 123: Storing the row index of disabled rows separately will lead to issues since the number of rows changes dynamically. Duplicates of disabled rows are not disabled, and deleting a disabled row in the middle of the table will cause the next row to be disabled when it takes the deleted row's spot.

I would suggest instead to adding a disabled field to the DataEntryRow interface, since whether a row is disabled or not makes sense as a property of the row.

react/src/pages/addParticipants/DataEntryTable.tsx
```diff
@@ -138,7 +165,23 @@ export default function DataEntryTable(props: DataEntryTableProps) {
...
-            if (index === rowIndex) {
+            if (autopopulate && index === rowIndex) {
+                const participant = findParticipant(newValue as string, col.field, value, families);
+                if (participant) {
+                    setDisabledRows([...disabledRows, rowIndex]);
+                    return setProp(
+                        participantCols.reduce(
+                            (row, currCol) => setProp(row, currCol, participant[currCol]),
+                            { ...value }
+                        ),
```
frewmack, line 176: Clever!

frewmack (changes requested):
Autopopulate works like a charm. Just row disabling needs work

kevinlul (commented):
Definitely getting autopopulating, which is good, but they aren't being disabled.

frewmack (approved):
Looks good to me!

PR #241, opened 2020-12-02
PATCH /api/tissue_samples/:id

Madelinehazel:
closes #210 
holding off on writing unit tests until #199 is merged

testing example: 
```
curl -X PATCH \
    -H "Content-Type: application/json"  \
    --data '{
         "tissue_sample_type": "Blood",
         "notes": "hello"
        }' \
    localhost:3000/api/tissue_samples/1
```

PR #261, opened 2020-12-10
Flask: return usernames for all endpoints

### LINKED ISSUE #209: Fix all endpoints that return user PKs to return usernames

delvinso:
Closes #209
Closes #242 
blocked by #241?

To be resolved separately: ~~GET endpoints (NOT /api/*/<int:id>) also return usernames for all children~~

### analyses.py
    * /api/analyses 
      * GET [x]
      * POST [x]
    * /api/analyses/<int:id>
      * GET [x]
      * DELETE [NA]
      * PATCH [x]
    
### families.py
    * /api/families  
      * GET [x]
      * POST [x]
    * /api/families/<int:id>
      * GET [x]
      * DELETE [NA]
      * PATCH [X]
    
### participants.py
    * /api/participants 
      * GET [x]
      * POST [x]
    * /api/participants/<int:id>
      * GET [x]
      * DELETE [NA]
      * PATCH [x]
    
### datasets.py
    * /api/datasets 
      * GET [x]
      * POST [x]
    * /api/datasets/<int:id>
      * GET [x]
      * DELETE [NA]
      * PATCH [x]

  # tissue_samples.py
    * /api/datasets 
      * GET [NA]
      * POST [x]
    * /api/datasets/<int:id>
      * GET [x]
      * DELETE [NA]
      * PATCH [x]


kevinlul:
Does this replace #242?

delvinso:
Yes, part of #242 also included some of the migrations but those ended up being split off

flask/app/families.py
```diff
@@ -200,7 +218,15 @@ def update_family(id: int):
...
         db.session.commit()
-        return jsonify(family)
+        return jsonify(
+            [
+                {
+                    **asdict(family),
+                    "updated_by": family.updated_by_id and family.updated_by.username,
+                    "created_by": family.created_by_id and family.created_by.username,
+                }
+            ]
```
kevinlul, line 228: This should be a single object, not an array

flask/app/families.py
```diff
@@ -241,4 +267,18 @@ def create_family():
...
+        jsonify(
+            [
+                {
+                    **asdict(fam_objs),
+                    "updated_by": fam_objs.updated_by_id
+                    and fam_objs.updated_by.username,
+                    "created_by": fam_objs.created_by_id
+                    and fam_objs.created_by.username,
+                }
+            ]
```
kevinlul, line 280: This should be a single object, not an array

flask/app/participants.py
```diff
@@ -231,4 +249,18 @@ def create_participant():
...
+        jsonify(
+            [
+                {
+                    **asdict(ptp_objs),
+                    "created_by": ptp_objs.created_by_id
+                    and ptp_objs.created_by.username,
+                    "updated_by": ptp_objs.updated_by_id
+                    and ptp_objs.updated_by.username,
+                }
+            ]
```
kevinlul, line 262: This should be a single object, not an array

kevinlul (changes requested):

flask/app/analyses.py
```diff
@@ -124,6 +124,10 @@ def get_analysis(id: int):
                     "participant_type": dataset.tissue_sample.participant.participant_type,
                     "sex": dataset.tissue_sample.participant.sex,
                     "family_codename": dataset.tissue_sample.participant.family.family_codename,
+                    "updated_by": dataset.tissue_sample.updated_by_id
+                    and dataset.tissue_sample.updated_by.username,
```
SamarthVP, line 128: Is the purpose of `and` here and everywhere else to make sure the .username doesn't throw an error if `*_by` is undefined?

delvinso: sorry didn't see this yesterday, I believe so since the username can be returned without the id

kevinlul (changes requested):
I think you need to add the user relationships to joinedload or you will end up making N+1 SELECT statements on serialization. You should also remove the Python type annotations from the foreign key columns so they don't get serialized by jsonify. Lastly, the null check isn't necessary if the database foreign key column is not nullable.

delvinso:
Got it, out of curiosity is there a way to format the SQL queries coming out of the logs?

kevinlul:
I found out about it from [here](https://docs.sqlalchemy.org/en/13/core/engines.html#configuring-logging). What are you looking for?

delvinso:
It's not that important, just formatting to make the queries human readable similar to the examples in their documentation 

```
SELECT
    addresses_1.id AS addresses_1_id,
    addresses_1.email_address AS addresses_1_email_address,
    addresses_1.user_id AS addresses_1_user_id,
    users.id AS users_id, users.name AS users_name,
    users.fullname AS users_fullname,
    users.nickname AS users_nickname
FROM users JOIN addresses
    ON users.id = addresses.user_id
LEFT OUTER JOIN addresses AS addresses_1
    ON users.id = addresses_1.user_id
WHERE users.name = ? AND addresses.email_address = ?
['jack', 'someaddress@foo.com']
```

delvinso:
> I think you need to add the user relationships to joinedload or you will end up making N+1 SELECT statements on serialization. You should also remove the Python type annotations from the foreign key columns so they don't get serialized by jsonify. Lastly, the null check isn't necessary if the database foreign key column is not nullable.

Does this also apply to `contains_eager` in the permissions?

flask/app/families.py
```diff
@@ -114,13 +128,19 @@ def get_family(id: int):
         [
             {
                 **asdict(family),
+                "updated_by": family.updated_by.username,
+                "created_by": family.created_by.username,
                 "participants": [
                     {
                         **asdict(participants),
```
kevinlul, line 135: Can you change this to be in the singular?

flask/app/families.py
```diff
@@ -114,13 +128,19 @@ def get_family(id: int):
...
+                        "updated_by": participants.updated_by.username,
+                        "created_by": participants.created_by.username,
                         "tissue_samples": [
                             {
                                 **asdict(tissue_samples),
```
kevinlul, line 140: Ditto

flask/app/tissue_samples.py
```diff
@@ -143,7 +172,11 @@ def update_tissue_sample(id: int):
     if user_id:
         tissue_sample = (
             models.TissueSample.query.filter_by(tissue_sample_id=id)
-            .options(contains_eager(models.TissueSample.datasets))
+            .options(
+                contains_eager(models.TissueSample.datasets),
+                # contains_eager(models.TissueSample.created_by),
+                # contains_eager(models.TissueSample.updated_by),
```
kevinlul, line 178: I had a careful look and I think it's okay for this PR if we don't return the users for the nested objects. If needed we can do it in a more focused PR.

kevinlul (changes requested):

flask/app/tissue_samples.py
delvinso: Right now the nested objects do return users, I thought the issue was that n+ 1 select statements are being generated as a result of inefficient joining?

eg. tissue samples

```
curl localhost:5000/api/tissue_samples/1\?user=1
{
  "created": "Mon, 07 Dec 2020 15:43:25 GMT", 
  "created_by": "admin", 
  "datasets": [
    {
      "batch_id": null, 
      "capture_kit": null, 
      "condition": "GermLine", 
      "created": "Mon, 03 Feb 2020 00:00:00 GMT", 
      "created_by": "admin", 
      "dataset_id": 1, 
      "dataset_type": "RGS", 
      "extraction_protocol": "Something", 
      "input_hpf_path": null, 
      "library_prep_date": null, 
      "library_prep_method": null, 
      "notes": "hello", 
      "read_length": null, 
      "read_type": null, 
      "sequencing_centre": "CHEO", 
      "sequencing_date": null, 
      "sequencing_id": null, 
      "tissue_sample_id": 1, 
      "updated": "Wed, 09 Dec 2020 19:45:55 GMT", 
      "updated_by": "admin"
    }
  ], 
  "extraction_date": "Sat, 04 Jan 2020 00:00:00 GMT", 
  "notes": "hello", 
  "participant_id": 1, 
  "tissue_processing": null, 
  "tissue_sample_id": 1, 
  "updated": "Fri, 11 Dec 2020 20:29:31 GMT", 
  "updated_by": "admin"
}
```

flask/app/tissue_samples.py
```diff
@@ -54,7 +62,16 @@ def get_tissue_sample(id: int):
...
+            "created_by": tissue_sample.created_by.username,
+            "updated_by": tissue_sample.updated_by.username,
+            "datasets": [
+                {
+                    **asdict(dataset),
+                    "created_by": dataset.created_by.username,
+                    "updated_by": dataset.updated_by.username,
+                }
+                for dataset in tissue_sample.datasets
+            ],
```
kevinlul, line 74: Yeah, this segment here does perform two additional queries per dataset, so can we leave it as is as `"datasets": tissue_sample.datasets`

flask/app/tissue_samples.py
```diff
@@ -26,7 +26,11 @@ def get_tissue_sample(id: int):
     if user_id:
         tissue_sample = (
             models.TissueSample.query.filter_by(tissue_sample_id=id)
-            .options(contains_eager(models.TissueSample.datasets))
+            .options(
+                contains_eager(models.TissueSample.datasets),
+                contains_eager(models.TissueSample.created_by),
+                contains_eager(models.TissueSample.updated_by),
```
kevinlul, line 32: These two should be `joinedload` since we don't explicitly join to the user table for these relationships below

flask/app/participants.py
```diff
@@ -93,10 +95,14 @@ def list_participants():
...
                 **asdict(participant),
                 "family_codename": participant.family.family_codename,
+                "updated_by": participant.updated_by.username,
+                "created_by": participant.created_by.username,
                 "tissue_samples": [
                     {
                         **asdict(tissue_sample),
                         "datasets": tissue_sample.datasets,
+                        "updated_by": tissue_sample.updated_by.username,
+                        "created_by": tissue_sample.created_by.username,
```
kevinlul, line 105: Take these out to avoid extra queries for now

kevinlul (changes requested):

flask/app/families.py
kevinlul: Sorry Delvin, I didn't mean the name of the of the list, but the variable being used in the list `**asdict(participants)` !

kevinlul (approved):

SamarthVP (approved):

PR #266, opened 2020-12-14
React: Detail Dialog refactor

frewmack:
- Components associated with `DetailSection` are now separated into different files and broken down further.
  - `GridFieldsDisplay` moved to separate file.
  - editable `TextField` overwrite in `DetailSection` moved to file with `FieldDisplayEditable`.
  - `FieldDisplay` moved to separate file.
- Above components updated to take `enums` as props, as opposed to fetching enums for each element that needs them. May be worth looking into making a custom "useEnums" hook in the future.
- State management should be more clear now 
  - Dialog holds the state of the selected 'record' (dataset, analysis, etc.).
  - Child elements are passed props with this state and a way to update it. 
  - `DetailSection` uses state to track edit mode changes that may be discarded.

### DIFF
--- a/react/src/pages/utils/components/GridFieldsDisplay.tsx
+++ b/react/src/pages/utils/components/GridFieldsDisplay.tsx
@@ -3,16 +3,21 @@ import { Grid } from "@material-ui/core";
 import { Field } from "../typings";
 import FieldDisplayEditable from "./FieldDisplayEditable";
 
-const infoWidth = 6;
+// Yes, really
+type Width = 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12;
 
 /**
- * Display a collection of Fields in two columns in the specified order.
+ * Display a collection of editable fields. Number of columns can be specified.
+ *
+ * @param orderPriority How to display the fields (left-right first or top-down first)
+ * @param columnWidth The width of each column (out of 12, must be an integer)
  */
 export default function GridFieldsDisplay(props: {
     fields: Field[];
     editMode: boolean;
     onEdit: (fieldName: string | undefined, value: any) => void;
     orderPriority?: "left-right" | "top-down";
+    columnWidth?: Width;
 }) {
     const [enums, setEnums] = useState(undefined);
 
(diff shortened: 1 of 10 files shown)


kevinlul:
I think the conflict here is from the just merged #232.

react/src/pages/utils/components/GridFieldsDisplay.tsx
```diff
@@ -0,0 +1,61 @@
...
+    editMode: boolean;
+    onEdit: (fieldName: string | undefined, value: any) => void;
+    enums: any;
+    orderPriority?: "left-right" | "top-down";
+    columnWidth?: Width;
+}) {
+    const infoWidth = props.columnWidth || 6;
+    const numColumns = Math.floor(12 / infoWidth);
+
+    let fieldColumns: Field[][] = [...Array(numColumns)].map(() => []);
```
kevinlul, line 26: You don't need to copy an array that you just constructed and are discarding anyway for the `map`

react/src/pages/utils/components/GridFieldsDisplay.tsx
```diff
@@ -0,0 +1,61 @@
+import React from "react";
+import { Grid, GridProps } from "@material-ui/core";
+import { Field } from "../typings";
+import FieldDisplayEditable from "./FieldDisplayEditable";
+
+// Grids are very picky about what sizes are allowed
+type Width = Exclude<GridProps["xs"], boolean | "auto" | undefined>;
```
kevinlul, line 7: Does `GridProps["xs"]` not work even though that's the defined type?

react/src/pages/utils/components/GridFieldsDisplay.tsx
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * @param orderPriority How to display the fields (left-right first or top-down first)
+ * @param columnWidth The width of each column (out of 12, must be an integer)
+ */
+export default function GridFieldsDisplay(props: {
+    fields: Field[];
+    editMode: boolean;
+    onEdit: (fieldName: string | undefined, value: any) => void;
+    enums: any;
+    orderPriority?: "left-right" | "top-down";
```
kevinlul, line 20: This might be better as a default value `orderPriority = "top-down"` but that means the typing for the props will need to be separated or the arguments will look really confusing. Alternatively document in the comment above what the defaults are.

kevinlul (changes requested):

react/src/pages/utils/components/GridFieldsDisplay.tsx
frewmack: I meant to specify that only integers between 1 and 12 are allowed. I originally wrote `1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 | 12` for the type but decided to change it to this instead.

react/src/pages/utils/components/GridFieldsDisplay.tsx
frewmack: `Array(numColumns)` only sets the `length` property but doesn't put any elements in the array, so `map` doesn't work. `[...Array(numColumns)]` has `numColumns` undefined elements in it, so `map` works (`Array(x).fill(undefined)` would also work). Just JS things...

kevinlul: Yeah, and we can't do `Array(x).fill([])` or the same array will be in every slot 🔥

kevinlul (approved):

PR #267, opened 2020-12-15
Flask: update GET /api/users, move to users.py, add tests

kevinlul:
Now returns an array of JSON objects of the username, email, admin status, last login, activation status, and an array of group codes.
Also hook up the user list to use this updated endpoint.

#193

### DIFF
--- a/flask/tests/test_users.py
+++ b/flask/tests/test_users.py
@@ -1,0 +1,37 @@
+import pytest
+
+
+def test_list_users(test_database, client, login_as):
+    login_as("admin")
+    response = client.get("/api/users")
+    assert response.status_code == 200
+
+    user_list = response.get_json()
+    assert len(user_list) == 2
+
+    # return order is unspecified
+    user_list.sort(key=lambda user: user["username"])
+
+    assert user_list == [
+        {
+            "username": "admin",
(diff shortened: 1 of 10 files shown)


flask/tests/test_users.py
```diff
@@ -0,0 +1,37 @@
...
+def test_list_users(test_database, client, login_as):
+    login_as("admin")
+    response = client.get("/api/users")
+    assert response.status_code == 200
+
+    user_list = response.get_json()
+    assert len(user_list) == 2
+
+    # return order is unspecified
+    user_list.sort(key=lambda user: user["username"])
```
SamarthVP, line 13: Clever! Out of curiosity, does SQLAlchemy by default obey some kind of ordering or are collections returned in no particular order.

SamarthVP (approved):
Solid backend changes

flask/tests/test_users.py
kevinlul: SQLAlchemy should be returning in the same order as the underlying SQL, which is unspecified unless an ORDER BY clause is used. So the default ordering is completely arbitrary and probably the order the database has in its index, which is likely insertion order.

frewmack (approved):
Frontend looks good to me!

PR #268, opened 2020-12-15
Flask: add GET, POST /api/users/:username

kevinlul:
#193 

~~Waiting for rebasing the last two commits against master first after #267 merged~~

### DIFF
--- a/flask/app/users.py
+++ b/flask/app/users.py
@@ -1,5 +1,5 @@
 from flask import current_app as app, jsonify
-from flask_login import login_required
+from flask_login import current_user, login_required
 from sqlalchemy.orm import joinedload
 
 from . import models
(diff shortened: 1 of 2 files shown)


flask/app/users.py
```diff
@@ -24,3 +27,84 @@ def list_users():
...
+def reset_minio_user(username: str):
+    if (
+        not app.config.get("LOGIN_DISABLED")
+        and not current_user.is_admin
+        and username != current_user.username
+    ):
+        return "Unauthorized", 401
+
+    if not request.json:
+        return "Request body must be JSON", 415
```
SamarthVP, line 73: Is there a reason we care about the body at all?

flask/app/users.py
```diff
@@ -24,3 +27,84 @@ def list_users():
...
+    access_key = os.urandom(8).hex()  # 16 ASCII characters
+    secret_key = os.urandom(16).hex()  # 32 ASCII characters
+    # Probability of conflict is negligible and not considered
+    minio_admin.add_user(access_key, secret_key)
+    for group in user.groups:
+        # MinIO requires a user to exist to create the group so the access policy
+        # might not be set on a group if this is the first user to be added, however
+        # we can guarantee from POST /api/groups that the policy exists in MinIO
+        minio_admin.group_add(group.group_code, access_key)
+        minio_admin.set_policy(group.group_code, group=group.group_code)
```
SamarthVP, line 104: Nicely done!! This would have been very easy to get wrong

SamarthVP (approved):
Very well done overall

flask/app/users.py
kevinlul: We don't really and we can go through later and change all these instances where the request body doesn't matter. Instead of this check, which also fails if the body is `{}` or `[]` we can also check the Content-Type instead. My reasoning for this is that regular HTML forms can perform GET and POST requests cross-domain, but they can only submit `application/x-www-form-urlencoded` and `multipart/form-data`, not JSON, so it's just a way to deny those types of API accesses.

SamarthVP: Ohhh I see... yeah we should leave replacing most pointless checks with checking Content-Type header for a future sprint

Issue #219, opened 2020-11-26
Add global error handlers that return a JSON response


kevinlul:
Currently, we are using the default Flask error handling, which means that when an exception is thrown, or generally when we return a not-OK status code, the response type is HTML. To confirm to the expectations of our API, we should instead always return JSON with meaningful error messages, instead of burying the stack traces in HTML in development. Flask natively supports custom global error handlers in a similar fashion as the custom global datetime JSON encoder. This will allow us to `raise ValueError`, `raise TypeError`, etc. from endpoints and let the error handler return the appropriate status code and message.

PR #269, opened 2020-12-16
Flask & React: update POST /api/users, user creation form

kevinlul:
Now accepts admin status and groups, creates a MinIO identity, and return response more closely conforms with REST expectations.
Add conflicting checks for already-used usernames, emails. (Basic idea for #219)
Add defensive checks for blank usernames, emails, passwords.
Move password matching to frontend.
Refactor frontend user creation error handling, use wait cursor on submit, prevent multiple submission and submitting before finishing filling out the form
Use success variant snackbar.

#193

### DIFF
--- a/flask/app/users.py
+++ b/flask/app/users.py
@@ -1,12 +1,14 @@
 import os
 
-from flask import current_app as app, jsonify, request
 from flask_login import current_user, login_required
 from sqlalchemy.orm import joinedload
 
+from flask import current_app as app
+from flask import jsonify, request
+
 from . import db, models
-from .routes import check_admin, transaction_or_abort
 from .madmin import MinioAdmin
+from .routes import check_admin, transaction_or_abort
 
 
 @app.route("/api/users", methods=["GET"])
@@ -29,6 +31,21 @@ def list_users():
     )
 
 
+def jsonify_user(user: models.User):
+    return jsonify(
+        {
+            "username": user.username,
+            "email": user.email,
+            "is_admin": user.is_admin,
+            "last_login": user.last_login,
+            "deactivated": user.deactivated,
+            "groups": [group.group_code for group in user.groups],
+            "minio_access_key": user.minio_access_key,
+            "minio_secret_key": user.minio_secret_key,
+        }
+    )
+
+
 @app.route("/api/users/<string:username>", methods=["GET"])
 @login_required
 def get_user(username: str):
@@ -45,18 +62,33 @@ def get_user(username: str):
         .first_or_404()
     )
 
-    return jsonify(
-        {
-            "username": user.username,
-            "email": user.email,
-            "is_admin": user.is_admin,
-            "last_login": user.last_login,
-            "deactivated": user.deactivated,
-            "groups": [group.group_code for group in user.groups],
-            "minio_access_key": user.minio_access_key,
-            "minio_secret_key": user.minio_secret_key,
-        }
+    return jsonify_user(user)
+
+
+def reset_minio_credentials(user: models.User) -> None:
+    minio_admin = MinioAdmin(
+        endpoint=app.config["MINIO_ENDPOINT"],
+        access_key=app.config["MINIO_ACCESS_KEY"],
+        secret_key=app.config["MINIO_SECRET_KEY"],
     )
+    if user.minio_access_key:
+        try:
+            minio_admin.remove_user(user.minio_access_key)
+        except RuntimeError as err:
+            app.logger.warning(err.args[0])
+    # Generate cryptographically random pair
+    access_key = os.urandom(8).hex()  # 16 ASCII characters
+    secret_key = os.urandom(16).hex()  # 32 ASCII characters
+    # Probability of conflict is negligible and not considered
+    minio_admin.add_user(access_key, secret_key)
+    for group in user.groups:
+        # MinIO requires a user to exist to create the group so the access policy
+        # might not be set on a group if this is the first user to be added, however
+        # we can guarantee from POST /api/groups that the policy exists in MinIO
+        minio_admin.group_add(group.group_code, access_key)
+        minio_admin.set_policy(group.group_code, group=group.group_code)
+    user.minio_access_key = access_key
+    user.minio_secret_key = secret_key
 
 
 @app.route("/api/users/<string:username>", methods=["POST"])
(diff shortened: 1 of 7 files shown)


kevinlul:
Requesting Samarth for backend and Anna for frontend

flask/app/users.py
```diff
@@ -102,9 +88,80 @@ def reset_minio_user(username: str):
...
+
+    if not valid_strings(request.json, "username", "email", "password"):
+        return "Missing fields", 400
+
+    user = models.User.query.filter_by(username=request.json["username"]).first()
+    if user is not None:
+        return "User already exists", 422, {"location": f"/api/users/{user.username}"}
+    user = models.User.query.filter_by(email=request.json["email"]).first()
+    if user is not None:
+        return "Email already used", 422, {"location": f"/api/users/{user.username}"}
```
SamarthVP, line 146: Optional: Could be done nicer using `|` similar to https://github.com/ccmbioinfo/ST2020/blob/ec00df4054838d05a4bbc1d197acdaea052d04e2/flask/app/groups.py#L143

flask/app/users.py
```diff
@@ -102,9 +88,80 @@ def reset_minio_user(username: str):
...
+
+    user = models.User(
+        username=request.json["username"],
+        email=request.json["email"],
```
SamarthVP, line 150: Should probably check if `request.json["username]` and `request.json["email]` are valid. So len(username) <= 30, and len(email) <= 150 and maybe email is a valid email?

flask/app/users.py
```diff
@@ -102,9 +88,80 @@ def reset_minio_user(username: str):
...
+    user.set_password(request.json["password"])
+
+    requested_groups = request.json.get("groups")
+    if requested_groups:
+        groups = models.Group.query.filter(
+            models.Group.group_code.in_(requested_groups)
+        ).all()
+        if len(requested_groups) != len(groups):
+            return "Invalid group code provided", 404
+        user.groups += groups
```
SamarthVP, line 162: Solid

flask/app/users.py
```diff
@@ -59,39 +47,37 @@ def get_user(username: str):
...
     user = (
         models.User.query.options(joinedload(models.User.groups))
         .filter(models.User.username == username)
         .first_or_404()
     )
 
+    return jsonify_user(user)
+
+
+def reset_minio_credentials(user: models.User) -> None:
```
SamarthVP, line 69: Good idea

flask/tests/test_users.py
```diff
@@ -141,3 +141,99 @@ def test_reset_minio_user(test_database, minio_policy, client, login_as):
     assert client.post("/api/users/admin").status_code == 401
 
     assert_reset("user", client)
+
+
+def assert_new_user(body, client, login_as):
```
SamarthVP, line 146: Damn, this is more robust than a brick wall

flask/tests/test_users.py
```diff
@@ -141,3 +141,99 @@ def test_reset_minio_user(test_database, minio_policy, client, login_as):
...
+    assert client.post("/api/users", json={"xml": "bad"}).status_code == 401
+
+    login_as("admin")
+    assert client.post("/api/users").status_code == 415
+    assert client.post("/api/users", json={"xml": "bad"}).status_code == 400
+    assert client.post("/api/users", json={"username": "dormammu"}).status_code == 400
+    assert (
+        client.post(
+            "/api/users",
+            json={"username": "dormammu", "password": "I've come to bargain"},
```
SamarthVP, line 174: You've come to... wHaT Is tHIssSs

flask/tests/test_users.py
```diff
@@ -141,3 +141,99 @@ def test_reset_minio_user(test_database, minio_policy, client, login_as):
...
+        ).status_code
+        == 400
+    )
+
+    assert_new_user(
+        {
+            "username": "dormammu",
+            "password": "I've come to bargain",
+            "email": "stephen.strange@agamotto.org",
```
SamarthVP, line 183: For some reason, emails from this address resend every 60 seconds 🤔

SamarthVP (changes requested):
I am not even sure if this is a problem we should try and solve:
I just ran the tests and it worked............. but then I ran it again to be sure and got:
```
FAILED tests/test_groups.py::test_delete_group - AssertionError: assert 422 == 204
FAILED tests/test_madmin.py::test_add_remove_users - AssertionError: assert 5 == 0
```
Some tests don't seem to reset the state of minio, but it gets even worse. Since the CI doesn't test for the tests 'reseting state' I reran tests on master which gave the same error on the second run (The issue is not with this PR). We now need to consider whether this is worth debugging, because this could likely be an issue with how we are tearing down tests, in which case it would not affect prod, or it could be an issue with an endpoint slipping past the tests... which would be a problem :/

kevinlul:
I see. I'm not cleaning up the temporary users and because you keep the same MinIO instance between test runs (you can down the containers with `docker-compose -p st2020test -f docker-compose.test.yaml down`) then every access pair for each test stays. I can clean this up in the `minio_policy` fixture regardless of what keys were created.

SamarthVP:
Yup, that fixed it

SamarthVP (approved):
LGTM

react/src/Admin/components/CreateUserModal.tsx
```diff
@@ -19,55 +20,87 @@ export interface CreateUserModalProps {
...
+        case 404:
+            return (
+                <DialogContentText id={props.id} color="secondary">
+                    Some requested permission groups do not exist.
+                </DialogContentText>
+            );
+        case 422:
+            return (
+                <DialogContentText id={props.id} color="secondary">
+                    Username or email already in use.
```
annnag, line 48: I think it is really useful for users to know which field is already in use, so it would be nice if we can differentiate between email and username

annnag (changes requested):
Looks good!

react/src/Admin/components/CreateUserModal.tsx
SamarthVP: LOL revert part of 1038fc8

kevinlul: I went a slightly different route and added something that's like what we'll want #219 to do for all endpoints.

SamarthVP (approved):

annnag (approved):

PR #271, opened 2020-12-16
React: resizable table columns on add dataset page

### LINKED ISSUE #228: Add participants: set different column widths
Right now many optional columns are too narrow and all set to the same width, notably the input path. We need a flexible way to configure this per column and maybe optionally let users resize

annnag:
Closes #228


kevinlul (approved):
LGTM 🚀

frewmack (approved):
Looks good to me!

PR #272, opened 2020-12-17
React: use type disjunction in UserList reducer

kevinlul:
Fixes typecasts and makes the state transition system better typed


frewmack (approved):
This looks much more clear, nice!

PR #273, opened 2020-12-17
Updates to db requested during Demo

### LINKED ISSUE #207: Database updates requested from the demo
- [ ] ParticipantType.Other
- [ ] Participant.institution as either string, enum, or FK string to another table like DatasetType

helrick:
Closes #207


kevinlul (changes requested):
Can you update the test database in `conftest.py` to include one or two institutions so the tests are restored? Some tests in `tests/test_datasets.py` and `tests/test_misc.py` probably aren't expecting a new required field so that needs to be updated as well.

Lastly, I don't think the primary key should be the human readable name because then there's little point to having this separate table over just another string column. I would do a regular integer foreign key (institution_id) and then set up an ORM relationship like with `created_by`

helrick:
> Can you update the test database in `conftest.py` to include one or two institutions so the tests are restored? Some tests in `tests/test_datasets.py` and `tests/test_misc.py` probably aren't expecting a new required field so that needs to be updated as well.
> 
> Lastly, I don't think the primary key should be the human readable name because then there's little point to having this separate table over just another string column. I would do a regular integer foreign key (institution_id) and then set up an ORM relationship like with `created_by`

I guess the only benefit to having it in a separate table without using primary keys is that we have a table that provides a degree of control over the options? Easy to change but DatasetType also has this relationship without primary keys.

flask/utils/run.sh
```diff
@@ -16,6 +16,10 @@ elif [[ "$1" == "prod" ]]; then
     $COMMAND db upgrade
     gunicorn wsgi:app "$@"
 else
+    if [ ! -d "migrations" ]; then
+        $COMMAND db init
+        $COMMAND db migrate -m "Initial migration."
+    fi
```
kevinlul, line 22: Interesting approach!

flask/app/models.py
```diff
@@ -89,24 +88,23 @@ class ParticipantType(str, Enum):
...
     sex: Sex = db.Column(db.Enum(Sex))
-    # Sample.SampleType
     participant_type: ParticipantType = db.Column(db.Enum(ParticipantType))
     month_of_birth: date = db.Column(
         db.Date, CheckConstraint("DAY(month_of_birth) = 1")
     )
-    # Sample.AffectedStatus
+    institution_id = db.Column(
+        db.Integer, db.ForeignKey("institution.institution_id", onupdate="cascade")
+    )
```
kevinlul, line 106: Is this meant to be nullable? I was wondering why the tests weren't failing despite no changes.

flask/utils/run.sh
helrick: Not sure if it's the best approach but I was getting sick of manually editing and it works!

flask/app/models.py
helrick: Hmm, good point. Currently it is nullable yes according to the metadata documentation. Since there's an 'Unknown' column though we _could_ make it be required

kevinlul (approved):
LGTM

PR #274, opened 2020-12-18
React: Hide admin pages and dataset delete

### LINKED ISSUE #216: Hide administrative pages and routes if the user is not an admin
Also hide dataset delete button

Keep in mind #214

SamarthVP:
Closes #216 
No admin page option if you're not logged in as an admin. Trying to access /admin route and other invalid routes redirects to /participants. The dataset delete button is hidden if not admin.

You can change the default admin's permissions by doing the following to the default database: 
```
docker exec -it st2020_mysql_1 /bin/bash
mysql -u $MYSQL_USER -p $MYSQL_DATABASE
# Enter MYSQL_PASSWORD
insert into users_groups values (1,1); insert into groups_datasets values (1,1);
# Toggle permissions
update user set is_admin=(not is_admin) where user_id=1;
```
Then refresh frontend

Take a good look, I probably missed something

### DIFF
--- a/react/src/App.tsx
+++ b/react/src/App.tsx
@@ -73,6 +73,7 @@ function BaseApp(props: { darkMode: boolean; toggleDarkMode: () => void }) {
                 setAuthenticated={setAuthenticated}
                 setLastLoginTime={setLastLoginTime}
                 setGlobalUsername={setUsername}
+                setIsAdmin={setIsAdmin}
             />
         );
     }
--- a/react/src/App.tsx
+++ b/react/src/App.tsx
@@ -35,7 +35,7 @@ function BaseApp(props: { darkMode: boolean; toggleDarkMode: () => void }) {
                 const loginInfo = await result.json();
                 setUsername(loginInfo.username);
                 setLastLoginTime(loginInfo.last_login);
-                setIsAdmin(loginInfo.is_admin)
+                setIsAdmin(loginInfo.is_admin);
             }
             setAuthenticated(result.ok);
         })();
--- a/react/src/App.tsx
+++ b/react/src/App.tsx
@@ -15,6 +15,7 @@ function BaseApp(props: { darkMode: boolean; toggleDarkMode: () => void }) {
     const [authenticated, setAuthenticated] = useState<boolean | null>(null);
     const [username, setUsername] = useState("");
     const [lastLoginTime, setLastLoginTime] = useState("");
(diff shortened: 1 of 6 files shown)


frewmack (approved):
Looks good to me!

react/src/App.tsx
```diff
@@ -34,6 +35,7 @@ function BaseApp(props: { darkMode: boolean; toggleDarkMode: () => void }) {
                 const loginInfo = await result.json();
                 setUsername(loginInfo.username);
                 setLastLoginTime(loginInfo.last_login);
+                setIsAdmin(loginInfo.is_admin);
```
kevinlul, line 38: This needs to be passed down to LoginForm too or the admin page will continue to be rendered/not rendered until a full refresh.

react/src/Navigation.tsx
```diff
@@ -286,21 +294,27 @@ export default function Navigation({
...
+                                            return <route.main username={username} />;
+                                        case "/datasets/:id?":
+                                            return <route.main isAdmin={isAdmin} />;
+                                        default:
+                                            return <route.main />;
+                                    }
+                                }}
+                            />
+                        ) : (
+                            <Redirect to={`/participants`} />
```
kevinlul, line 315:
```suggestion
                            <Redirect key={index} to={`/participants`} />
```

react/src/Navigation.tsx
```diff
@@ -264,14 +268,18 @@ export default function Navigation({
...
+                        {routes.map((route, index) =>
+                            !route.requiresAdmin || isAdmin ? (
+                                <ListItemRouterLink
+                                    key={index}
+                                    to={route.linkTo ? route.linkTo : "" + route.path}
+                                    primary={route.pageName}
+                                    children={route.icon}
+                                />
+                            ) : (
+                                <></>
```
kevinlul, line 280:
```suggestion
                                <React.Fragment key={index} />
```
This seems silly but it'll complain the console otherwise since it can't figure out what component the empty item originally was.

kevinlul (changes requested):

react/src/Navigation.tsx
SamarthVP: 🤔  That is silly

react/src/App.tsx
SamarthVP: Right, in case they re-sign in, good catch!

kevinlul (approved):
LGTM

PR #275, opened 2020-12-21
React: Hookup admin, settings pages

### LINKED ISSUE #257: Manage MinIO credentials in settings
View MinIO credentials and reset MinIO credentials on settings page. The interface should be similar to the administrative one, but it only affects the current user.

Depends on #193 to be actually functional

frewmack:
Followup to #268 
Admin page:
- MinIO keys are fetched when admin 'expands' element for a particular user
- 'Reset MinIO Credentials' resets the MinIO keys for that user
- Group membership is shown in user element as small chips with group code
- Group select hooked up

Settings page:
- MinIO keys are fetched here; same component as one in Admin page
- 'Reset MinIO Credentials' here too
- Permission groups that the user is a member of is displayed in the header to the right of the 'Hello {name}' title

Known issues / To-dos:
- Changing group membership of a user doesn't work, since the groups are not made via `POST /api/groups` and thus are missing their MinIO policies and buckets (thanks Kevin for helping me with that)
- I feel like group chips could use some colour to better distinguish them. They are all grey for now.

Closes #257


react/src/Admin/components/CreateUserModal.tsx
```diff
@@ -11,13 +11,48 @@ import {
...
+    switch (action.type) {
+        case "set":
+            const { type, ...payload } = action;
+            return { ...state, ...payload };
+        case "reset":
+            return initState;
+        default:
+            return state;
+    }
+}
```
kevinlul, line 48: Nice!

react/src/Admin/components/CreateUserModal.tsx
```diff
@@ -63,40 +98,31 @@ function ErrorText(props: ErrorTextProps) {
...
     const [submitting, setSubmitting] = useState(false);
     const [errorCode, setErrorCode] = useState(0);
     const [errorDetails, setErrorDetails] = useState({ error: "", message: "" });
 
+    // Reset on open/close as side effect
+    useEffect(() => {
+        dispatch({ type: "reset" });
+        setErrorCode(0);
+    }, [props.open]);
+
```
kevinlul, line 113: This is slightly different than only resetting on a successful submission but it works just as well!

kevinlul (commented):
Resetting MinIO credentials works really well! I don't know if it would be possible to switch both the access key and secret key to visible when you reset because that's probably what you want to see. I also noticed that the skeleton system obscuring the credentials reveals the size of the keys; at present the access key bar will appear half the width of the secret key bar.

Setting groups on user creation works well and switching seems to give positive feedback but I'm not exactly sure why the changes aren't persisting.

I'm okay to merge this and then keep working on it since the scope of this PR is pretty big already.

kevinlul:
Tip for the group situation:

Install the MinIO Client and configure the `local` host with the root credentials for your local instance.

Create this `policy.json`:
```json
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "s3:*"
      ],
      "Resource": [
        "arn:aws:s3:::bcch/*"
      ]
    }
  ]
}
```

```console
$ mc admin policy add local bcch policy.json
```

Rename the `bcch` in the policy JSON and repeat for each group (e.g. replace with `ach` and add a policy called `ach`).
Also do `mkdir -p minio/BUCKET` or `mc mb local BUCKET` for each but this doesn't affect the functionality in this PR.

Could be a to-do item for automatically setting up the development environment like `add-dummy-data`

frewmack:
> Resetting MinIO credentials works really well! I don't know if it would be possible to switch both the access key and secret key to visible when you reset because that's probably what you want to see. I also noticed that the skeleton system obscuring the credentials reveals the size of the keys; at present the access key bar will appear half the width of the secret key bar.

I can fix this quickly then call it ready for merge. I have a feeling the group persisting issue would take more time to figure out.

frewmack:
'Showing keys on key reset' can wait too. The way these components are setup makes it pretty awkward to setup.

kevinlul (approved):
🚀

PR #276, opened 2020-12-21
Flask: blueprints re-factor, extract helpers and separate extensions

### LINKED ISSUE #221: Blueprints refactor
- [ ] use blueprints
- [ ] remove as many circular imports as possible
- [ ] extract shared helpers out of routes.py

delvinso:
closes #221

- separating extensions from `__init__.py` into `extensions.py`
- as part of removing circular imports (imports from .extensions as opposed to import from app), logging and cli commands are also registered to the flask application through `__init__.py`
- blueprint refactor and registering
- extract shared helpers into `utils.py`

### DIFF
--- a/flask/app/__init__.py
+++ b/flask/app/__init__.py
@@ -25,10 +25,10 @@ def create_app(config):
     app = Flask(__name__)
     app.config.from_object(config)
 
+    config_logger(app)
     register_extensions(app)
     register_commands(app)
     register_blueprints(app)
-    config_logger(app)
 
     return app
 
--- a/flask/app/__init__.py
+++ b/flask/app/__init__.py
@@ -9,6 +9,19 @@ login = LoginManager()
 migrate = Migrate()
 
 
+from app import (
+    buckets,
+    routes,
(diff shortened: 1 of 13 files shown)


flask/app/__init__.py
```diff
@@ -16,32 +23,47 @@ def create_app(config):
...
     app.config.from_object(config)
 
-    if config.SQLALCHEMY_LOG:
-        logging.basicConfig()
-        logging.getLogger("sqlalchemy.engine").setLevel(logging.INFO)
-        logging.getLogger("sqlalchemy.pool").setLevel(logging.INFO)
+    register_extensions(app)
+    register_commands(app)
+    register_blueprints(app)
+    config_logger(app)
```
kevinlul, line 31: Logger should be registered first before the extensions are registered to match the order before.

kevinlul (approved):
Amazing!

flask/app/__init__.py
delvinso: fixed

PR #278, opened 2020-12-21
Flask: initialize MinIO policies for groups in dummy data

kevinlul:
Other:
- ignore .eslintcache
- Docker: prevent accidentally running prod images in dev and vice versa
- Compose: restart prod containers on failure, not dev


helrick (approved):
LGTM! Out of curiosity, were you able to simulate a failure to test the prod docker restart?

kevinlul:
You can try sending a signal to the main process in the container. Generally this is why the app dev container keeps restarting when it can't migrate the database so I trust Compose enough to do its job on this configuration.

PR #286, opened 2020-12-22
Flask: update DELETE /api/users to DELETE /api/users/:username

kevinlul:
#193

### DIFF
--- a/flask/tests/test_users.py
+++ b/flask/tests/test_users.py
@@ -273,3 +273,4 @@ def test_delete_user(test_database, minio_policy, client, login_as):
     error = json.loads(exc.value.args[0])
     assert error["error"]["message"] == "Unable to get user info"
     assert error["error"]["cause"]["error"]["Code"] == "XMinioAdminNoSuchUser"
+    assert User.query.filter_by(username="user").first() is None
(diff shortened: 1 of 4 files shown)


flask/tests/test_users.py
```diff
@@ -237,3 +241,35 @@ def test_create_conflicting_user(test_database, client, login_as):
...
+
+
+def test_delete_user_with_datasets(test_database, client, login_as):
+    login_as("admin")
+    assert client.delete(f"/api/users/admin").status_code == 422
+
+
+def test_delete_user(test_database, minio_policy, client, login_as):
+    minio_admin = minio_policy
+    minio_admin.add_user("user", "BatmanDiesInEndgame")
```
SamarthVP, line 263: :O

flask/tests/test_users.py
```diff
@@ -237,3 +241,35 @@ def test_create_conflicting_user(test_database, client, login_as):
...
+    user = User.query.filter_by(username="user").first()
+    user.minio_access_key = "user"
+    user.minio_secret_key = "BatmanDiesInEndgame"
+    db.session.commit()
+
+    login_as("admin")
+    assert client.delete(f"/api/users/user").status_code == 204
```
SamarthVP, line 270: Should probably check that the user is gone from the db as well

SamarthVP (changes requested):
Nice!

SamarthVP (approved):

PR #288, opened 2020-12-23
React: useEnums hook for retrieving cached /api/enums result

frewmack:
Related to #277, but doesn't close it as calls like `/api/groups` are still not cached.

I intend to push a more general solution later, but this works for now.

I also found an npm package called [react-query](https://react-query.tanstack.com/) that might be worth checking out in the future if we ever plan on refactoring how API calls are done on the front-end. (this PR doesn't use this BTW; all from scratch)


kevinlul (commented):
Do you anticipate that useContext logic will be complex enough to need to separate the hook and the provider in separate files? Is this a recommended practice?

frewmack:
> 
> 
> Do you anticipate that useContext logic will be complex enough to need to separate the hook and the provider in separate files? Is this a recommended practice?

I think it is recommended. For this particular case it seems like overkill, but I can definitely foresee the hook getting complex enough to warrant being in a separate file.

react/src/contexts/enums/index.tsx
```diff
@@ -0,0 +1,2 @@
+export { default as EnumProvider } from "./EnumProvider";
+export { default as useEnums } from "./useEnums";
```
kevinlul, line 2: If we want these two to be internal to the `./contexts/enums` module (i.e., never import directly through their individual files but through this file instead), we can drop the `default` from all three files and save some typing. This is optional as you already wrote the code and there's no functional difference.

kevinlul (approved):
🎁

react/src/contexts/enums/index.tsx
frewmack: Yeah I'd leave it as it is since I'll replace all this later anyways.

PR #289, opened 2020-12-23
Flask: add PATCH /api/users/:username

### LINKED ISSUE #188: Redesign administration page
User stories
- [x] As an administrator, I want to see last login times and activation status of all users.
- [x] As an administrator, I want to view and reset MinIO credentials for any user on demand.
- [x] As an administrator, I want to activate and deactivate users on demand.
- [x] As an administrator, I want to see what groups any user belongs to and add or remove them from those groups.

### LINKED ISSUE #193: Epic: /api/users

kevinlul:
Replaces PUT /api/users and POST /api/password
Hook up to frontend

Closes #193
Closes #188

### DIFF
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -99,50 +99,6 @@ def validate_user(request_user: dict):
     return False
 
 
-@routes.route("/api/users", methods=["PUT"])
-@login_required
-@check_admin
-def update_user():
-    rq_user = request.get_json()
-    if not validate_user(rq_user):
-        return "Bad request", 400
-
-    db_user = models.User.query.filter_by(username=rq_user["username"]).first_or_404()
-    if "password" in rq_user and len(rq_user["password"]):
-        db_user.set_password(rq_user["password"])
-    if "email" in rq_user:
-        db_user.email = rq_user["email"]
-
-    try:
-        db.session.commit()
-        return "Updated", 204
-    except:
-        db.session.rollback()
-        return "Server error", 500
-
-
-@routes.route("/api/password", methods=["POST"])
-@login_required
-def change_password():
-    params = request.get_json()
-    if "current" not in params or "password" not in params or "confirm" not in params:
-        return "Bad request", 400
-
-    if params["password"] != params["confirm"]:
-        return "Passwords do not match", 400
-
-    if not current_user.check_password(params["current"]):
-        return "Incorrect password", 401
-
-    current_user.set_password(params["password"])
-    try:
-        db.session.commit()
-        return "Updated", 204
-    except:
-        db.session.rollback()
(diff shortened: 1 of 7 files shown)


Madelinehazel:
I might be missing something, but I don't see a way to deactivate a user from the frontend; there's just an 'Active User' box that isn't responsive.

kevinlul:
Is your frontend up-to-date? I changed that in this PR too.

Madelinehazel:
I pulled your branch, so it should be up to date (and then `yarn` and `yarn start`); is there anything else i'd need to do?

Madelinehazel (approved):
LGTM

frewmack (changes requested):
Admin page looks good to me. However the endpoint on the Settings page needs to be updated as POST /api/password is no longer used.

kevinlul:
Good catch, there's actually a bit more on the frontend to change.

frewmack (approved):
Looks good to me!

PR #290, opened 2020-12-24
Flask: restrict user actions for analyses

### LINKED ISSUE #255: Restrict user actions for analyses
- [ ] Backend: users can only request analyses on datasets they have access to (`POST /api/analyses`)
- [ ] Backend and frontend: analysis state changes are restricted to administrators, except for cancelling analyses (`PATCH /api/analyses/:id`)

Spun off from #117

delvinso:
closes #255 

- [x] Backend: users can only request analyses on datasets they have access to (`POST /api/analyses`)
- [x] Backend and frontend: analysis state changes are restricted to administrators, except for cancelling analyses (`PATCH /api/analyses/:id`)

examples:
- need to enter mysql container and `insert into users_groups values (1,1); insert into groups_datasets values (1,1);` 
```
# user has access to dataset id 1
curl -X POST -H "Content-Type: application/json" -d '{"datasets" : [1], "pipeline_id" : 1}' localhost:5000/api/analyses\?user=1
# user doesn't have access to dataset id 2 
curl -X POST -H "Content-Type: application/json" -d '{"datasets" : [1, 2], "pipeline_id" : 1}' localhost:5000/api/analyses\?user=1
# no user specified? (can't create an analysis)
curl -X POST -H "Content-Type: application/json" -d '{"datasets" : [1, 2], "pipeline_id" : 1}' localhost:5000/api/analyses
```


flask/app/analyses.py
```diff
@@ -178,6 +178,38 @@ def create_analysis():
     except:  # LOGIN DISABLED
         requester_id = updated_by_id = 1
 
+    if app.config.get("LOGIN_DISABLED") or current_user.is_admin:
+        user_id = request.args.get("user")
+    else:
+        user_id = current_user.user_id
```
kevinlul, line 184: You should move this part up earlier so the preceding `requester_id` assignment can make use of this assumed user identity as well.

flask/app/analyses.py
```diff
@@ -178,6 +178,38 @@ def create_analysis():
...
+            models.Dataset.dataset_id
+            == models.groups_datasets_table.columns.dataset_id,
+        )
+        .join(
+            models.users_groups_table,
+            models.groups_datasets_table.columns.group_id
+            == models.users_groups_table.columns.group_id,
+        )
+        .filter(models.users_groups_table.columns.user_id == user_id)
+    )
```
kevinlul, line 206: This function is `create_analysis` so the Analysis doesn't exist yet! This means you want to query against Dataset rather than Analysis.

flask/app/analyses.py
```diff
@@ -178,6 +178,38 @@ def create_analysis():
...
+
+    permitted_ids = set([x for x, in query.values(models.Dataset.dataset_id)])
+
+    if not all([i in permitted_ids for i in dts_pks]):
+        return "Not all datasets are accessible to the current user", 400
```
kevinlul, line 211: The reasoning is okay here but it's not very efficient as it takes `O(datasets * requested datasets)` time. We've taken this approach for some `/api/users` and `/api/groups` to have the database server do the hard work instead:
https://github.com/ccmbioinfo/ST2020/blob/4a0595de65e9e043f569cb2a13c8f51d9ac42df9/flask/app/groups.py#L105-L109

flask/app/analyses.py
```diff
@@ -173,11 +174,37 @@ def create_analysis():
...
+    if app.config.get("LOGIN_DISABLED") or current_user.is_admin:
+        user_id = request.args.get("user")
+    else:
+        user_id = current_user.user_id
+
     try:
-        requester_id = updated_by_id = current_user.user_id
+        requester_id = updated_by_id = user_id
     except:  # LOGIN DISABLED
         requester_id = updated_by_id = 1
```
kevinlul, line 186: This should no longer be a try-except and just a simple assignment that falls back to `current_user.user_id`

flask/app/analyses.py
```diff
@@ -173,11 +174,37 @@ def create_analysis():
...
+            models.groups_datasets_table.columns.group_id
+            == models.users_groups_table.columns.group_id,
+        )
+        .filter(models.users_groups_table.columns.user_id == user_id)
+        .filter(models.Dataset.dataset_id.in_(dts_pks))
+        .all()
+    )
+
+    if len(permitted_ids) != len(dts_pks):
+        return "Not all datasets are accessible to the current user", 400
```
kevinlul, line 206: Should do a 404

flask/app/analyses.py
```diff
@@ -173,11 +174,37 @@ def create_analysis():
...
+            models.groups_datasets_table,
+            models.Dataset.dataset_id
+            == models.groups_datasets_table.columns.dataset_id,
+        )
+        .join(
+            models.users_groups_table,
+            models.groups_datasets_table.columns.group_id
+            == models.users_groups_table.columns.group_id,
+        )
+        .filter(models.users_groups_table.columns.user_id == user_id)
```
kevinlul, line 200: I should have clarified that this restriction shouldn't apply to admins

flask/app/analyses.py
```diff
@@ -244,6 +271,14 @@ def update_analysis(id: int):
         user_id = request.args.get("user")
     else:
         user_id = current_user.user_id
+        if request.json.get("analysis_state"):
+            if request.json.get("analysis_state") in [
```
kevinlul, line 275: Don't need a nested if. `None in []` and `None in ["Foo"]` are both False and don't raise exceptions.

kevinlul (changes requested):

flask/app/analyses.py
delvinso: this will now fail if there is no user_id supplied

flask/app/analyses.py
```diff
@@ -173,10 +174,34 @@ def create_analysis():
...
-        requester_id = updated_by_id = current_user.user_id
-    except:  # LOGIN DISABLED
-        requester_id = updated_by_id = 1
+
+    if app.config.get("LOGIN_DISABLED") or current_user.is_admin:
+        user_id = request.args.get("user")
+    else:
+        user_id = current_user.user_id
+
+    requester_id = updated_by_id = current_user.user_id
```
kevinlul, line 183: Okay, I think here we need a solution for the logic
- LOGIN_DISABLED -> must specify ?user= to populate this
- admin -> use ?user= and fallback to the admin's real ID
- regular -> real ID

flask/app/analyses.py
```diff
@@ -244,6 +269,13 @@ def update_analysis(id: int):
         user_id = request.args.get("user")
     else:
         user_id = current_user.user_id
+        if request.json.get("analysis_state") in [
+            "Requested",
+            "Running",
+            "Done",
+            "Error",
+        ]:
```
kevinlul, line 277: Might it be better to check if this is not Cancelled, since that's the only move authorized for regular users?

kevinlul (changes requested):

flask/app/analyses.py
delvinso: Something along the lines of this? I'm not sure I understand what the fallback value (real ID) is exactly, see the ??? below
```
    if app.config.get("LOGIN_DISABLED") or current_user.is_admin:
        user_id = requester_id = updated_by_id = request.args.get(
            "user", ???
        )
    else: # reg user
        user_id = requester_id = updated_by_id = current_user.user_id
```

kevinlul: Not quite as we currently depend on user_id being None to determine that the user is an admin and not restrict the query on datasets

kevinlul:
```python
if not user_id and app.config.get("LOGIN_DISABLED"):
    requester_id = updated_by_id = 1
else:
    requester_id = updated_by_id = user_id or current_user.user_id
```
I _think_ this solves the problem

delvinso: for my own understanding, since user_id is evaluated to determine if someone is an admin (for permission purposes) - what are the cases where it is assigned None? 
```
    if app.config.get("LOGIN_DISABLED") or current_user.is_admin:
        user_id = request.args.get("user") # assigned None here if no user query parameter is specified b/c of the get method 
    else:
        user_id = current_user.user_id # this will always return a value right because this would only be evaluated if the user is not an admin and therefore a regular user?
```
and then the check you have up there is basically accounting for that case where user_id is None, as a result of it not being specified as a query parameter. does this sound correct to you @kevinlul

kevinlul: None means either LOGIN_DISABLED or is_admin but didn't provide a `?user=` to masquerade as

kevinlul:
See if this works better?

delvinso:
> See if this works better?

Looks good to me, thanks for cleaning it up

kevinlul:
@Madelinehazel Since both Delvin and I ended up touching this, it would be best if you had a brief look

flask/app/analyses.py
```diff
@@ -152,68 +152,74 @@ def get_analysis(id: int):
...
-            "updated_by": obj.updated_by_id and obj.updated_by.username,
-        }
+    return (
+        jsonify(
+            {
+                **asdict(obj),
+                "assignee": obj.assignee_id and obj.assignee.username,
+                "requester": obj.requester_id and obj.requester.username,
+                "updated_by": obj.updated_by_id and obj.updated_by.username,
+            }
```
Madelinehazel, line 219: Would it make sense to return the dataset ids used to create the analysis?

flask/app/analyses.py
```diff
@@ -152,68 +152,74 @@ def get_analysis(id: int):
...
+    if app.config.get("LOGIN_DISABLED"):
+        user_id = request.args.get("user")
+        requester_id = updated_by_id = user_id or 1
+    elif current_user.is_admin:
+        user_id = request.args.get("user")
+        requester_id = updated_by_id = user_id or current_user.user_id
+    else:
+        requester_id = updated_by_id = user_id = current_user.user_id
+
+    if user_id:
```
Madelinehazel, line 176: Just noting that when I provide a non-existent user id with valid datasets (e.g. `curl -X POST -H "Content-Type: application/json" -d '{"datasets" : [7,8,9], "pipeline_id" : 2}' localhost:5000/api/analyses\?user=10`), I get the error `Some datasets were not found`. Not sure if this is an issue or not, as we would only be providing the user id when testing.

kevinlul: This isn't an issue because it's pretty much only for testing with a potential extension in the future to actually support assuming identites.

Madelinehazel (approved):
LGTM

PR #291, opened 2020-12-24
/api/_bulk: add group query string parameter

### LINKED ISSUE #258: /api/_bulk: add group query string parameter
e.g. `POST /api/_bulk?group=ach`, `POST /api/_bulk=group=cheo,sch`

Accepts a comma-delimited list of group codenames.
Admins can specify any groups that exist. Users can only specify groups they belong to or they get a 403.
The datasets are added to the specified permission groups.

Deny access to `POST /api/_bulk` if the user doesn't have any permission groups.
This query string parameter is optional if the user has only one permission group. In this case the permission group to use for the uploaded datasets is assumed to be that one.
This parameter is required if the user belongs to multiple permission groups, and if not provided, fail with 400.

Madelinehazel:
closes #258 

to test user permissions, need to add user in to users_groups table in mysql container

e.g

` INSERT INTO users_groups VALUES(1, 3);`

and test:

```
curl -X POST \
    --data '[{
            "family_codename": "1001",
            "participant_codename": "1411",
            "tissue_sample": "Blood",
            "dataset_type": "WGS",
            "condition": "GermLine",
            "sequencing_date": "2020-12-17"
        }]' \
    "localhost:3000/api/_bulk?group=ach&user=1"
```

### DIFF
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -244,10 +244,8 @@ def bulk_update():
         updated_by_id = 1
         created_by_id = 1
 
-    if app.config.get("LOGIN_DISABLED"):
-        user_id = request.args.get("user")
-    elif current_user.is_admin:
-        user_id = models.User.user_id
+    if app.config.get("LOGIN_DISABLED") or current_user.is_admin:
+        user_id = request.args.get("user", models.User.user_id)
     else:
         user_id = current_user.user_id
 
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -384,8 +384,15 @@ def bulk_update():
         else:
             files = row.get("linked_files", [])
         dataset.files += [models.DatasetFile(path=path) for path in files if path]
-        for g in user_group:
-            dataset.groups.append(g)
+
+        if group_query:
+            dataset.groups += [
+                models.Group.query.filter_by(group_code=g).first_or_404()
+                for g in group_query
+            ]
+        else:
+            dataset.groups.extend(user_group)
+
         db.session.add(dataset)
         transaction_or_abort(db.session.flush)
         dataset_ids.append(dataset.dataset_id)
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -244,50 +244,41 @@ def bulk_update():
         updated_by_id = 1
         created_by_id = 1
 
-    # get all group codes present in db
-    all_groups = models.Group.query.all()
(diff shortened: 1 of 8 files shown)


flask/app/routes.py
```diff
@@ -339,6 +384,8 @@ def bulk_update():
         else:
             files = row.get("linked_files", [])
         dataset.files += [models.DatasetFile(path=path) for path in files if path]
+        for g in user_group:
+            dataset.groups.append(g)
```
kevinlul, line 388: We need to use the groups specified in the query string if it's provided. This looks like it unconditionally adds the dataset to all of the user's permission groups.

kevinlul (changes requested):

flask/app/routes.py
Madelinehazel: ah good point, fixed now!

flask/app/routes.py
```diff
@@ -244,6 +244,51 @@ def bulk_update():
...
+        return "User does not belong to any permission groups", 403
+
+    if group_query:
+        group_query = group_query.split(",")
+        codes = [code.group_code for code in user_group]
+        for g in group_query:
+            if g not in all_codes:
+                return f"Group {g} does not exist", 404
+            if g not in codes:
+                return f"User does not belong to group {g}", 403
```
kevinlul, line 284: https://github.com/ccmbioinfo/stager/blob/master/flask/app/users.py#L182-L189

Can you adopt this approach to checking if the groups exist and the user is allowed access? This makes the database do the work. You'll need to adapt by doing your join and filter above but this should otherwise simply this a lot.

Madelinehazel: simplified in 60361f91d6c4c01f7d8a8fbd256ba82b058a0164

kevinlul (changes requested):

flask/app/routes.py
```diff
@@ -244,6 +244,42 @@ def bulk_update():
         updated_by_id = 1
         created_by_id = 1
 
+    if app.config.get("LOGIN_DISABLED"):
+        user_id = request.args.get("user")
+    elif current_user.is_admin:
+        user_id = models.User.user_id
```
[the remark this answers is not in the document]
Madelinehazel: Yes, rather than duplicating code I was looking for a way to filter against something that would not apply a condition, and found this!

kevinlul (approved):

PR #292, opened 2020-12-24
Creating default group/user + data hierarchies for testing

### LINKED ISSUE #208: Populate user_groups and groups_datasets entries in default database
The example datasets do not have groups, so only administrators can see them.

helrick:
Closes #208


flask/app/manage.py
```diff
@@ -264,191 +147,155 @@ def add_dummy_data():
...
+        # build trio
+        for sex in ["-", "Female", "Male"]:
+            participant = Participant(
+                participant_codename=f"{group.group_code.upper()}{participant_code_iter:04}",
+                institution_id=institution.institution_id,
+                affected=True if sex == "-" else False,
+                participant_type=ParticipantType.Proband
+                if sex == "-"
+                else ParticipantType.Parent,
+                sex=getattr(Sex, random.choice(["Female", "Male"]))
```
kevinlul, line 250: 🎲

kevinlul (approved):
Does the @with_appcontext decorator need to go on every function AND the click function?

flask/app/manage.py
delvinso: > Does the @with_appcontext decorator need to go on every function AND the click function?

I think the context needs to be given to only commands registered to the app (as in only the click functions, ie those which would be called through the command line)

PR #293, opened 2020-12-31
React: rename to Stager with app name in REACT_APP_NAME envvar

kevinlul:
Renaming the repository on merge


frewmack (approved):
Looks good to me.

PR #294, opened 2021-01-04
React: useFetchCache hook for caching any api call for session duration

### LINKED ISSUE #277: useContext and cache some common fetches like /api/enums to sessionStorage

frewmack:
Closes #277. 
API calls made through `useFetchCache` are only fetched once, and taken from cache for every subsequent call. This may cause issues later for calls like `/api/groups` which can be updated during a session, so we should revisit this later and use a library like react-query to save the hassle of setting up refetch policies from scratch.

### DIFF
--- a/react/src/contexts/fetchCache/FetchCacheProvider.tsx
+++ b/react/src/contexts/fetchCache/FetchCacheProvider.tsx
@@ -6,21 +6,32 @@ import { ClientContext, ResultContext } from "./FetchCacheContext";
  */
 export function FetchCacheProvider(props: { children: React.ReactNode }) {
     const [url, setURL] = useState("");
-    const [cachedResult, setCachedResult] = useState<any>();
+    const [cache, setCache] = useState<{ [url: string]: any }>({});
 
+    // Problem: If you ask for resource B when you already asked for resource A,
+    // the hook gives resource A back for a split-second while resource B gets
+    // fetched.
     useEffect(() => {
+        function updateCache(key: string, value: any) {
+            setCache(oldCache => {
+                const newCache = { ...oldCache };
+                newCache[key] = value;
+                return newCache;
+            });
+        }
+
         // use url as cache key
         const stored = sessionStorage.getItem(url);
         if (stored) {
             // hit, no fetch
-            setCachedResult(JSON.parse(stored));
+            updateCache(url, JSON.parse(stored));
         } else {
             // miss, refetch
             fetch(url).then(async response => {
                 if (response.ok) {
                     const data = await response.json();
                     sessionStorage.setItem(url, JSON.stringify(data));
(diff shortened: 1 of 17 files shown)


react/src/contexts/fetchCache/FetchCacheProvider.tsx
```diff
@@ -0,0 +1,51 @@
+import React, { useEffect, useState } from "react";
+import { ClientContext, ResultContext } from "./FetchCacheContext";
+
+/**
+ * Provider for useFetchCache hook.
+ */
+export function FetchCacheProvider(props: { children: React.ReactNode }) {
+    const [url, setURL] = useState("");
```
kevinlul, line 8: Naively thinking, shouldn't url not be a state and just be a prop? Isn't this the cause of the problem you mentioned or is that not how contexts work?

frewmack: I made the url a state variable because I knew that calling `setURL` would trigger a re-render and cause the `useEffect` code to run. 

The issue I had initially was that I stored only one cached result in a state variable as opposed to storing the entire cache. Storing one cached result caused an issue where if you ask for resource A, and then ask for resource B, the provider would give you resource A until it finished fetching B, causing undesirable behaviour. Now when you ask for A, it will return cache[A], which may be undefined indicating that it is not fetched yet.

kevinlul: Ah, so the comment below is justifying the `useEffect`, not saying that the problem exists now?

frewmack: Yep. The one hook can handle multiple urls now.

react/src/contexts/fetchCache/useFetchCache.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+/**
+ * Hook that returns the JSON response from
+ * GET-ing the provided url. Result is cached and
+ * reused. Cache expires on session end.
+ */
+export function useFetchCache(url: string) {
+    useContext(ClientContext)(url);
+    const result = useContext(ResultContext);
+    return result ? (result[url] ? result[url] : undefined) : undefined;
```
kevinlul, line 12:
```suggestion
    return result && result[url];
```

kevinlul (changes requested):
Seems to be working!

kevinlul:
I agree that we will have to look at a library for this on the frontend. In addition to the problem you mentioned with this cache being session-permanent, we also have a setState while rendering due to the implementation:
```
Warning: Cannot update a component (`FetchCacheProvider`) while rendering a different component (`ParticipantTable`).
```

kevinlul (approved):

PR #295, opened 2021-01-05
React: replace manage uploaders tab with group management page

### LINKED ISSUE #214: Replace manage uploaders tab with group management page
As an administrator, I want to see what permission groups exist in the system, add or remove groups, or change the display name of these groups.

The manage uploaders tab shouldn't exist and how it's implemented by wrapping the child in a Typography results in a lot of DOM nesting errors as not everything can be nested under a `<p>`

annnag:
Closes #214


react/src/Groups/components/CreateGroupModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+                    helperText={group.group_code === "" ? "This field cannot be empty." : ""}
+                />
+            </DialogContent>
+            <DialogActions>
+                <Button color="default" variant="outlined" onClick={onClose}>
+                    Cancel
+                </Button>
+                <Button type="submit" color="primary" variant="contained" onClick={submit}>
+                    Create
+                </Button>
```
kevinlul, line 82: Can you disable the submit button while a form submit is in progress like the create user process to avoid glitches when spam-clicking?

react/src/Navigation.tsx
```diff
@@ -159,6 +161,12 @@ const routes: RouteItem[] = [
         main: Analyses,
         icon: <ShowChartIcon />,
     },
+    {
+        pageName: "Groups",
+        path: "/groups",
+        main: Groups,
+        icon: <SupervisedUserCircle />,
+    },
```
kevinlul, line 169: Can you move this after the admin page and mark it admin-only?

kevinlul (changes requested):

react/src/Groups/components/CreateGroupModal.tsx
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    async function submit(e: React.FormEvent) {
+        e.preventDefault();
+
+        const response = await fetch("/api/groups", {
+            method: "POST",
+            credentials: "same-origin",
+            headers: { "Content-Type": "application/json" },
+            body: JSON.stringify(group),
```
frewmack, line 35: I don't know if group codes are case-sensitive as far as MinIO is concerned, but since the codes are converted to upper-case in the group boxes, I think they should be made lower-case before sending.

react/src/Groups/index.tsx
```diff
@@ -0,0 +1,122 @@
...
+                                        { variant: "success" }
+                                    );
+                                } else {
+                                    enqueueSnackbar(
+                                        `Failed to edit group ${group.group_name}. Error: ${response.status} - ${response.statusText}`,
+                                        { variant: "error" }
+                                    );
+                                }
+                            }}
+                            onDelete={async () => {
```
frewmack, line 84: Clicking "delete" on the group box should open a ConfirmModal to make sure that the user definitely means to delete a group.

frewmack (changes requested):
Looks good. Some minor UX changes needed.

kevinlul (approved):
I believe that's everything!

PR #296, opened 2021-01-06
Flask: fix empty group deletion returning 422 on success

kevinlul:
I don't think the frontend is handling the 204 even though it receives the response when I look in the Network tab...


kevinlul:
@annnag Can you confirm that this fixes the problem for you and I can merge this to your PR?

annnag (approved):
Yes, this fixes the problem!

PR #298, opened 2021-01-06
React: Fix material-table filters being deleted upon first typing

frewmack:
Quick fix for this weird behaviour. 

There is still a bug where the page re-renders when you open/close the Drawer, or when you click the dark mode switch on the AppBar. My guess is it's related to how the routes are rendered or something, but I haven't figured it out.


kevinlul (approved):
Was it this simple? It was exclusive to those two pages?

frewmack:
Yep. I noticed the problem didn't happen in the Participants page, which didn't save the param ID in a state variable.

PR #301, opened 2021-01-07
docs: separate from README and expand

kevinlul:
Feedback very much wanted due to the curse of knowledge


kevinlul:
This is easiest to read without checking out locally if you look at the branch on GitHub:

https://github.com/ccmbioinfo/stager/tree/kevinlul/devdocs

https://github.com/ccmbioinfo/stager/tree/kevinlul/devdocs/docs

Note that all of the internal links won't work until merged because they assume the files are already on `master`.

frewmack (approved):
Front-end and Docker look good to me. Can't think of anything else to add.

docs/react.md
```diff
@@ -0,0 +1,31 @@
...
+1. Switch to the `react` directory.
+1. Create a `.env` with these parameters that you can change as needed.
+
+   ```
+   REACT_APP_NAME=Stager
+   REACT_APP_MINIO_URL=http://localhost:9000/
+   ```
+
+1. Install dependencies with `yarn`.
+1. Start a development server in watch mode with `yarn start`.
```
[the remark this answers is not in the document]
kevinlul: This is actually how ordered lists work in [markdown](https://guides.github.com/features/mastering-markdown/) and I think this is to facilitate inserting items in between (or the fact that it ends up rendering as <li> in HTML so the number is of no consequence). I do agree that source looks weird when starting out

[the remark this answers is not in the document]
frewmack: Oh interesting. That actually makes a lot of sense.

docs/react.md
```diff
@@ -0,0 +1,31 @@
...
+```bash
+docker-compose up -d --build
+```
+
+1. Switch to the `react` directory.
+1. Create a `.env` with these parameters that you can change as needed.
+
+   ```
+   REACT_APP_NAME=Stager
+   REACT_APP_MINIO_URL=http://localhost:9000/
```
annnag, line 25: Maybe also include `REACT_APP_DEFAULT_OPTIONAL_COLUMNS` here?

annnag (commented):
Looking good!

docs/react.md
kevinlul: Good idea!

annnag (approved):
👍

docs/database.md
```diff
@@ -0,0 +1,118 @@
...
+
+Apply the migration and commit your changes to Git.
+
+```bash
+flask db upgrade
+```
+
+Because `flask db upgrade` is automatically called by the startup script [`./utils/run.sh`](https://github.com/ccmbioinfo/stager/blob/master/flask/utils/run.sh),
+you may experience problems when switching between branches that expect different database versions.
+Generally we just reset the database at this point, but you could avoid this by running
```
delvinso, line 81: by reset does this mean deleting ie.`rm -rf`-ing the mysql folder?

docs/flask.md
```diff
@@ -0,0 +1,202 @@
...
+
+To add a development dependency, add the line to `requirements-dev.in` instead.
+
+```bash
+pip-compile requirements-dev.in
+pip install -r requirements-dev.in
+```
+
+Commit the changes to Git. If you are using Docker, you should rebuild the image
+for changes to persist.
```
delvinso, line 82: I think it would be helpful to add the command (`docker-compose up --build` )for rebuilding the docker image here since this has come up a few times in slack after a new python dependency is introduced

docs/database.md
kevinlul: That is usually how we do it and that's described in the Docker documentation under "Resetting bind mounts". It doesn't generally apply if you're running MySQL outside of Docker, where it might be easier to drop all the tables in the database. Should I clarify on both?

docs/flask.md
```diff
@@ -0,0 +1,202 @@
...
+
+## Application architecture
+
+The backend is a RESTful API only. It is not responsible for serving the frontend or HTTPS.
+The main entrypoint is [`wsgi.py`](https://github.com/ccmbioinfo/stager/blob/master/flask/wsgi.py),
+which instructs Flask (or any WSGI server in [production](https://github.com/ccmbioinfo/stager/blob/master/docs/production.md))
+to create an app instance with the default config and start the server to listen for HTTP requests.
+
+Configuration is mostly read from environment variables in [`config.py`](https://github.com/ccmbioinfo/stager/blob/master/flask/app/config.py).
+Because [Flask-Login](https://flask-login.readthedocs.io/) is authenticates via a session cookie by default,
```
delvinso, line 92: small typo on authenticates

docs/flask.md
kevinlul: Yeah, I'll do that instead of assuming a reference to the Docker part of the documentation that I wrote first.

docs/flask.md
```diff
@@ -0,0 +1,202 @@
...
+Each subroute is loaded via a Flask [Blueprint](https://flask.palletsprojects.com/en/1.1.x/tutorial/views/).
+This also registers the SQLAlchemy ORM and Flask-Migrate. More on this in the
+[database docs](https://github.com/ccmbioinfo/stager/blob/master/docs/database.md).
+
+The `add-default-admin` and `add-default-data` commands are registered to aid in development.
+They are automatically run by the startup shell script. Most of the remaining files each
+hold a Blueprint and implement a set of endpoints for that subroute.
+
+## Tips
+
```
delvinso, line 108: would it be worth the time to add a small section on how to add new click (command line) commands and extensions or are these set in stone at this point of development?

kevinlul: The commands aren't set in stone but I think the best I can do is link to relevant documentation instead of reproducing it here.

docs/database.md
delvinso: that makes sense, I'm leaning towards clarifying on both

docs/flask.md
```diff
@@ -0,0 +1,202 @@
...
+docker-compose -p stager-test -f docker-compose.test.yaml run --rm -T app
+```
+
+`--rm` removes the test container after it's done running.
+`-T` is optional. It does not connect a TTY, which used to cause problems with a
+particular test but has since been [resolved](https://github.com/minio/mc/issues/3499).
+
+To run a specific test file or function instead of the entire test suite, append `pytest ARGUMENTS`
+to the command above. For example, `pytest tests/test_users.py` will only run the integration tests
+in the specified file. For more information, see [pytest usage](https://docs.pytest.org/en/stable/usage.html).
```
delvinso, line 188: I was actually wondering if there was a way to do this since I've been waiting for all tests to run when adding or testing new tests!

delvinso (approved):
looks great! super helpful to have all this information in a single place

PR #299, opened 2021-01-06
React: hook up linking multiple files and make sequencing_date a required field

annnag:
Closes #281

PR #302, opened 2021-01-07
React: Permission group multiselect in Add Datasets

### LINKED ISSUE #259: Add datasets: add multiselect for permission group if user has multiple
Depends on #258

frewmack:
Closes #259. Also changes /api/login to send the user's permission groups on login.


react/src/AddDatasets/components/DataEntryTable.tsx
```diff
@@ -377,6 +387,12 @@ function DataEntryToolbar(props: {
                 <Box display="flex" flexGrow={1}>
                     <Typography variant="h6">Enter Metadata</Typography>
                 </Box>
+                <GroupDropdownSelect
+                    selectedGroupCodes={props.groups}
+                    allGroupCodes={props.allGroups}
+                    onChange={props.setGroups}
+                    disabled={props.allGroups.length <= 1}
```
kevinlul, line 394: When I first sign in I always get an error on this line that `allGroups` is undefined when navigating to add datasets. This is fixed by a hard refresh.

kevinlul (changes requested):

react/src/AddDatasets/index.tsx
```diff
@@ -28,21 +28,33 @@ const useStyles = makeStyles(theme => ({
...
     useEffect(() => {
         document.title = `Add Datasets | ${process.env.REACT_APP_NAME}`;
         handleDataChange(createEmptyRows(1)); // sets errorMessage on initial render
     }, []);
 
+    useEffect(() => {
+        if (props.groups.length > 0) {
+            setAsGroups(props.groups);
+        } else {
+            setErrorMessage("Cannot submit. You are not part of any permission groups.");
```
kevinlul, line 51: Can a similar message be added for when the user attempts to submit with the field not filled out?

kevinlul (changes requested):
This works great and I like that the selection is disabled if the user cannot make a meaningful choice.

react/src/AddDatasets/index.tsx
```diff
@@ -52,6 +52,13 @@ export default function AddParticipants(props: {
         }
     }, [props.groups]);
 
+    function onChangeGroups(newGroups: string[]) {
+        setAsGroups(newGroups);
+        if (newGroups.length === 0) {
+            setErrorMessage("Cannot submit. You must select a permission group.");
+        }
```
kevinlul, line 59: I think the error condition needs to consider both the filled-out state of the entire form and the dropdown. This worked initially, but when I filled out the table, the submit button was enabled again.

kevinlul (changes requested):

kevinlul (approved):
LGTM!

annnag (changes requested):
Looks good to me! Everything works as expected.
One super minor problem: when one dataset is added, the snack bar says `1 datasets...` instead of `1 dataset...`

annnag (approved):

kevinlul:
We have a merge conflict because #299 was merged

PR #303, opened 2021-01-12
Flask: add limit, page, order_by, order_dir, filter parameters to /api/participants

### LINKED ISSUE #283: Epic: reducing data overfetch to allow for progressive loading
We'll explicitly program filters that correspond to those on the frontend instead of trying for a generic solution due to the complex nature of the problem. The only parameters that should be generic are the `page`, `limit`, and ordering parameters.

- [x] /api/participants
  - [x] `participant_codename`, `notes`: LIKE
  - [x] `participant_type`, `sex`: IN list
  - [x] `affected`, `solved`: exact match (== true/false/null)
  - [x] `family_codename` LIKE (needs join)
  - [x] ~~`dataset_type`: IN list (needs join)~~ cancelled
- [x] /api/datasets
  - [x] `family_codename`, `participant_codename`, `updated_by`: LIKE (needs join)
  - [x] `tissue_sample_type`: IN list (needs join)
  - [x] `dataset_type`, `condition`: IN list
  - [x] `notes`: LIKE
  - [x] `files`: LIKE (needs join or querying on a different table); querying by the number of files is optional
  - [x] `updated`: date range (leq/geq)
- [x] /api/analyses
  - [x] `assignee`, `requester`: LIKE (needs join)
  - [x] `updated`: date range (leq/geq)
  - [x] `path_prefix` ,`notes`: LIKE
  - [x] `status`: IN list
  - [x] ~`pipeline`: skip~ 

## Original
**Overfetch: add limit, order, query parameters to all list endpoints**

We want to add parameters that correspond to LIMIT BY (reducing rows returned) and ORDER BY (so the order is predictable), along with a parameter that specifies the index (which page to return) to reduce slow loading as a result of massive overfetch.

We want to add query parameters that correspond to WHERE clauses with LIKE conditions so that the frontend filters can be hooked into these under progressive loading. Thus we need to specify what column we are querying on and what value specifically, exact match, or range.

Optionally, we can add a parameter that specifies exactly which columns to return in this request to further reduce overfetch.

Madelinehazel:
#283 

For now, I've just modified the participants endpoint so we can flesh it out before applying to families, analyses, etc

Do we still want the 'starts_with' query to filter by participant_codename? This was something we had implemented previously.


kevinlul:
filter supersedes start_with, and we can merge separate PRs for each affected endpoint to keep PRs smaller!

flask/app/participants.py
```diff
@@ -37,22 +37,36 @@
...
     # parsing query parameters
-    max_rows = request.args.get("max_rows", default=100)
+    limit = request.args.get("limit", default=10)
+    page = request.args.get("page", default=1)
     starts_with = request.args.get("starts_with", default="", type=str)
     starts_with = f"{starts_with}%"  # sql syntax
     order_by_col = request.args.get("order", default="participant_id", type=str)
-    # need some default or we need an ifelse statement, one with order_by method and one without. AFAIK there is no 'default' parameter we can pass into order_by to get the default sql ordering scheme
+    filter_by = request.args.get("filter", default="participant_id,", type=str)
+    filter_by_col, filter_val = filter_by.split(",")[0], filter_by.split(",")[1]
```
kevinlul, line 46: You'll get an IndexError here if there's no comma. Is this meant to be a basis to support multiple filters?

flask/app/participants.py
```diff
@@ -37,22 +37,36 @@
...
     columns = models.Participant.__table__.columns.keys()
 
     if order_by_col not in columns:
         return f"Column name for ordering must be one of {columns}", 400
-    column = getattr(models.Participant, order_by_col)
+    if filter_by_col not in columns:
+        return f"Column name for filtering must be one of {columns}", 400
+
+    order_column = getattr(models.Participant, order_by_col)
+    filter_column = getattr(models.Participant, filter_by_col)
```
kevinlul, line 69: This can be made more Pythonic by doing a try-except on getattr and catching the [AttributeError](https://docs.python.org/3/library/functions.html#getattr)

flask/app/participants.py
Madelinehazel: No, just to combine the query parameters for the filter column and filter value. Instead, I can separate these into two query parameters, e.g. filter_column and filter_value.

flask/app/participants.py
```diff
@@ -37,22 +37,36 @@
...
-        int(max_rows)
+        int(limit)
     except:
-        return "Max rows must be a valid integer", 400
+        return "Limit must be a valid integer", 400
+
+    try:
+        int(page)
+    except:
+        return "Page must be a valid integer", 400
```
kevinlul, line 57: If the `type` parameter works as it does above with `order_by_col` and `filter_by` then I think we can do `type=int` for the numeric parameters and wrap the `.get` calls in a try-except that catches the [ValueError](https://werkzeug.palletsprojects.com/en/1.0.x/datastructures/#werkzeug.datastructures.MultiDict)

flask/app/participants.py
Madelinehazel: fixed in 920f9b2c23911b11e012ee51312d3751d1142d33

flask/app/participants.py
Madelinehazel: fixed in 920f9b2c23911b11e012ee51312d3751d1142d33

flask/app/participants.py
Madelinehazel: This was something I noticed when we were implementing a limit last time-  when type=int is specified and a string is passed instead of int, no error is raised.

kevinlul: I see, for some reason it returns None instead of raising a ValueError...

flask/app/participants.py
```diff
@@ -37,22 +37,33 @@
...
-    starts_with = request.args.get("starts_with", default="", type=str)
-    starts_with = f"{starts_with}%"  # sql syntax
-    order_by_col = request.args.get("order", default="participant_id", type=str)
-    # need some default or we need an ifelse statement, one with order_by method and one without. AFAIK there is no 'default' parameter we can pass into order_by to get the default sql ordering scheme
+    limit = request.args.get("limit", default=10)
+    page = request.args.get("page", default=1)
+    order_by_col = request.args.get("order_by_by", default="participant_id", type=str)
+    filter_by_col = request.args.get("filter_by", default="participant_id", type=str)
+    filter_val = request.args.get("filter_val", default="", type=str)
+    filter_val = f"%{filter_val}%"
```
kevinlul, line 45: This works but to support all the filters in the frontend tables, we need to support multiple filters. i.e. support any number of fields of the form `column_name=value` as long as it refers to a valid column name that can be read. We may need to change the form so non-strings work too and that might be a bit complicated to do generically vs just implementing specific cases for those columns here.

flask/app/participants.py
```diff
@@ -81,8 +94,9 @@ def list_participants():
                 == models.users_groups_table.columns.group_id,
             )
             .filter(models.users_groups_table.columns.user_id == user_id)
-            .order_by(column)
-            .limit(max_rows)
+            .order_by(order_column)
```
frewmack, line 97: We should also include order direction as a parameter since it's supported by material-table (order_direction="asc" or "desc").
```
from sqlalchemy import asc, desc

query(...)
.order_by(asc(order_column))  # or desc(order_column)
```

Madelinehazel: noted, i'll look into this!

Madelinehazel:
Revised code to allow for dynamic filtering in 15a76836a62698d2a0a63d16157c07eb1e32c752
example: 
```
curl http://localhost:3000/api/participants?filter=participant_id;ge;9&user=3
curl "http://localhost:3000/api/participants?filter=participant_codename;like;%B%"
```
Note that the query should be in the form filter=column;operation;value, where operation corresponds to the  SQLAlchemy column comparators (e.g. `eq` for `==`, `like` for `like`, `ge` for` >=`)

This will likely need some refinement for validating that the query value is appropriate to the operation.

kevinlul:
How are multiple filters handled in this form?

Madelinehazel:
> How are multiple filters handled in this form?

Ah, when I was reconfiguring the function to return a filter rather than a query I didn't take this into account; fixed in b21d083eed92d31a4283af4dcbdb852bec0d89c7

flask/app/participants.py
Madelinehazel: added in 4882bdcc9ea81fc1a227bc99568e6c7540c38041

flask/app/participants.py
```diff
@@ -37,38 +37,65 @@
...
+    offset = (int(page) * int(limit)) - int(limit)
 
     columns = models.Participant.__table__.columns.keys()
 
-    if order_by_col not in columns:
-        return f"Column name for ordering must be one of {columns}", 400
-    column = getattr(models.Participant, order_by_col)
+    raw_filters = request.args.getlist("filter")
+    if raw_filters:
+        filt = filter_query(models.Participant, raw_filters)
```
frewmack, line 62: Is it possible to pass in the joined tables as the model here? We should be able to filter by any column that gets sent out at the end (eg. family_codename, dataset_types)

frewmack (changes requested):

flask/app/participants.py
```diff
@@ -81,23 +108,26 @@ def list_participants():
...
+            ),
+            joinedload(models.Participant.created_by),
+            joinedload(models.Participant.updated_by),
         )
+        for f in filt:
+            participants = participants.filter(f)
+        participants = participants.order_by(order()).limit(limit).offset(offset)
+
+    if 400 in participants:
+        return f"{participants}", 400
```
kevinlul, line 130: What does this do?

kevinlul (commented):
I'm kind of curious how you're able to iterate over a Query object rather than the results. It seems to work so I'm inclined to merge and try this out and refine it as we go. I don't know if the default limit/page make sense and this technically breaks behaviour, although I don't know if we should allow requesting the entire collection anymore anyway.

flask/app/participants.py
Madelinehazel: oops, this is no longer needed. removed in 57d248be0f491e0d8ad31f75add0169ad14b86d0

kevinlul (approved):

frewmack (approved):
We can merge this now and handle filtering by `family_codename` and `dataset_types` in a later PR 🚀

PR #305, opened 2021-01-15
React: Fix empty string check in Add Datasets

frewmack:
I was not able to reproduce the bug, but seeing as calling `.trim(...)` requires the string to be defined, I replaced it with a helper function that checks if undefined and tests the string using `RegEx.test`.


kevinlul (changes requested):
It seems like now the participant type field that should be prefilled by selecting an existing participant just gets disabled, so you can't submit

frewmack:
> It seems like now the participant type field that should be prefilled by selecting an existing participant just gets disabled, so you can't submit

I think this was expected behaviour. A participant already exists => disable the fields, and the values are invalid => cannot submit.

I've changed it so that the fields should only be disabled if the required fields that are filled out by the pre-existing participant are incorrect. This is going to have some weird implications for submission though.

kevinlul (approved):
LGTM

PR #319, opened 2021-01-19
React: Fix permission group dropdown defaults in Add Datasets

### LINKED ISSUE #310: Uncheck all permission groups in add datasets by default to prevent accidents

frewmack:
Closes #310 

- Users in only one group will be automatically filled in
- Users in multiple groups will start with no group selected to prevent mistakes


kevinlul (approved):
Nice!

PR #321, opened 2021-01-20
Flask: add GET /api/metadatasettypes

### LINKED ISSUE #317: GET /api/metadatasettypes
Returns an object where the keys are the names of meta-dataset types and the values are lists of corresponding dataset types.

Meant to replace the workaround in /api/enums

delvinso:
closes #317
curl localhost:5000/api/metadatasettypes 

```
{
  "Exome": [
    "CES", 
    "CPS", 
    "RCS", 
    "RDC", 
    "RDE", 
    "RES", 
    "WES"
  ], 
  "Genome": [
    "CGS", 
    "RGS", 
    "WGS"
  ], 
  "Other": [
    "RLM", 
    "RMM", 
    "RTA"
  ], 
  "RNA": [
    "RRS"
  ]
}
```

### DIFF
--- a/flask/tests/conftest.py
+++ b/flask/tests/conftest.py
@@ -60,6 +60,48 @@ def client(application):
 
 @pytest.fixture
 def test_database(client):
+
+    dataset_types = [
+        "RES",
+        "CES",
+        "WES",
+        "CPS",
+        "RCS",
+        "RDC",
+        "RDE",
+        "RGS",
+        "CGS",
+        "WGS",
+        "RRS",
+        "RLM",
+        "RMM",
+        "RTA",
+    ]
+
+    for d in dataset_types:
+        db.session.add(DatasetType(dataset_type=d))
+    db.session.flush()
+
+    metadataset_types = ["Genome", "Exome", "RNA", "Other"]
+
+    for m in metadataset_types:
+        db.session.add(MetaDatasetType(metadataset_type=m))
+    db.session.flush()
+
+    md_d = {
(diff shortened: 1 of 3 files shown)


flask/tests/conftest.py
```diff
@@ -97,9 +139,6 @@ def test_database(client):
 
     wes = DatasetType(dataset_type="WES")
     wgs = DatasetType(dataset_type="WGS")
-    db.session.add(wes)
-    db.session.add(wgs)
-    db.session.flush()
```
kevinlul, line 102: Since you're dropping the two hardcoded cases in favour of the entire array above, you should remove the two variables above too and refer to the literal string where they are used below.

flask/tests/test_misc.py
```diff
@@ -27,6 +27,16 @@ def test_get_enums(test_database, client, login_as):
...
 
 
+# GET /api/metadatasettypes
+def test_get_metadatasettypes(test_database, client, login_as):
+    login_as("admin")
+    response = client.get("/api/metadatasettypes")
+    assert response.status_code == 200
+    assert len(response.get_json()) == 4
+    for metadataset_type, _ in response.get_json().items():
+        assert metadataset_type is not None
```
kevinlul, line 37: JSON keys cannot be null so this check doesn't do anything. It would be better to assert that the values aren't null and are arrays.

kevinlul (changes requested):

kevinlul (approved):
LGTM!

PR #335, opened 2021-01-21
React: Replace fetchCache context with react-query hooks

frewmack:
Adds a new `react-query` dependency for caching API calls. I plan to setup fetching like this for most other API calls, but this is just a starter to replace the old fetchCache functions.


kevinlul (approved):
LGTM 🎸

PR #338, opened 2021-01-22
React: Refactor /api/groups hookup to use react-query

frewmack:
These custom hooks each correspond to the 'groups' endpoints, and help for managing fetching behaviour in one place.

### DIFF
--- a/react/src/Groups/index.tsx
+++ b/react/src/Groups/index.tsx
@@ -4,7 +4,7 @@ import { GroupAdd } from "@material-ui/icons";
 import { useSnackbar } from "notistack";
 import { Group as GroupCard } from "./components/Group";
 import CreateGroupModal from "./components/CreateGroupModal";
-import { Group } from "../typings";
+import { useGroups, useGroupsPatch, useGroupsDelete } from "../hooks";
 
 const useStyles = makeStyles((theme: Theme) => ({
     root: {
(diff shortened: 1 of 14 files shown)


kevinlul (commented):
Is naming the hook after the HTTP method the best or a verb for what the action actually is (create, update)?

frewmack:
> Is naming the hook after the HTTP method the best or a verb for what the action actually is (create, update)?

I did have an idea to include an 'alias' hook called `useGroupsEndpoint` that provides `.post(), .patch(), .delete()` which just call the `.mutate()` functions for their respective mutations.

Other than that, I don't know what the standard is. Even using a verb is inaccurate since it returns a mutation object that you then use to 'mutate' the query data by interacting with the backend.

kevinlul:
Okay, I guess as-is is fine.

react/src/Groups/index.tsx
```diff
@@ -31,17 +31,16 @@ const useStyles = makeStyles((theme: Theme) => ({
 
 export default function Groups() {
     const classes = useStyles();
-    const [groups, setGroups] = useState<Group[]>([]);
-    const [openNewGroup, setOpenNewGroup] = useState<boolean>(false);
+    const groups = useGroups();
+    const groupPatch = useGroupsPatch();
+    const groupDelete = useGroupsDelete();
+    const [openNewGroup, setOpenNewGroup] = useState(false);
```
kevinlul, line 37: It seems like on the rerender now when we add or remove a group, `/api/groups` and each group are re-fetched now and I guess this makes sense due to cache invalidation but I don't think this was the behaviour before?

frewmack: Yeah this wasn't the behaviour before. I'm looking into it.

kevinlul (approved):
LGTM!

PR #346, opened 2021-01-26
Add institution field to APIs returning participants 

### LINKED ISSUE #311: Add institution field to APIs returning participants
The field should contain the readable name after the join to the Institution table and not the primary key. See how the Analysis-User relationship and the pipeline supported type relationships are used.

Madelinehazel:
closes #311
 Not sure what 'See how the Analysis-User relationship and the pipeline supported type relationships are used.' in the issue description is referring to.

### DIFF
--- a/flask/app/participants.py
+++ b/flask/app/participants.py
@@ -246,7 +246,7 @@ def create_participant():
         else:
             institution_obj = models.Institution(institution=institution)
             db.session.add(institution_obj)
-            transaction_or_abort(db.session.commit)
+            transaction_or_abort(db.session.flush)
             institution_id = institution_obj.institution_id
 
     ptp_objs = models.Participant(
--- a/flask/app/participants.py
+++ b/flask/app/participants.py
@@ -235,6 +235,20 @@ def create_participant():
     if enum_error:
         return enum_error, 400
 
+    # get institution id
+    institution = request.json.get("institution")
+    if institution:
+        institution_obj = models.Institution.query.filter(
+            models.Institution.institution == institution
+        ).one_or_none()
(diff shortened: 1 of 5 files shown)


kevinlul:
I was referring to how a custom property could be used to retrieve the desired value across a relationship

https://github.com/ccmbioinfo/stager/blob/master/flask/app/models.py#L378-L382

flask/app/participants.py
```diff
@@ -237,6 +243,7 @@ def create_participant():
         affected=request.json.get("affected"),
         solved=request.json.get("solved"),
         participant_type=request.json.get("participant_type"),
+        institution_id=request.json.get("institution_id"),
```
kevinlul, line 246: Can you make this endpoint accept the text form of institution instead and do the same for `/api/_bulk`?

Madelinehazel: revised in 4f944d550680490e9aded34026c3e4d37721f918

flask/app/participants.py
```diff
@@ -235,6 +235,20 @@ def create_participant():
...
+    if institution:
+        institution_obj = models.Institution.query.filter(
+            models.Institution.institution == institution
+        ).one_or_none()
+        if institution_obj:
+            institution_id = institution_obj.institution_id
+        else:
+            institution_obj = models.Institution(institution=institution)
+            db.session.add(institution_obj)
+            transaction_or_abort(db.session.commit)
```
kevinlul, line 249:
```suggestion
            transaction_or_abort(db.session.flush)
```

kevinlul (approved):
LGTM

PR #350, opened 2021-01-28
React: Refactor /api/users hookup to use react-query

frewmack:
Main changes are made to the Admin and Settings pages. All /api/users fetch requests and server state handling is replaced with custom hooks.

All custom hooks were renamed to match the react-query naming scheme (use{Name}Query, use{Name}{Action}Mutation).

Query-type hooks were all changed to return the useQuery result instead of just the data, which allows more nuanced handling of loading and error states when fetching. May need to update pages in the future to include loading / error displays.

### DIFF
--- a/react/src/Admin/components/CreateUserModal.tsx
+++ b/react/src/Admin/components/CreateUserModal.tsx
@@ -96,7 +96,7 @@ export default function CreateUserModal(props: CreateUserModalProps) {
     const [state, dispatch] = useReducer(reducer, initState);
     const [errorCode, setErrorCode] = useState(0);
     const [errorDetails, setErrorDetails] = useState({ error: "", message: "" });
-    const groups = useGroupsQuery();
+    const { data: groups } = useGroupsQuery();
     const userCreateMutation = useUsersCreateMutation();
     const { enqueueSnackbar } = useSnackbar();
 
--- a/react/src/Admin/components/CreateUserModal.tsx
+++ b/react/src/Admin/components/CreateUserModal.tsx
@@ -13,8 +13,7 @@ import {
 } from "@material-ui/core";
 import { NewUser } from "../../typings";
 import GroupSelect from "./GroupSelect";
-import { useGroups } from "../../hooks";
-import { useUserPost } from "../../hooks/users/useUserPost";
+import { useGroupsQuery, useUsersCreateMutation } from "../../hooks";
 import { useSnackbar } from "notistack";
 
 const initState = {
@@ -97,8 +96,8 @@ export default function CreateUserModal(props: CreateUserModalProps) {
     const [state, dispatch] = useReducer(reducer, initState);
     const [errorCode, setErrorCode] = useState(0);
     const [errorDetails, setErrorDetails] = useState({ error: "", message: "" });
-    const groups = useGroups();
-    const userCreateMutation = useUserPost();
+    const groups = useGroupsQuery();
+    const userCreateMutation = useUsersCreateMutation();
     const { enqueueSnackbar } = useSnackbar();
 
     const submitting = userCreateMutation.status === "loading";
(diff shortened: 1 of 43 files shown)


react/src/Admin/components/CreateUserModal.tsx
```diff
@@ -13,16 +13,8 @@ import {
...
-interface CreateUserState {
-    username: string;
-    email: string;
-    is_admin: boolean;
-    password: string;
-    confirmPassword: string;
-    groups: string[];
-}
+import { useGroupsQuery, useUsersCreateMutation } from "../../hooks";
+import { useSnackbar } from "notistack";
```
kevinlul, line 17: I think this is the wrong import order in this file

react/src/Admin/components/UserRow.tsx
```diff
@@ -125,14 +98,15 @@ export default function UserRow(props: {
...
-                    />
+                    {open && (
+                        <div id={props.user.username}>
+                            <UserDetails
+                                user={props.user}
+                                onSave={props.onSave}
+                                onDelete={props.onDelete}
+                            />
+                        </div>
+                    )}
```
kevinlul, line 109: What's the reason for this change?

react/src/hooks/groups/useGroupsCreateMutation.tsx
```diff
@@ -0,0 +1,28 @@
...
+        group_code: newGroup.group_code.toLowerCase(),
+    });
+}
+
+/**
+ * Return a mutation object for POST /api/groups.
+ *
+ * Used for creating a new group.
+ */
+export function useGroupsCreateMutation() {
```
kevinlul, line 17: This only creates one group so I'm wondering if the naming convention makes sense if we're naming mutations after what they do instead of the HTTP verb.

react/src/hooks/groups/useGroupsCreateMutation.tsx
```diff
@@ -0,0 +1,28 @@
...
+    const queryClient = useQueryClient();
+    const mutation = useMutation<Group, Response, Group>(postNewGroup, {
+        onSuccess: (receivedGroup, sentGroup) => {
+            queryClient.setQueryData(["groups", sentGroup.group_code.toLowerCase()], receivedGroup);
+            // destructuring to keep consistent with GET /api/groups format
+            const { users, ...listableGroup } = receivedGroup;
+            addToCachedList<Group>("groups", queryClient, listableGroup);
```
kevinlul, line 24: Neat!

react/src/hooks/utils.tsx
```diff
@@ -1,14 +1,104 @@
...
+) {
+    const response = await fetch(url, {
+        method: method,
+        credentials: "same-origin",
+        headers: body ? { "Content-Type": "application/json" } : undefined,
+        body: body ? JSON.stringify(body) : undefined,
+    });
+    if (response.ok) {
+        if (overrides?.onSuccess) return overrides.onSuccess(response);
+        return response.json();
```
kevinlul, line 42: You can actually use optional chaining with function calls too, but this might be confusing to read
```ts
overrides?.onSuccess?.(response) ?? await response.json();
```
So the return type of this handler is not necessarily the response because of the overrides, and it doesn't necessarily throw if onError doesn't throw?

react/src/Admin/components/UserRow.tsx
frewmack: For props, most of the state handling stuff is handled by react-query so some props were no longer needed.

For the wrapping div, I'm honestly not sure why I put that there 😅. I just took it out and it made zero difference so I'll go ahead with removing it.

react/src/hooks/groups/useGroupsCreateMutation.tsx
frewmack: I saw it as "groups" is the API name, and "create" is the general action. I can change it to be more specific.

react/src/hooks/utils.tsx
frewmack: This is actually a good point regarding the typing. I'm going with the assumption for now that anyone who adds their own overrides will know what they're doing, but I can look into fixing the typing.

kevinlul (approved):
🚀

PR #351, opened 2021-01-29
Flask: deny MinIO delete actions

### LINKED ISSUE #314: Remove MinIO object delete permissions from policy

kevinlul:
Closes #314

This doesn't prevent overwriting the object with a blank file because that's the same permission in S3. At most we could enable [object versioning](https://stackoverflow.com/questions/10592541/amazon-s3-acl-for-read-only-and-write-once-access on the [instance](https://docs.minio.io/docs/minio-bucket-versioning-guide.html). We should revisit the subject with our collaborators.

To test this, you probably need to reset your db and local minio so it can recreate the policies correctly and not make assumptions based on the current state.

### DIFF
--- a/flask/app/groups.py
+++ b/flask/app/groups.py
@@ -6,7 +6,7 @@ from sqlalchemy.orm import contains_eager, joinedload
 from .utils import check_admin, transaction_or_abort, mixin
 
 from minio import Minio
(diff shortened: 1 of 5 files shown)


delvinso (approved):
LGTM

PR #352, opened 2021-01-29
React: Hookup Participants Page with progressive loading using react-query 

### LINKED ISSUE #284: Epic: use react-query to reduce backend hits and enable progressive loading for the main tables
Depends on #283 

Depends on #219 to standardize error handling on the frontend

- [x] /api/groups*
- [x] /api/users* 
- [x] /api/enums
- [x] /api/metadatasettypes
- [x] /api/families
- [x] /api/participants
- [x] /api/datasets
- [ ] /api/analyses
- [ ] /api/unlinked
- [x] POST /api/_bulk

frewmack:
#284 

Known issues:
- Filtering for boolean columns does not work 
  - `GET /api/participants?filter=affected;like;true`, response: `[]`
  - Still working on mitigating this on frontend using `eq` operator for boolean columns
- ~~JSON Response needs to include metadata including page number, and number of total participants _after_ filtering, _before_ limiting~~
  - QueryResult format here: https://github.com/mbrn/material-table/blob/master/types/index.d.ts#L95
  - Next / Previous page buttons are greyed out until `totalCount` is added to response JSON.
- Enumerable columns (eg. `participant_type`, `sex`) can only be filtered one at a time, but front-end supports filtering by any subset of values in enum (eg. filter by `participant_type` for `Parent` or `Sibling`)
- Columns not part of the Participant model cannot be filtered by (`family_codename`, `dataset_types`)
- Tests needed for new features
  - Filtering
    - String column (eg. `participant_codename`)
    - Boolean column (eg. `affected`, `solved`)
    - Enum column (eg. `participant_type`). One or more values may be entered
  - Pagination, limit
  - Order by, order direction


react/src/hooks/utils.tsx
```diff
@@ -20,6 +22,54 @@ export async function basicFetch(url: string, options?: FetchOptions) {
...
+    // page information
+    searchParams.append("page", `${query.page + 1}`);
+    searchParams.append("limit", `${query.pageSize}`);
+
+    const response = await fetch(url + "?" + searchParams.toString());
+    if (response.ok) {
+        const result = await response.json();
+        return {
+            data: result.data,
+            page: result.page - 1,
```
kevinlul, line 65: Okay, if we need to do `-1` and `+1` here, might it be better to change the backend to use zero-indexing? This would also make more sense with how indexing in both languages works anyway.

flask/app/participants.py
```diff
@@ -94,6 +94,8 @@ def list_participants():
         )
         for f in filt:
             participants = participants.filter(f)
+
+        totalCount = participants.count()
```
kevinlul, line 98: Use snake_case in Python please

flask/app/participants.py
```diff
@@ -124,25 +126,30 @@ def list_participants():
...
+                    "tissue_samples": [
+                        {**asdict(tissue_sample), "datasets": tissue_sample.datasets}
+                        for tissue_sample in participant.tissue_samples
+                    ],
+                }
+                for participant in participants
+            ],
+            "page": int(page),
+            "totalCount": totalCount,
+        }
```
kevinlul, line 152: Do you need help updating the tests to match?

react/src/hooks/utils.tsx
```diff
@@ -1,5 +1,7 @@
 // Function patterns that commonly occur in this set of hooks.
 
+import { Query as MTQuery, QueryResult as MTQueryResult } from "material-table";
```
kevinlul, line 3: Does the name conflict with a different import? The default name is used in a different file above.

flask/app/participants.py
frewmack: I actually forgot about the tests. I'll try updating it myself.

react/src/hooks/utils.tsx
frewmack: I think it's fine this way. Page isn't used for indexing on the backend. It's just used to calculate the offset and needs to start at 1.

kevinlul: If you look at line 56 on the backend, `offset = (int(page) * int(limit)) - int(limit)`, this is actually subtracting 1 from page as well: (page - 1) * limit

frewmack: Oh you're right. Starting at 0 makes a lot more sense then.

kevinlul (approved):
LGTM 🚀

Issue #318, opened 2021-01-19
Update frontend to use /api/metadatasettypes


kevinlul:
- [x] Replace dataset type chip buttons with meta-dataset type chip buttons that configure the filter for all dataset types that are this meta-type (requires #317)
- [ ] Optionally, prevent analysis submissions that would fail due to incompatible types (requires #316)
- [ ] Once this is done, the `/api/enums` workaround can be removed https://github.com/ccmbioinfo/stager/blob/d8424adf838daa472e181010fd2daa5021ccd263/flask/app/routes.py#L123-L132

kevinlul:
Not implemeting due to #1055

PR #359, opened 2021-02-02
React: use /api/metadatasettypes, switch chip bar type

kevinlul:
- useMemo for postprocessing of /api/enums
- add typing for enums passed down as prop

#318

### DIFF
--- a/react/src/AddDatasets/components/utils.tsx
+++ b/react/src/AddDatasets/components/utils.tsx
@@ -82,7 +82,7 @@ export function getOptions(
     col: DataEntryHeader,
     rowIndex: number,
     families: Family[],
-    enums: any,
+    enums: Record<string, string[]> | undefined,
     files: string[]
 ): Option[] {
     const row = rows[rowIndex];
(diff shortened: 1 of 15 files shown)


react/src/hooks/useMetadatasetTypesQuery.tsx
```diff
@@ -0,0 +1,14 @@
+import { useQuery } from "react-query";
+import { basicFetch } from "./utils";
+
+/**
+ * Return JSON result of GET /api/metadatasettypes, or undefined if the fetch is unsuccessful.
```
frewmack, line 5: The query object is always returned, and its `data` field is what may be undefined if the fetch is unsuccessful. I noticed the comment is based on `useEnumsQuery` which is also out-of-date after I changed the hooks return the query object as opposed to just the result.

react/src/AddDatasets/components/utils.tsx
```diff
@@ -196,7 +199,7 @@ export function checkParticipant(participant: Participant, enums: any): boolean
 
             if (enumerableColumns.includes(column)) {
                 // if enumerable, must be valid value
-                return enums[index].includes(participant[column]);
+                return enums?.[index].includes(participant[column]);
```
frewmack, line 202: Neat! I didn't know you could use optional chaining like this.

frewmack (approved):
Works for me! Only suggestion is to update the comments of some query hooks.

PR #369, opened 2021-02-03
React: Move signout button, add CCM logo

### LINKED ISSUE #363: Change signout button location and add CCM logo
Kind of open-ended for where the logo should go, but peopel said the signout button is more intuitive on the top right.

frewmack:
Closes #363


react/src/Navigation.tsx
```diff
@@ -253,6 +258,11 @@ export default function Navigation({ signout, darkMode, toggleDarkMode }: Naviga
                         <Tooltip title={"Logged in as " + currentUser.username} arrow>
                             <AccountCircleIcon fontSize="large" />
                         </Tooltip>
+                        <Tooltip title="Log out">
```
kevinlul, line 261:
```suggestion
                        <Tooltip title="Sign out">
```
Match how it was before (and sign in)

kevinlul (commented):
I think we might need to change the app bar's colour if the logo is staying there because the contrast between the dark portion of the logo and the background isn't great.

image

frewmack:
> I think we might need to change the app bar's colour if the logo is staying there because the contrast between the dark portion of the logo and the background isn't great.

Yeah I agree. I might look into how themes are handled in Material UI so that we can make changes like this more easily across the app.

frewmack:
I decided to move the logo to the drawer, which gets hidden when the drawer is closed. The colours contrast nicely with both the light and dark theme backgrounds, and this avoids us having to come up with a colour scheme that accounts for the logo's colours for now.

kevinlul (approved):
light
dark

This is a W!

PR #371, opened 2021-02-04
Flask: reduce data overfetch from participants

Madelinehazel:
Part of #283

Submitting as draft PR because I'm sure it will need major revisions. Also, one test is failing (test_list_participants_admin). and I'm not sure why:
`      assert response.get_json()["total_count"] == 3
E       assert 4 == 3`

 Things to consider:

Should filter queries be submitted as a json?
Should enums (e.g. sex, participant type) be validated?

Example test:
`curl "http://localhost:3000/api/participants?family_codename=2002&affected=true"`
Returns:
```
{
  "data": [
    {
      "affected": true, 
      "created": "2021-02-03T18:15:53", 
      "created_by": "admin", 
      "family_codename": "2002", 
      "institution": "Children's Hospital of Eastern Ontario", 
      "month_of_birth": null, 
      "notes": null, 
      "participant_aliases": null, 
      "participant_codename": "CHEO0007", 
      "participant_id": 7, 
      "participant_type": "Proband", 
      "sex": "Male", 
      "solved": null, 
      "tissue_samples": [
        {
          "created": "2021-02-03T18:15:53", 
          "datasets": [
            {
              "batch_id": null, 
              "capture_kit": null, 
              "condition": "GermLine", 
              "created": "2021-02-03T00:00:00", 
              "dataset_id": 7, 
              "dataset_type": "RGS", 
              "extraction_protocol": null, 
              "library_prep_date": null, 
              "library_prep_method": null, 
              "linked_files": [], 
              "notes": null, 
              "read_length": null, 
              "read_type": null, 
              "sequencing_centre": null, 
              "sequencing_date": null, 
              "sequencing_id": null, 
              "tissue_sample_id": 7, 
              "updated": "2021-02-03T18:15:53"
            }
          ], 
          "extraction_date": null, 
          "notes": null, 
          "participant_id": 7, 
          "tissue_processing": null, 
          "tissue_sample_id": 7, 
          "updated": "2021-02-03T18:15:53"
        }
      ], 
      "updated": "2021-02-03T18:15:53", 
      "updated_by": "admin"
    }
  ], 
  "page": 0, 
  "total_count": 1
}
```

### DIFF
--- a/flask/app/participants.py
+++ b/flask/app/participants.py
@@ -11,13 +11,7 @@ from sqlalchemy import exc, asc, desc
 from sqlalchemy.orm import aliased, contains_eager, joinedload
 from werkzeug.exceptions import HTTPException
 
-from .utils import (
-    check_admin,
-    transaction_or_abort,
-    mixin,
-    enum_validate,
-    string_to_bool,
-)
+from .utils import check_admin, transaction_or_abort, mixin, enum_validate, filter_bool
 
 
 editable_columns = [
@@ -63,9 +57,7 @@ def list_participants():
     solved = request.args.get("solved", default=None, type=str)
     family_codename = request.args.get("family_codename", default=None, type=str)
     family_codename = f"%{family_codename}%" if family_codename else "%"
-    dataset_type = request.args.get("dataset_type", default=None, type=str)
 
-    # enum_error = enum_validate(models.Participant, request.json, editable_columns)
     # for some reason type=int doesn't catch non-integer queries
     try:
         int(limit)
@@ -81,10 +73,13 @@ def list_participants():
 
     columns = models.Participant.__table__.columns.keys()
 
-    try:
-        order_column = getattr(models.Participant, order_by_col)
-    except AttributeError:
-        return f"Column name must be one of {columns}", 400
+    if order_by_col == "family_codename":
+        order_column = getattr(models.Family, order_by_col)
+    else:
+        try:
+            order_column = getattr(models.Participant, order_by_col)
+        except AttributeError:
+            return f"Column name must be one of {columns}", 400
 
     try:
         order = getattr(order_column, order_dir)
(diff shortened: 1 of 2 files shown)


flask/app/utils.py
```diff
@@ -66,41 +67,9 @@ def enum_validate(
...
-            except IndexError:
-                return f"Invalid filter operator: {op}"
-            if value == "null":
-                value = None
-            filt = getattr(column, attr)(value)
-        sql_filters.append(filt)
-    return sql_filters
+def string_to_bool(string: str) -> Union[None, str]:
+    str_bool = loads(string.lower())
+    return str_bool
```
kevinlul, line 72: This actually parses JSON and the return typing is off. It would be better in the cases for affected and solved to just check for `== "true"` and `== "false"`

flask/app/participants.py
```diff
@@ -57,19 +81,6 @@ def list_participants():
...
-    else:
-        filt = [
-            (
-                models.Participant.participant_codename
-                == models.Participant.participant_codename
-            )
-        ]
-
     try:
         order_column = getattr(models.Participant, order_by_col)
```
frewmack, line 85: We should be able to order by `family_codename` as well. Not sure if we should allow ordering by `dataset_types` because I'm not sure what the expected behaviour should be.

frewmack (changes requested):
Filtering works as expected. I updated the front-end and filtering works perfect. Only bug I found was in regards to `order_by`.

kevinlul (commented):
I'm thinking we can merge this first and resolve the count issue and other endpoints on top of this work. I'm guessing the tests no longer fail because the test database doesn't contain an example of gaining access to a dataset through multiple permission groups, and after dropping support for the dataset_type filter, that's the only means we can multiply the resulting rows?

kevinlul (approved):

PR #372, opened 2021-02-04
React: fix potentially undefined enum indexing and remove extraction protocol

kevinlul:


frewmack (changes requested):
I think we should remove `extraction_protocol` from DataEntryRowOptional as well in typings

kevinlul:
Sorry if it wasn't clear — extraction protocol is not being removed, it's just no longer an enum.

frewmack (approved):
🚀

PR #375, opened 2021-02-05
React: Fix groups mutations to only update group list

frewmack:
Invalidating queries with option `{ exact: true }` ensures that only queries that exactly match the given key will be invalidated, instead of all queries that contain it. The group create, update, delete hooks now use this option so that only the /api/groups query is invalidated, and not other /api/groups/:id queries that aren't affected by the action.


kevinlul (approved):
LGTM!

PR #376, opened 2021-02-05
Flask: add global error handler to return json response 

### LINKED ISSUE #219: Add global error handlers that return a JSON response

delvinso:
closes #219 

- ~~added global error handler which is registered to werkzeug HTTP exceptions. Whenever an error is thrown, a json object with structure `{“error”: “description”} ` is now returned~~
- added generic, global error handler as a blueprint which returns a json object with structure `{“error”: “description”}. also handles werkzeug HTTP exceptions
- 
- couple of minor fixes to test as some did not expect json output (eg in test_login.py)
- changes to users.py to use global error handler and corresponding tests 
```
curl -D - -X PATCH \                                 
    --data '{
         "notes": "hello",
         "analysis_state": "Dofne"
        }' \
    localhost:5000/api/analyses/2
HTTP/1.0 400 BAD REQUEST
Content-Length: 106
Date: Fri, 05 Feb 2021 20:27:44 GMT

{
  "error": "\"analysis_state\" must be one of ['Requested', 'Running', 'Done', 'Error', 'Cancelled']"
}
```
```
curl -D - -X DELETE localhost:5000/api/participants/1
HTTP/1.0 422 UNPROCESSABLE ENTITY
Content-Length: 63
Date: Fri, 05 Feb 2021 20:27:37 GMT

{
  "error": "Participant has tissue samples, cannot delete"
}
```


kevinlul:
Don't worry about the CodeQL warning. I'll have to think about it.

kevinlul:
I'm confused as to why the [automatic run](https://github.com/ccmbioinfo/stager/actions/runs/547946160) on `master` didn't also pick this up...

flask/app/__init__.py
```diff
@@ -27,6 +28,9 @@ def create_app(config):
     app.config.from_object(config)
     app.json_encoder = DateTimeEncoder
 
+    for ex in default_exceptions:
+        app.register_error_handler(ex, handle_error)
```
kevinlul, line 32: Can this be registered generically for all exceptions instead? I got an HTML response after disconnecting the database.

kevinlul (commented):
Can you remove the old on_http_exception as this will succeed it?

flask/app/__init__.py
delvinso: See: 9d788b98472d4f96cb16de571a04991f7c9e238c 

Also added the handler as a blueprint. The drawback with the generic handler is similar to try catching an Exception, in this case all unhandled exceptions (non HTTP) will default to 500, but at least the error message is captured?

kevinlul: That's what should happen! There's no other appropriate error code.

kevinlul (approved):
LGTM

PR #386, opened 2021-02-10
React: Hookup datasets page using react-query

frewmack:
Part of #284

PR #387, opened 2021-02-10
React: Hookup unlinked files with react-query

frewmack:
Part of #284

### DIFF
--- a/react/src/hooks/unlinked/useUnlinkedFilesQuery.tsx
+++ b/react/src/hooks/unlinked/useUnlinkedFilesQuery.tsx
@@ -1,4 +1,5 @@
 import { useQuery } from "react-query";
+import { basicFetch } from "../utils";
 
 async function fetchFiles() {
     return await basicFetch("/api/unlinked");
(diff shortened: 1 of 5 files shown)


react/src/hooks/unlinked/useUnlinkedFilesQuery.tsx
```diff
@@ -0,0 +1,19 @@
...
+/**
+ * Return result of GET /api/unlinked.
+ *
+ * That is, return a sorted list of filenames
+ * for all unlinked files in MinIO.
+ */
+export function useUnlinkedFilesQuery() {
+    const result = useQuery<string[], Response>("unlinked", fetchFiles);
+    if (result.isSuccess)
+        result.data = result.data.sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()));
```
kevinlul, line 17: Should we sort on the backend instead and the frontend is only responsible for keeping its own representation sorted?

kevinlul (approved):
LGTM!

react/src/hooks/unlinked/useUnlinkedFilesQuery.tsx
frewmack: Yeah that might be a good idea. I don't think there's any frontend actions that add to unlinked files, only remove, so no extra care needs to be taken to keep the list sorted.

kevinlul: Some frontend actions do add to the list but care was already taken then to sort the list before the state is updated. You can just call `.sort()` or `sorted` on the list before it's passed to `jsonify` in the endpoint.

kevinlul:
Merging #386 resulted in an nontrivial merge conflict here 👀

PR #388, opened 2021-02-11
Flask: finish implementing progressive loading and filtering on /api/participants

kevinlul:
Update default data to provide multiple children in one-to-many relationships to help test .count() behaviour.
Fix a severe N+1 type problem in this endpoint and potential security problems with user input feeding into `getattr`

#283

### DIFF
--- a/flask/app/participants.py
+++ b/flask/app/participants.py
@@ -112,7 +112,9 @@ def list_participants() -> Response:
         filters.append(filter_in_enum_or_abort(models.Participant.sex, models.Sex, sex))
     affected = request.args.get("affected", type=str)
     if affected:
-        filters.append(filter_nullable_bool_or_abort(models.Participant.affected, affected))
+        filters.append(
+            filter_nullable_bool_or_abort(models.Participant.affected, affected)
+        )
     solved = request.args.get("solved", type=str)
     if solved:
         filters.append(filter_nullable_bool_or_abort(models.Participant.solved, solved))
--- a/flask/app/participants.py
+++ b/flask/app/participants.py
@@ -89,15 +89,15 @@ def list_participants() -> Response:
     filters = []
     family_codename = request.args.get("family_codename", type=str)
     if family_codename:
-        filters.append(models.Family.family_codename.like(f"%'{family_codename}'%"))
+        filters.append(func.instr(models.Family.family_codename, family_codename))
     participant_codename = request.args.get("participant_codename", type=str)
     if participant_codename:
         filters.append(
-            models.Participant.participant_codename.like(f"%'{participant_codename}'%")
+            func.instr(models.Participant.participant_codename, participant_codename)
         )
     notes = request.args.get("notes", type=str)
     if notes:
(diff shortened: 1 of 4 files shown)


kevinlul:
I messed up escaping user input to LIKE. We need to make sure that when users search for `%` and similar, it's searching for a literal `%` and not interpreted.

flask/app/participants.py
```diff
@@ -33,75 +38,101 @@
...
+                models.ParticipantType,
+                participant_type,
+            )
+        )
+    sex = request.args.get("sex", type=str)
+    if sex:
+        filters.append(filter_in_enum_or_abort(models.Participant.sex, models.Sex, sex))
+    affected = request.args.get("affected", type=str)
+    if affected:
+        filters.append(filter_nullable_bool_or_abort(models.Participant, affected))
```
Madelinehazel, line 115:
```suggestion
        filters.append(filter_nullable_bool_or_abort(models.Participant.affected, affected))
```

flask/app/participants.py
```diff
@@ -33,75 +38,101 @@
...
+    solved = request.args.get("solved", type=str)
+    if solved:
+        filters.append(filter_nullable_bool_or_abort(models.Participant, solved))
```
Madelinehazel, line 118:
```suggestion
        filters.append(filter_nullable_bool_or_abort(models.Participant.solved, solved))
```

flask/app/participants.py
```diff
@@ -33,75 +38,101 @@
...
+    if family_codename:
+        filters.append(func.instr(models.Family.family_codename, family_codename))
+    participant_codename = request.args.get("participant_codename", type=str)
+    if participant_codename:
+        filters.append(
+            func.instr(models.Participant.participant_codename, participant_codename)
+        )
+    notes = request.args.get("notes", type=str)
+    if notes:
+        filters.append(func.instr(models.Participant.notes, notes))
```
Madelinehazel, line 100: We also need to be able to handle 'notes=null'

Madelinehazel (changes requested):
Nice work!

flask/app/participants.py
kevinlul: What would that do?

frewmack (commented):
Frontend works as expected after slight changes to accommodate "{column}={value}" format. Should I push the changes here instead of a separate PR?

delvinso (approved):
Great work, this looks like it was a huge undertaking! Looks good to me.

kevinlul:
Separate PR is fine so this doesn't get more messy than it needs to

frewmack (approved):

flask/app/participants.py
Madelinehazel: Get records where there are no notes associated with the participant. Right now, if you pass notes=null, no participants are returned. BUT, I guess this isn't relevant from a use-case point of view (presumably, we wouldn't need to filter for participants that do not have notes associated with them).

PR #389, opened 2021-02-11
Flask: implement progressive loading and filtering on /api/datasets

kevinlul:
#283

### DIFF
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -1,5 +1,4 @@
 from dataclasses import asdict
-from datetime import datetime
 
 from flask import Blueprint, Response, abort, current_app as app, jsonify, request
 from flask_login import current_user, login_required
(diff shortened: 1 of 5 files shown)


kevinlul:
Just discovered that the `updated_by` filter doesn't work. Working on it

frewmack:
Having an issue filtering by dataset_type. Looking into it right now
```
curl "http://localhost:3000/api/datasets?dataset_type=RGS&page=0&limit=10"
{
  "error": "__init__() takes 1 positional argument but 2 were given"
}
```

kevinlul:
This is my mistake. DatasetType isn't an enum anymore, so it's trying to call the initializer for an entity rather than an enum.

kevinlul:
I think I should add more comprehensive test cases for the new behaviours but it could take some time.

frewmack (approved):
Looks good to me! Frontend filtering works as expected.

PR #399, opened 2021-02-16
Docker: lock MinIO to previous version to fix CI failures

kevinlul:
minio/minio#11544

Should unblock all backend PRs


frewmack (approved):
LGTM!

kevinlul:
It's still broken though 😭

kevinlul:
It was both MinIO and the command-line client...

frewmack (approved):
Good catch! Backend checks pass so I think we're clear.

PR #400, opened 2021-02-16
React: Hookup analyses page with react-query

frewmack:
Part of #284

### DIFF
--- a/react/src/Analyses/index.tsx
+++ b/react/src/Analyses/index.tsx
@@ -507,11 +507,7 @@ export default function Analyses() {
                     editable={{
                         onRowUpdate: async (newData, oldData) => {
                             analysisUpdateMutation.mutate(
-                                {
-                                    analysis_id: newData.analysis_id,
-                                    notes: newData.notes,
-                                    result_path: newData.result_path,
-                                },
+                                { ...newData, source: "row-edit" },
                                 {
                                     onSuccess: newRow => {
                                         enqueueSnackbar(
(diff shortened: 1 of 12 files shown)


kevinlul (commented):
When I mark a single analysis as Running from Requested through the frontend, the entire table is blanked unless the analysis was already Running.

frewmack:
> 
> 
> When I mark a single analysis as Running from Requested through the frontend, the entire table is blanked unless the analysis was already Running.

This happens because material-table keeps track of selected rows by injecting `tableData: { checked: true }` into selected row objects, and this field is lost when the rows are re-fetched after being updated. Currently working on a fix.

kevinlul (commented):
This is still glitchy when testing, like fields get blanked in the display or some actions like cancel don't work. Does this solution still have problems with adapting to progressive loading?

frewmack:
Yeah I think this needs more work. If it's the Pipeline field getting blanked out then I know how to fix it right away.

kevinlul (approved):
Just need to merge in master

frewmack:
There's still a couple fixes that I want to make. I got cancel working but material-table doesn't seem to like when dialogs open and it undoes row selections.

frewmack:
All the analysis table functions should work as expected now. 

Only weirdness still present that I can't figure out is that the table will de-select rows when the Cancel Dialog is open, and re-select rows when it is closed (which is an issue that I feel like has come up before). It doesn't seem to affect the cancellation process now, but it still looks off.

PR #403, opened 2021-02-17
React: Fix current user groups not being updated in admin

### LINKED ISSUE #374: Bug: Updating current user's groups in Admin page does not update groups in Add Datasets
If you are logged in as an admin, changing your own groups via the Admin page does not change the list of permission groups in the Add Datasets page unless the user refreshes the page. This is due to the current user groups only being set at login.

frewmack:
Closes #374


kevinlul (approved):
LGTM

PR #406, opened 2021-02-17
Flask: implement progressive loading and filtering on /api/analyses

### LINKED ISSUE #283: Epic: reducing data overfetch to allow for progressive loading

kevinlul:
Branched from #389, to be reviewed after

Fix some N+1 bugs
Closes #283

### DIFF
--- a/flask/app/analyses.py
+++ b/flask/app/analyses.py
@@ -35,6 +35,7 @@ def list_analyses(page: int, limit: int) -> Response:
         "result_path",
         "notes",
         "analysis_state",
+        "pipeline_id",
         "assignee",
         "requester",
     ]
@@ -87,6 +88,16 @@ def list_analyses(page: int, limit: int) -> Response:
                 analysis_state,
             )
         )
+    pipeline_id = request.args.get("pipeline_id", type=str)
+    if pipeline_id:
+        try:
+            filters.append(
+                models.Analysis.pipeline_id.in_(
+                    [int(pk) for pk in pipeline_id.split(",")]
+                )
+            )
+        except ValueError as err:
+            abort(400, description=err)
 
     if app.config.get("LOGIN_DISABLED") or current_user.is_admin:
         user_id = request.args.get("user")
--- a/flask/app/analyses.py
+++ b/flask/app/analyses.py
@@ -1,19 +1,23 @@
-from datetime import datetime
-import json
-from typing import Any, Callable, Dict, List, Union
 from dataclasses import asdict
+from datetime import datetime
+from typing import Container
 
-from .extensions import db, login
+from flask_login import current_user, login_required
+from sqlalchemy import distinct, func
+from sqlalchemy.orm import aliased, contains_eager
+
+from flask import Blueprint, Response, abort, current_app as app, jsonify, request
+
 from . import models
-
-from flask import abort, jsonify, request, Response, Blueprint, current_app as app
-from flask_login import login_user, logout_user, current_user, login_required
-from sqlalchemy import exc
-from sqlalchemy.orm import contains_eager, aliased, joinedload
-from werkzeug.exceptions import HTTPException
-
-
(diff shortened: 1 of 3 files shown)


flask/app/analyses.py
```diff
@@ -23,60 +27,114 @@
...
     if app.config.get("LOGIN_DISABLED") or current_user.is_admin:
         user_id = request.args.get("user")
     else:
         user_id = current_user.user_id
-
-    since_date = request.args.get("since", default="0001-01-01T00:00:00-04:00")
-    try:
-        since_date = datetime.fromisoformat(since_date)
-    except:
-        abort(400, description="Malformed query date")
```
frewmack, line 36: The `since` parameter is needed for the notifications feature. It's only ever invoked by itself, without any other query parameters.

flask/app/analyses.py
```diff
@@ -23,60 +27,114 @@
...
+def list_analyses(page: int, limit: int) -> Response:
+    order_by = request.args.get("order_by", type=str)
+    allowed_columns = [
+        "updated",
+        "result_path",
+        "notes",
+        "analysis_state",
+        "assignee",
+        "requester",
+    ]
```
frewmack, line 40: `analysis_id` and `pipeline` should be allowed as well, where `pipeline` is a comma-separated list of `pipeline_id`'s (the frontend doesn't support this yet but I can update the filter component to just be a checkbox list of pipelines available).

frewmack (changes requested):
Looks good for the most part. Spotted some missing details

flask/app/analyses.py
kevinlul: I expect that it can be updated to use `updated=after,iso-datetime`

frewmack: Ah that makes sense.

flask/app/analyses.py
kevinlul: Can we just use the individual GET endpoint for `analysis_id`?

frewmack: I guess so. Meanwhile we can either hide the `analysis_id` column in the table, or disable filtering for that column.

frewmack (approved):
LGTM!

PR #408, opened 2021-02-18
React: Update participants hookup for progressive loading

frewmack:
Goes with #388

### DIFF
--- a/react/src/hooks/utils.tsx
+++ b/react/src/hooks/utils.tsx
@@ -46,15 +46,13 @@ export async function queryTableData<RowData extends object>(
     // add filters
     for (let filter of query.filters) {
         const isBoolean = ["affected", "solved"].includes("" + filter.column.field);
-        let operator = "like";
         let value = filter.value;
         // booleans use "eq" as operator
         if (isBoolean) {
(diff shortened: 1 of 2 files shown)


react/src/hooks/utils.tsx
```diff
@@ -46,15 +46,13 @@ export async function queryTableData<RowData extends object>(
...
         let value = filter.value;
         // booleans use "eq" as operator
         if (isBoolean) {
-            operator = "eq";
             value = stringToBoolean(filter.value);
         }
-        if (filter.value || isBoolean) {
-            searchParams.append("filter", `${filter.column.field};${operator};${value}`);
+        if (filter.column.field && (filter.value || isBoolean)) {
+            searchParams.append(`${filter.column.field}`, `${value}`);
```
kevinlul, line 55: If these values are strings, you should be able to pass them literally.

frewmack: column.field may be a string, number or Symbol. filter.value can be anything but it's usually a string or an array of strings.

kevinlul (approved):
LGTM

PR #416, opened 2021-02-22
React: Add date filter component for datasets, analyses

### LINKED ISSUE #360: Date range filter component
Investigate the viability of and implement a date range componet for the `updated` filters on the Datasets and Analyses pages, for use with #283

frewmack:
Closes #360. 
Depends on #284 for /api/datasets, /api/analyses

Allows filtering before or after specific dates, and sends strings of the form `{before,after},yyyy-mm-dd` to the Query object used in material-table remote data feature.


react/src/components/DateFilterComponent.tsx
```diff
@@ -0,0 +1,76 @@
...
+    },
+}));
+
+/**
+ * A filter component for material-table for sorting before or after
+ * a specific date. Passes up a value of the format "{before,after},yyyy-mm-dd".
+ */
+export default function DateFilterComponent<RowData extends object>(props: {
+    columnDef: Column<RowData>;
+    onFilterChanged: (rowId: string, value: any) => void;
```
kevinlul, line 25: The typing for `value` should be `string`, right?

frewmack: The typing here matches that of the filterComponent field of Column here: https://github.com/mbrn/material-table/blob/master/types/index.d.ts#L179. I expect it to be a string, but material-table annoyingly does it this way.

I also can't avoid typing it again by typing it as `Parameters<Column<RowData>["filterComponent"]>[0]` because it gives me a big union of other types that don't work.

kevinlul (approved):

Issue #361, opened 2021-02-02
Add detailed logging to backend


kevinlul:
We currently have SQLAlchemy logs and the default Flask per-request logs in Apache-ish style. If something goes awry, we have little information to debug from the logs besides the response code. This is a decently large issue but otherwise easy issue to add logging (via the native Python logging module or `app.logger`) to aid in debugging production faults.

We can split this up into smaller items to be easier to code and review.

delvinso:
to keep track of modules which are currently being worked on/or are finished 
- [ ] analyses.py (overfetch WIP?)
- [ ] buckets.py
- [ ] datasets.py (overfetch WIP?)
- [ ] families.py 
- [ ] groups.py
- [ ] participants.py 
- [ ] routes.py 
- [ ] tissue_samples.py
- [ ] users.py 

unsure if utils.py, madmin.py require logging?

PR #419, opened 2021-02-25
Flask: add logging to datasets

Madelinehazel:
part of #361

Also added a requirement that "condition" is specified in the POST query since it is non-nullable in the database.

### DIFF
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -41,12 +41,16 @@ datasets_blueprint = Blueprint(
 @datasets_blueprint.route("/api/datasets", methods=["GET"])
 @login_required
 def list_datasets():
+    app.logger.debug("Retrieving user id...")
     if app.config.get("LOGIN_DISABLED") or current_user.is_admin:
         user_id = request.args.get("user")
+        app.logger.debug("User is admin with ID '%s'", user_id)
     else:
         user_id = current_user.user_id
+        app.logger.debug("User is regular with ID '%s'", user_id)
 
     if user_id:
+        app.logger.debug("Processing query - restricted based on user id")
         query = (
             models.Dataset.query.join(models.groups_datasets_table)
             .join(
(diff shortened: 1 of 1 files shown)


flask/app/datasets.py
```diff
@@ -177,26 +197,32 @@ def update_dataset(id: int):
...
         )
     else:
+        app.logger.debug("Processing query - unrestricted based on user id")
         dataset = models.Dataset.query.filter_by(dataset_id=id).first_or_404()
 
     enum_error = mixin(dataset, request.json, editable_columns)
 
     if enum_error:
+        app.logger.error("%s", enum_error)
         abort(400, description=enum_error)
```
delvinso, line 207: Looks good to me, not a big deal but I think we could add the following to the enum checks so it's in line with the rest of the endpoints
```suggestion
    app.logger.debug("Validating enums..")
    enum_error = mixin(dataset, request.json, editable_columns)

    if enum_error:
        app.logger.error("%s", enum_error)
        abort(400, description=enum_error)
    else:
        app.logger.debug("All enums supplied are valid.")
```

flask/app/datasets.py
```diff
@@ -219,52 +246,72 @@ def delete_dataset(id: int):
...
     models.TissueSample.query.filter_by(
         tissue_sample_id=tissue_sample_id
     ).first_or_404()
 
+    app.logger.debug("Validating enums")
     enum_error = enum_validate(models.Dataset, request.json, editable_columns)
 
     if enum_error:
+        app.logger.error("%s", enum_error)
         abort(400, description=enum_error)
```
delvinso, line 300:
```suggestion
    else:
        app.logger.debug("All enums supplied are valid.")
```

delvinso (approved):

kevinlul:
These are conflicting because of the progressive loading changes so I'm not sure how to proceed here.

PR #428, opened 2021-03-02
Docker: provide ST_SECRET_KEY, bake FLASK_ENV into image

kevinlul:


frewmack:
I'm getting the following error from logging in:
```
{ "error": "The session is unavailable because no secret key was set. Set the secret_key on the application to something unique and secret." }
```

It seems that Docker defaults the `ST_SECRET_KEY` env variable to an empty string if no such env variable is set, which breaks logging in if it's not set, and also broke all the pytests that use the login endpoint.

frewmack (approved):
LGTM

PR #430, opened 2021-03-03
React: Update datasets table to use progressive loading

frewmack:
Part of #284

~~Todo before review:~~ Will solve these issues in a separate PR. Seems to be much more complicated than just assigning columns to a state variable.

~~- [ ] Fix dataset filter buttons erasing other filters.~~
~~- [ ] Fix delayed disappearing filters after new query.~~


frewmack:
Two issues still present in this PR:

- Metadataset Type chip buttons clear all other filters.
- Filters disappear after some time; likely related to react-query refetching stale data.

I think these issues should be handled in a separate PR. Other than that, filtering and progressive loading should work as expected.

kevinlul (approved):
Sure, we'll go with that!

PR #442, opened 2021-03-10
React: Add pipeline filter for analyses page

frewmack:
Related to #406

Also fixes an issue with useAnalysesQuery, and updates the notification popover to use `updated=after,{date}` parameter format.

EDIT: Added analysis_id filter changes here because front-end on `master` is broken, so this branch is easier to test on.

### DIFF
--- a/react/src/Analyses/components/PipelineFilter.tsx
+++ b/react/src/Analyses/components/PipelineFilter.tsx
@@ -17,7 +17,7 @@ function pipelineName(p: Pipeline) {
  */
 export default function PipelineFilter(props: FilterProps) {
     const pipelinesQuery = usePipelinesQuery();
-    const pipelines = useMemo(() => {
+    const pipelinesObj = useMemo(() => {
         if (pipelinesQuery.isSuccess) {
             const obj: { [key: number]: Pipeline } = {};
             pipelinesQuery.data.forEach(p => {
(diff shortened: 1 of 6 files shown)


react/src/Analyses/components/PipelineFilter.tsx
```diff
@@ -0,0 +1,64 @@
...
+                <Select
+                    multiple
+                    value={selected}
+                    onChange={onFilterChange}
+                    onClose={onClose}
+                    renderValue={selected =>
+                        (selected as number[]).map(id => pipelineName(pipelines[id])).join(", ")
+                    }
+                >
+                    {Object.values(pipelines).map(p => (
```
kevinlul, line 54: If `pipelines` if only used in `Object.values` form then it may be useful to have `useMemo` perform this transformation and memoize it.

react/src/hooks/analyses/useAnalysesPage.tsx
```diff
@@ -14,9 +15,40 @@ async function fetchAnalyses(query: Query<Analysis>) {
...
-            fetchAnalyses(query)
-        );
-    };
+    const func = useCallback(
+        async (query: Query<Analysis>) => {
+            // If analysis_id is specified, we'll use the individual GET endpoint
+            const idFilter = query.filters.find(
+                filter => filter.column.field === "analysis_id" && filter.value
+            );
+            if (idFilter && typeof idFilter.value === "string") {
```
kevinlul, line 24: Is the type not well-defined?

frewmack: material-table types the filter values as "any". The string type-check is just to avoid unexpected behaviour and to make the filter value types play nice in the if-block.

kevinlul (approved):
LGTM

Issue #441, opened 2021-03-10
db schema for gene/variant level viewer


delvinso:
- [x] draft schema for variant viewer
- [x] implement schema through sqlalchemy orm

PR #443, opened 2021-03-10
Flask: update models.py and migrations for gene pull schema

delvinso:
part of #441


flask/app/models.py
```diff
@@ -398,3 +400,68 @@ class PipelineDatasets(db.Model):
...
+    )
+    conserved_in_20_mammals: int = db.Column(db.Float, nullable=True)
+    sift_score: int = db.Column(db.Float, nullable=True)
+    polyphen_score: int = db.Column(db.Float, nullable=True)
+    cadd_score: int = db.Column(db.Float, nullable=True)
+    gnomad_af: int = db.Column(db.Float, nullable=True)
+
+
+class Variants_Analyses_Association(db.Model):
+    "variants_analyses",
```
kevinlul, line 458: What's this random string for?

flask/app/models.py
```diff
@@ -398,3 +400,68 @@ class PipelineDatasets(db.Model):
...
+        db.ForeignKey("gene.gene_id", onupdate="cascade", ondelete="restrict"),
+    )
+    conserved_in_20_mammals: int = db.Column(db.Float, nullable=True)
+    sift_score: int = db.Column(db.Float, nullable=True)
+    polyphen_score: int = db.Column(db.Float, nullable=True)
+    cadd_score: int = db.Column(db.Float, nullable=True)
+    gnomad_af: int = db.Column(db.Float, nullable=True)
+
+
+class Variants_Analyses_Association(db.Model):
```
kevinlul, line 457:
```suggestion
class AnalyzedVariant(db.Model):
```

kevinlul (changes requested):

flask/app/models.py
delvinso: oops it was suppose to be the tablename since this was originally a `db.Table`, fixed now

flask/app/models.py
```diff
@@ -398,3 +398,69 @@ class PipelineDatasets(db.Model):
...
+        db.Integer,
+        db.ForeignKey("gene.gene_id", onupdate="cascade", ondelete="restrict"),
+    )
+    conserved_in_20_mammals: int = db.Column(db.Float, nullable=True)
+    sift_score: int = db.Column(db.Float, nullable=True)
+    polyphen_score: int = db.Column(db.Float, nullable=True)
+    cadd_score: int = db.Column(db.Float, nullable=True)
+    gnomad_af: int = db.Column(db.Float, nullable=True)
+
+    analyses = db.relationship("AnalyzedVariant", backref="analysis")
```
kevinlul, line 454: The backref will be on the `AnalyzedVariant` entity, and it will be a reference to a `Variant` entity. This means that it should be called `variant`. Then on `AnalyzedVariant`, we can define the corresponding `analysis` relationship. Given that the foreign keys are both defined in `AnalyzedVariant`, it might make more sense to define both relations there then, since a `Variant` is a less dynamic entity within the database.

kevinlul (commented):
The build failure is because of the new `Variation` enum, which causes corresponding `/api/enums` tests to fail. We can exclude this enum from that endpoint for now.

flask/app/models.py
delvinso: That's my bad, the backref should have referred to `variant`. If I understand correctly then I added the relationships only to the association table and I believe since I'm using`backref` the reverse relationship is automatically established.

kevinlul (approved):

PR #461, opened 2021-03-22
CI: only login to registry on push workflows

kevinlul:
ghcr.io/ccmbioinfo/stager is now public
Works around dependabot/dependabot-core#3253


frewmack (approved):
Looks good to me

PR #463, opened 2021-03-24
Flask: schema changes to gene pull viewer

delvinso:
part of #441 

- add gene viewer schema to models and reset migrations so there are only two versions, before and after the gene viewer


kevinlul:
There's a few details we could work on in the scope of this. The first is what we discussed yesterday: if any constraints should be applied to the `Gene` table. 

Other stuff is just general clean up: I think it would work out best if we reset the Alembic migrations again so that we only have two: one is the "initial", pre-variant viewer, and the second is everything variant viewer, including this pull request. The initial migration should also include these check constraints that seem to have been accidentally lost as they have to be manually added to the file after it's generated by Flask-Migrate https://github.com/ccmbioinfo/stager/blob/30fd2baa95e1f8c78cc52651754471e0c7935fa5/flask/migrations/versions/0e8799c17f5a_adding_month_of_birth_field_with_check.py#L22 https://github.com/ccmbioinfo/stager/blob/30fd2baa95e1f8c78cc52651754471e0c7935fa5/flask/migrations/versions/567a9f6ac119_adding_constraints_to_groupname.py#L21-L23

flask/app/models.py
```diff
@@ -252,7 +254,7 @@ class Dataset(db.Model):
     )
     notes: str = db.Column(db.Text)
     condition: DatasetCondition = db.Column(db.Enum(DatasetCondition), nullable=False)
-    extraction_protocol: str = db.Column(db.String(100))
+    extraction_protocol: str = db.Column(db.String(150))
```
kevinlul, line 257: Where did this change come from?

flask/app/models.py
```diff
@@ -421,25 +424,31 @@ class Variation(str, Enum):
...
-    Stop_Retained_Variant = "Stop_Retained_Variant"
 
 
+@dataclass
 class Gene(db.Model):
     gene_id: int = db.Column(db.Integer, primary_key=True)
-    hgnc_gene_id: int = db.Column(db.Integer)
+    hgnc_gene_id: int = db.Column(db.Integer, unique=True)
     ensembl_id: int = db.Column(db.Integer)
+    gene: str = db.Column(db.String(50))
```
kevinlul, line 436: What does this represent now?

delvinso: The `gene` column in the reports which are not necessarily hgnc gene names (unfortunately)

flask/app/models.py
delvinso: I thought I saw it was a length of 150 at some point in the gene pull migration. I believe it was carried over by setting `compare_type = True` in `context.configure()`. Anyways I can reset it to a length of 100?

flask/app/models.py
delvinso: after giving it some thought, here are a few options we can take to have a more meaningful ‘gene’ column (since it is the goal of this feature after all)
- query the variant position against an external resource to obtain the hgnc gene name/ensembl id/etc. unfortunately i think is easier said than done unless someone knows of a resource where all this information is single handedly kept? it's also a bit roundabout since i believe this is what was done in the first place to generate the report
- take the existing gene column and attempt to map it back to the same identifiers above (hgnc gene name/id/ensembl id, etc). also easier said than done 
- leave the gene column and assume it is a single source of truth since it was generated from the reports

thoughts?

kevinlul: Should the ensembl_id also be unique?

flask/app/models.py
```diff
@@ -421,25 +424,31 @@ class Variation(str, Enum):
...
     hgnc_gene_name: str = db.Column(db.String(50))
+    variant = db.relationship("Variant", back_populates="gene")
```
kevinlul, line 438: How is this relationship supposed work?

flask/app/models.py
```diff
@@ -450,17 +459,30 @@ class Variant(db.Model):
...
-    variant = db.relationship("Variant", backref="analysis")
+    dataset_id: int = db.Column(
+        db.Integer,
+        db.ForeignKey("dataset.dataset_id"),
+        primary_key=True,
+    )
+
+    variants = db.relationship("Variant", backref="variant_dataset_analysis")
+    analyses = db.relationship("Analysis", backref="variant_dataset_analysis")
+    datasets = db.relationship("Dataset", backref="variant_dataset_analysis")
```
kevinlul, line 484: I'm confused about the naming of these relationships and also they're considering the foreign keys independently rather than in conjunction.

flask/app/models.py
delvinso: it is suppose to be a one-to-one relationship between gene_id and variant_id but it looks like i'm missing the `useList = False` argument making it currently a one-to-many

flask/app/models.py
delvinso: i used the documentation for reference on how to create an association table - https://docs.sqlalchemy.org/en/13/orm/basic_relationships.html#association-object, hopefully that clears up the relationships?

as for the constraint i didn't explicitly add one as the three primary keys are considered a composite primary key (see lines 105 to 106 in the alembic migration)

kevinlul: Did you have to write the composite constraint in the migration manually or was Alembic able to figure that out automatically?

kevinlul:
Can you link the schematic to compare? I'm unsure if some of the relationship names are correct.

flask/app/models.py
delvinso: Alembic figured it out automatically

delvinso:
> Can you link the schematic to compare? I'm unsure if some of the relationship names are correct.

sure! https://dbdiagram.io/d/6061f962ecb54e10c33dcb85

flask/app/models.py
```diff
@@ -450,17 +464,29 @@ class Variant(db.Model):
...
-    variant = db.relationship("Variant", backref="analysis")
+    dataset_id: int = db.Column(
+        db.Integer,
+        db.ForeignKey("dataset.dataset_id"),
+        primary_key=True,
+    )
+
+    variants = db.relationship("Variant", backref="genotype")
+    analyses = db.relationship("Analysis", backref="genotype")
+    datasets = db.relationship("Dataset", backref="genotype")
```
kevinlul, line 488: Examining the schematic, these should all be singular. I don't know if the backref really makes sense as they will be aggregations, e.g.:

the Variant backref would have the zygosities for all datasets with this variant in this analysis
the Analysis backref would have _all_ the zygosities computed for all variants found for all datasets in this analysis
the Dataset backref would have all zygosities across all variants found in all analyses including this variant

They might be useful, since they won't be eager loaded anyway, for performing specific queries across these relationships though.

flask/app/models.py
```diff
@@ -421,25 +424,36 @@ class Variation(str, Enum):
     Stop_Lost = "Stop_Lost"
```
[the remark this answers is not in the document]
delvinso: Good catch! I'd say it's pretty redundant but they were taken verbatim from the reports. I can remove `_Variant` from both the enum names and the values.

flask/app/models.py
```diff
@@ -421,25 +424,36 @@ class Variation(str, Enum):
...
+@dataclass
 class Gene(db.Model):
     gene_id: int = db.Column(db.Integer, primary_key=True)
-    hgnc_gene_id: int = db.Column(db.Integer)
-    ensembl_id: int = db.Column(db.Integer)
+    hgnc_gene_id: int = db.Column(db.Integer, unique=True)
+    ensembl_id: int = db.Column(db.Integer, unique=True)
+    gene: str = db.Column(db.String(50))
     hgnc_gene_name: str = db.Column(db.String(50))
+    variant = db.relationship("Variant", backref="gene")
```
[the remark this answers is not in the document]
delvinso: Might be misunderstanding something but I don't think it would be possible to have the relationship on the variant (child) table? I'm looking at the documentation and it states that for a one-to-many relationship the relationship must be created through the parent

flask/app/models.py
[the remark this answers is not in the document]
kevinlul: If they're verbatim from the reports, I assume they're fairly static unless we have a pipeline update that detects more types of variants, but a string field may also be valid to avoid a database migration when that happens. If `_Variant` is verbatim then it's easier to keep it instead of having to remap.

flask/app/models.py
delvinso: Looking into your comment below re: redundancy of the `variant_id` unique constraint it looks like Alembic figured out the composite primary key, but I must have manually added the unique constraint (which would also be redundant I believe).

kevinlul:
Screen Shot 2021-04-07 at 11 38 18

EER diagram from MySQLWorkbench. The composite foreign key constraints don't seem to be in effect for Genotype, and it could as a result have a semantically invalid combination of dataset_id, analysis_id, and variant_id.

delvinso:
> Screen Shot 2021-04-07 at 11 38 18
> 
> EER diagram from MySQLWorkbench. The composite foreign key constraints don't seem to be in effect for Genotype, and it could as a result have a semantically invalid combination of dataset_id, analysis_id, and variant_id.

I think in this case since the three are a composite primary key (meaning there is an implied unique constraint) we shouldn't have to worry about it. AFAIK it is not possible to explicitly define a composite foreign key through remote tables.

kevinlul (commented):
My concern isn't the uniqueness in the table, it's that the structure will allow specifying a combination of dataset_id + analysis_id that doesn't make sense, since it's not a composite foreign key to the datasets_analyses table. Similar reasoning for variant_id + analysis_id, which should be a composite foreign key to the Variant table. I'm not sure if this exact spec is possible due to the overlap, but SQLAlchemy does support [composite foreign keys](https://docs.sqlalchemy.org/en/13/core/constraints.html)

delvinso:
Ah that's my bad! I've looked at the documentation before, which has a composite foreign key from a single table as an example, but it doesn't seem like one stemming from multiple tables is possible?

delvinso:
how does this look? 
Untitled_-_MySQL_Workbench

kevinlul (approved):
🚀

PR #465, opened 2021-03-26
clear non-static query cache on logout

cklamann:
Remove queries from client on logout, preserving constants like enums.


frewmack (approved):
Works for me! 🚀

PR #473, opened 2021-03-29
define inline functions and components to avoid rerenders

cklamann:
It seems that the MaterialTable instance in `DatasetTable.tsx` was rerendering when `api/unlinked` was fetched in the background, causing filters on the table to reset. The table was rerendering b/c some of its props were inline functions which, according to react's equivalency checks, are technically 'new' on every render. Defining these functions outside of the component seems to have fixed the issue on my tests, but confirmation is needed.


frewmack (approved):
Seems to work on my end. Compared it with master and didn't have any issues with filters being cleared after refetching.

PR #478, opened 2021-03-31
use tableRef to update dataset type filter on dataset page

cklamann:
Changing `defaultFilter` prop on column config causes a rerender of dataset table, wiping out filters that are already set. By using a ref to get the table instance, we can make the filter change imperatively and then force a state change by calling `onQueryChange`.

### DIFF
--- a/react/src/Datasets/components/DatasetTable.tsx
+++ b/react/src/Datasets/components/DatasetTable.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo, useState } from "react";
+import React, { useMemo, useRef, useState } from "react";
 import { useParams } from "react-router-dom";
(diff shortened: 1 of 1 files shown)


frewmack (commented):
The chip buttons no longer clear other filters on my end. However, it doesn't seem to refetch the data with the new dataset_type filters, but instead it's applying the new filters locally to the data that was fetched previously. Is there a way to force a refetch with the new filters?

cklamann:
Thanks, @frewmack. The filters were getting updated but not the query. Should be working better after the latest commit.

frewmack (approved):
Looks good to me!

kevinlul (approved):
sorcery

PR #484, opened 2021-04-01
Flask: add validate_json decorator to replace request.json checks

cklamann:
Move json request type checks into a decorator that can be added to any route expecting a `Content-Type: application/json` header. 
- This decorator is actually a function that returns a decorator, and it takes an optional `Validator` argument (currently not implemented, but in the future might be an instance of something like [this](https://docs.python-cerberus.org/en/stable/validation-rules.html)?). I added this feature b/c it seems reasonable to validate the json body in the same place that we confirm it's actually json. With this pattern, validators can be built and tested separately from the controllers (and decorator) and easily reused. 
- Since the decorator allows us to test the json-content validation logic in one place (`test_validation.py`), I removed the json content type checks from other tests. Routes decorated with `@validate_json` will now fail without the correct header, so I subclassed `FlaskClient` and added 2 convenience methods, `patch_json` and `post_json` that will add the headers 'automatically' to reduce boilerplate.    
- Maybe this is overkill?


flask/app/routes.py
```diff
@@ -83,9 +83,9 @@ def login():
 @routes.route("/api/logout", methods=["POST"])
 @login_required
 def logout():
-    if not request.json:
-        app.logger.error("Logout failed as request body is not JSON")
-        abort(400, description="Request body must be JSON!")
+    if request.headers.get("Content-Type") != "application/json":
+        app.logger.error("Logout failed as Content-Type is not JSON")
+        abort(400, description="Content-Type must be JSON!")
```
kevinlul, line 88: I think we can update this status code to 415 to be consistent with the rest.

flask/tests/conftest.py
```diff
@@ -37,6 +38,26 @@ def application():
...
+class TestClient(FlaskClient):
+    """ Class that adds 2 convenience methods for testing routes constrained by json content-type """
+
+    def patch_json(self, *args, **kwargs):
+        """ add content-type: json to patch request """
+        return super().patch(*args, **add_json_content_header(kwargs))
+
+    def post_json(self, *args, **kwargs):
+        """ add content-type: json to post request """
+        return super().post(*args, **add_json_content_header(kwargs))
```
kevinlul, line 58: I think this might be extra as the content type is already set if the `json=` named parameter is passed to the `post` and `patch` methods.

flask/tests/test_validation.py
```diff
@@ -0,0 +1,32 @@
+""" test validators """
+from unittest.mock import patch
+from unittest import TestCase
+from werkzeug.exceptions import UnsupportedMediaType
+from app.decorators import validate_json
+
+
+class ValidationTest(TestCase):
```
kevinlul, line 8: I like where this class is going! As an aside, I'm not sure if it's possible or really that good of a practice to mock out the ORM or MinIO over full integration tests though. I have some concerns about how long the test suites take now and they're not really do much yet.

flask/app/decorators.py
```diff
@@ -0,0 +1,21 @@
...
+
+def validate_json(Validator: Any = None):
+    """ HOF that returns a decorator with optional Validator object (currently not implemented) """
+
+    def wrapper(f):
+        @wraps(f)
+        def validate(*args, **kwargs):
+            """ validate content-type header and run optional Validator on input """
+            if request.headers.get("Content-Type") != "application/json":
+                error("Content-Type is not application/json")
```
kevinlul, line 15: Does the standard logging library work the same as the logger in the Flask app instance?

kevinlul (commented):
I think using a decorator is a great idea! I don't believe you need to use parentheses if no arguments are passed to the decorator. Our other decorators are in `utils`, so the decorator should probably go there unless you think it makes sense to separate them.

flask/app/routes.py
cklamann: updated in [1412c73](https://github.com/ccmbioinfo/stager/pull/484/commits/1412c73c9bf5de331aa6ccac700c43f117706816#diff-5ccfa28a10586b56ab898a8aac81bd2a24552a0de717b1a1bc982205f6be0c37R88).

flask/tests/conftest.py
cklamann: Gotcha. I removed the subclass and gave the tests in `test_users.py` without a `json=` explicit headers (e.g., [here](https://github.com/ccmbioinfo/stager/pull/484/commits/2dde857477321fde2f75d30373db9fa28de5cf0a#diff-ff159d3d293a3d1e6547afd4f56d2941f16ed154f21b86e61b5b5be88c48c285L158)).

flask/app/decorators.py
cklamann: Evidently not. Seems like we'd need to configure a root logger first (something like [this](https://flask.palletsprojects.com/en/1.1.x/logging/#basic-configuration)). I was concerned about logging having a dependency on the app, since it seemed like overkill to spin up the app context just to unit test a function, but I ended up just removing the logging code from the decorator altogether, since the werkzeug logger will print the `415`.

flask/app/routes.py
```diff
@@ -83,9 +83,9 @@ def login():
 @routes.route("/api/logout", methods=["POST"])
 @login_required
 def logout():
-    if not request.json:
-        app.logger.error("Logout failed as request body is not JSON")
-        abort(400, description="Request body must be JSON!")
+    if request.headers.get("Content-Type") != "application/json":
+        app.logger.error("Logout failed as Content-Type is not JSON")
+        abort(415, description="Content-Type must be JSON!")
```
kevinlul, line 88: What's the reason for not using the decorator here?

flask/tests/conftest.py
```diff
@@ -37,6 +38,14 @@ def application():
...
 
 
+def add_json_content_header(kwargs: dict):
+    """helper method for injecting content-type: json into headers"""
+    headers = kwargs.pop("headers", {})
+    headers["Content-Type"] = "application/json"
+    kwargs["headers"] = headers
+    return kwargs
+
+
```
kevinlul, line 48: I think this is no longer used now as well

kevinlul (changes requested):

flask/app/routes.py
cklamann: oops! fixed in  [75f8c38](https://github.com/ccmbioinfo/stager/pull/484/commits/75f8c38e245b8d2ffed64bd849c9df2c948638af)

flask/tests/conftest.py
cklamann: fixed in [75f8c38](https://github.com/ccmbioinfo/stager/pull/484/commits/75f8c38e245b8d2ffed64bd849c9df2c948638af).

kevinlul (approved):
LGTM

PR #490, opened 2021-04-05
React: limit notification list to 50 most recent

cklamann:


kevinlul (commented):
The endpoint also makes it possible to limit the amount of data returned. Could that be done?

kevinlul (approved):
LGTM

PR #495, opened 2021-04-08
Flask: pip-compile --upgrade

kevinlul:
pymysql[rsa] 1.0.2 is required for the database seed to work correctly. The rest seems to be a shortcoming of dependabot, maybe related to dependabot/dependabot-core#3447. If this continues to be an issue we can create a workflow to do this on a sprint cycle for pip and yarn instead.


delvinso (approved):
Looks good to me

Issue #483, opened 2021-04-01
Variant viewer frontend


kevinlul:
Title: search variants
Icon: magnifying glass
Layout: just a search bar for genes

See [original Sample Tracker](https://sampletracker.ccm.sickkids.ca/GeneReports)

https://github.com/ccmbioinfo/sample-tracker/blob/master/static/js/GeneReports.jsx

kevinlul:
![image](https://user-images.githubusercontent.com/6320810/113893542-21ec7d00-9795-11eb-98cf-8a2378d69c0e.png)

![image](https://user-images.githubusercontent.com/6320810/113893591-2b75e500-9795-11eb-9947-2ab33ba4fd28.png)

![image](https://user-images.githubusercontent.com/6320810/113893627-33ce2000-9795-11eb-8c94-50a62a0e3879.png)

![image](https://user-images.githubusercontent.com/6320810/113893664-3b8dc480-9795-11eb-9d37-afa88f82f690.png)

kevinlul:
Compare with https://sgc.garvan.org.au/ (suggested by Hillary), old version of its Angular source here https://github.com/shusson/sgc

Searching by position is not presently in-scope

kevinlul:
Last item is to add the participant-wise report download option to frontend

kevinlul:
Closed in #567. Future changes related to #542 will be considered separately.

Issue #480, opened 2021-04-01
Variant viewer: individual GET /api/summary/genes/...


kevinlul:
- [x] GET /api/summary/genes/:hgnc_gene_name 
       Returns a JSON gene object for the specified gene, or 404 otherwise. 
- [x] GET /api/summary/genes/hgnc/:id 
       Returns a JSON gene object for the specified gene based on its HGNC ID, or 404 otherwise. 
- [x] GET /api/summary/genes/ensg/:id 
       Returns a JSON gene object for the specified gene based on its Ensembl ID, or 404 otherwise.

kevinlul:
These were actually done in #502, they just may be reorganized later based on what identifiers we actually have.

PR #502, opened 2021-04-13
Create gene search autocomplete component and summary gene endpoints

### LINKED ISSUE #479: Variant viewer: GET /api/summary/genes

Responds with an array of genes, i.e., the meaningful columns of the Gene database model. The default response should be very cacheable assuming the database is seeded with a large enough quantity of analysis reports. 

The 2020-03-03 data frame contains almost 25000 genes or variants, so we additionally support INSTR queries on _hgnc_gene_name_ via an optional query string parameter of the same name. Due to the potentially large volume of data returned here, we may need to implement paging like in our other list endpoints.

cklamann:
- closes #479, #480, #483
* create mui autocomplete component for searching on hgnc gene name
  -  currently doing nothing with selection, but once `summary/variants` endpoint is operational, will fetch and download the variant csv based on selected gene
* create endpoints `[GET] api/summary/genes`, `[GET] api/summary/genes/:hgnc_gene_name`, `[GET] api/summary/genes/hgnc/:id`, `[GET] api/summary/genes/ensg/:id` and integration tests


flask/app/genes.py
```diff
@@ -0,0 +1,104 @@
...
+genes_blueprint = Blueprint(
+    "genes",
+    __name__,
+)
+
+
+@genes_blueprint.route("/api/summary/genes", methods=["GET"])
+@paged
+@login_required
+def genes(page, limit):
```
kevinlul, line 20: add type hints here

flask/app/genes.py
```diff
@@ -0,0 +1,104 @@
...
+    gene_query = db.session.query(Gene)
+    if search:
+        gene_query = gene_query.filter(func.instr(Gene.hgnc_gene_name, search))
+
+    total_count = gene_query.with_entities(
+        func.count(func.distinct(Gene.gene_id))
+    ).scalar()
+    if limit and page:
+        gene_query.limit(limit).offset(page * limit)
+    results = gene_query.all()
```
kevinlul, line 32: You can use the established one-line syntax `.limit(limit).offset(page * (limit or 0)).all()`.

Maybe I should leave in a comment why it works:
- `limit` is either positive or None. `.limit(None)` is a safe no-op
- `page` is non-negative. Requesting page 0 with no limit is also the same as without

flask/app/genes.py
```diff
@@ -0,0 +1,104 @@
...
+        writer = DictWriter(
+            csv_data,
+            fieldnames=colnames,
+            quoting=QUOTE_MINIMAL,
+        )
+        writer.writeheader()
+        for row in results:
+            dct = row.__dict__
+            dct.pop("_sa_instance_state")
+            writer.writerow(dct)
```
kevinlul, line 56: Use `asdict` from the dataclasses module instead

flask/app/genes.py
```diff
@@ -0,0 +1,104 @@
...
+
+@genes_blueprint.route("/api/summary/genes/<string:hgnc_gene_name>", methods=["GET"])
+@login_required
+def get_gene_hgnc_name(hgnc_gene_name):
+    """ return a gene based on ghnc gene name """
+    gene_query = db.session.query(Gene)
+    result = gene_query.filter(Gene.hgnc_gene_name == hgnc_gene_name).first()
+
+    if not result:
+        abort(404, f"No gene found with hgnc name {hgnc_name}")
```
kevinlul, line 76: Under flask_sqlalchemy, `.first_or_404` is a method for this exact use case.

flask/app/genes.py
```diff
@@ -0,0 +1,104 @@
...
+@login_required
+def get_gene_hgnc_id(hgnc_id):
+    """ return a gene based on ghnc gene id """
+    gene_query = db.session.query(Gene)
+    result = gene_query.filter(Gene.hgnc_gene_id == hgnc_id).first()
+
+    if not result:
+        abort(404, f"No gene found with hgnc_id {hgnc_id}")
+
+    return jsonify(result)
```
kevinlul, line 91: e.g. in one line since we have no real logging requirement
```suggestion
    return jsonify(Gene.query.filter_by(hgnc_gene_id = hgnc_id).first_or_404())
```

flask/app/genes.py
```diff
@@ -0,0 +1,104 @@
...
+        response.headers["Content-Type"] = "text/csv"
+
+        return response
+
+    abort(415, "Mime-Type must be text/csv or application/json!")
+
+
+@genes_blueprint.route("/api/summary/genes/<string:hgnc_gene_name>", methods=["GET"])
+@login_required
+def get_gene_hgnc_name(hgnc_gene_name):
```
kevinlul, line 70: add type annotation to match

kevinlul (changes requested):

react/src/typings.tsx
```diff
@@ -250,3 +250,29 @@ export interface Option {
...
+    alt_allele: string;
+    variation: string;
+    refseq_change: string;
+    depth: number;
+    gene_id: number;
+    conserved_in_20_mammals: number;
+    sift_score: number;
+    polyphen_score: number;
+    cadd_score: number;
+    gnomad_af: number;
```
kevinlul, line 277: These should be optional.

react/src/typings.tsx
```diff
@@ -250,3 +250,29 @@ export interface Option {
...
+}
+
+export interface Variant {
+    variant_id: number;
+    analysis_id: number;
+    position: string;
+    reference_allele: string;
+    alt_allele: string;
+    variation: string;
+    refseq_change: string;
```
kevinlul, line 270:
```suggestion
    refseq_change?: string;
```

react/src/typings.tsx
```diff
@@ -250,3 +250,29 @@ export interface Option {
...
     selected?: boolean;
 }
+
+export interface Gene {
+    gene_id: number;
+    hgnc_gene_id: number;
+    ensembl_id: number;
+    gene: string;
+    hgnc_gene_name: string;
+    variants?: Variant[];
```
kevinlul, line 260: These are all nullable but we might be changing this a bit given the discussion today.

react/src/Navigation.tsx
```diff
@@ -189,6 +191,12 @@ const routes: RouteItem[] = [
         icon: <SupervisedUserCircle />,
         requiresAdmin: true,
     },
+    {
+        pageName: "Search Variants",
+        path: "/search-variants",
+        main: SearchVariants,
+        icon: <SearchIcon />,
+    },
```
kevinlul, line 199: Let's move this below Analyses in the sidebar and shorten the label to `Variants`. Shorten the path to match too.

react/src/SearchVariants/components/Autocomplete.tsx
```diff
@@ -0,0 +1,73 @@
+import React, { useState } from "react";
+import { Autocomplete } from "@material-ui/lab";
+import CircularProgress from "@material-ui/core/CircularProgress";
+import { useGenesQuery } from "./../../hooks/genes";
+import { OutlinedInput, useTheme } from "@material-ui/core";
+import { Search } from "@material-ui/icons";
```
kevinlul, line 6: Import `node_modules/` dependencies first in alphabetical order, then our own code. If importing from mui core, use the destructured import. For the first-party import, `./` isn't needed since it's in a parent directory.

react/src/hooks/utils.tsx
```diff
@@ -13,24 +13,54 @@ import {
...
+        const filename = (response.headers.get("content-disposition") || "foo=report.csv").replace(
+            /.+=/,
+            ""
+        );
+        const url = URL.createObjectURL(new Blob(["\ufeff", csv]));
+        downloadLink.href = url;
+        downloadLink.download = filename;
+        document.body.appendChild(downloadLink);
+        downloadLink.click();
+        document.body.removeChild(downloadLink);
```
kevinlul, line 63: I was hoping that this approach wouldn't be needed if we already have a content-disposition header on the backend. `window.open` might get blocked by a pop-up blocker, but would directly navigating to the endpoint work?

react/src/hooks/utils.tsx
```diff
@@ -13,24 +13,54 @@ import {
...
+    const paramString = Object.keys(params).length ? `?${new URLSearchParams(params)}` : "";
+    const response = await fetch(`${url}${paramString}`, { ...options, headers });
+    if (response.ok) {
+        const csv = await response.text();
+        const downloadLink = document.createElement("a");
+        const filename = (response.headers.get("content-disposition") || "foo=report.csv").replace(
+            /.+=/,
+            ""
+        );
+        const url = URL.createObjectURL(new Blob(["\ufeff", csv]));
```
kevinlul, line 58: Do we really need to insert a BOM?

kevinlul (changes requested):

flask/app/genes.py
delvinso, line 32: if a gene does not have a hit in the database the response is : 
```
{
  "data": [], 
  "page": 0, 
  "total_count": 0
}
```
should there be a check for whether a result is returned from the database? eg. if len(results) == 0 then abort.

flask/app/genes.py
cklamann: updated in [616323a](https://github.com/ccmbioinfo/stager/commit/616323a7b787eef2dd3f960032ee95a1cf415d5f#diff-74bfb2f82124f4d2a0aeb38cdff367fb949bf5e0542ceb4dbebe22e7241caf72R20)

flask/app/genes.py
cklamann: updated in [616323a](https://github.com/ccmbioinfo/stager/commit/616323a7b787eef2dd3f960032ee95a1cf415d5f#diff-74bfb2f82124f4d2a0aeb38cdff367fb949bf5e0542ceb4dbebe22e7241caf72R31)

flask/app/genes.py
cklamann: updated in [616323a](https://github.com/ccmbioinfo/stager/commit/616323a7b787eef2dd3f960032ee95a1cf415d5f#diff-74bfb2f82124f4d2a0aeb38cdff367fb949bf5e0542ceb4dbebe22e7241caf72R53)

flask/app/genes.py
cklamann: updated in [616323a](https://github.com/ccmbioinfo/stager/commit/616323a7b787eef2dd3f960032ee95a1cf415d5f)

flask/app/genes.py
cklamann: updated in [616323a](https://github.com/ccmbioinfo/stager/commit/616323a7b787eef2dd3f960032ee95a1cf415d5f)

flask/app/genes.py
cklamann: updated in [616323a](https://github.com/ccmbioinfo/stager/commit/616323a7b787eef2dd3f960032ee95a1cf415d5f)

react/src/typings.tsx
cklamann: updated in [bec4861](https://github.com/ccmbioinfo/stager/commit/bec48617015303e9dc4870e8a56d498c8052d646)

react/src/typings.tsx
cklamann: updated in [bec4861](https://github.com/ccmbioinfo/stager/commit/bec48617015303e9dc4870e8a56d498c8052d646)

react/src/typings.tsx
cklamann: updated in [bec4861](https://github.com/ccmbioinfo/stager/commit/bec48617015303e9dc4870e8a56d498c8052d646)

react/src/SearchVariants/components/Autocomplete.tsx
cklamann: updated in [5195119](https://github.com/ccmbioinfo/stager/commit/51951199963d68492f614fc067aa268ed2dca411)

react/src/Navigation.tsx
cklamann: updated in [e75eaec](https://github.com/ccmbioinfo/stager/commit/e75eaec8e72fd65b311c9be77b5887c0a58c4988)

react/src/hooks/utils.tsx
cklamann: update in [bc01cf3](https://github.com/ccmbioinfo/stager/compare/cklamann/variant-viewer-fe-and-route)

react/src/hooks/utils.tsx
cklamann: I don't think so -- afaik if we're making an async fetch the browser won't pay attention to the content-disposition header. We could navigate the user to the url if the url held a static file (e.g., the endpoint doesn't return the csv but saves it somewhere publicly accessible and returns the path and then we point the browser there), but if the file comes from an api route, redirecting the user there will just reload the app and the router will take over.

flask/app/genes.py
cklamann: I don't think so. Generally I think that the reasoning is that the client is asking the server for the results of a search against an index route, and the server is (in this case at least) obliged to send back a json representation of the results. In the case of no matches, the results are an empty json array. A 400 implies that the client has made an error, but I wouldn't construe an empty search as an error -- maybe an empty result set is informative or even exactly what they were looking for! A 400 (or 404) might be confusing in this case, making the client think they sent a bad param or tried a non-existent endpoint. This is different than, say, asking for a specific value, like `[GET] api/foo/2` where `2` is the id of some model `foo`. If the user asks for a model that doesn't exist, a 404 might be a good response (some frameworks return 404 automatically here). If there's persuasive documentation suggesting an error is is appropriate for an empty result set on an index route search, I can certainly be persuaded, though I haven't seen this implementation much in the wild :slightly_smiling_face: .

delvinso: i appreciate the detailed answer - that makes sense to me!

react/src/hooks/utils.tsx
```diff
@@ -13,24 +13,53 @@ import {
...
+
+    const response = await fetch(`${url}${paramString}`, { ...options, headers });
+
+    if (response.ok) {
+        const csvBlob = await response.blob();
+        const downloadLink = document.createElement("a");
+        const filename = (
+            response.headers.get("content-disposition") || "filename=report.csv"
+        ).replace(/.+=/, "");
+        const url = URL.createObjectURL(csvBlob);
```
kevinlul, line 59: If using createObjectURL, we need to free it after the click with https://developer.mozilla.org/en-US/docs/Web/API/URL/revokeObjectURL since this is an SPA.

react/src/hooks/utils.tsx
kevinlul: Is it because this is used in a react-query hook rather than just being a standalone side effect function? Or is the problem that we're distinguishing the return type based on the Accept header rather than something in the URL, which would allow it to be used as a link?

cklamann: The latter, I think. Explanation can be found [here](https://create-react-app.dev/docs/proxying-api-requests-in-development/) in terms of current setup. If we point the browser at, say, `stager.local/api/download-some-file`, the server will see that the browser request's Accept header lists text/html before */* and respond with index.html, and then react will navigate to `/api/download-some-file`, which will display whatever view the router defaults to rendering at that path.

flask/app/genes.py
```diff
@@ -51,8 +50,8 @@ def genes(page, limit):
         )
         writer.writeheader()
         for row in results:
-            dct = row.__dict__
-            dct.pop("_sa_instance_state")
+            dct = asdict(row)
+            dct.pop("_sa_instance_state", None)
```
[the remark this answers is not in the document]
cklamann: updated in [83a6b06](https://github.com/ccmbioinfo/stager/commit/83a6b06f4c5b16f4a1fd95d7e2fa9ffae08afe88)

react/src/hooks/utils.tsx
cklamann: updated in [e84858e](https://github.com/ccmbioinfo/stager/commit/e84858e223e49af440be41c032199826052c68da)

react/src/hooks/utils.tsx
kevinlul: I see, the problem is that it wouldn't work against the development server, though it would work in production. This never came up as we don't open the endpoints in the browser, which would send `Accept: text/html; other stuff; */*`. `fetch` and `curl` both use `*/*`.

kevinlul (approved):
LGTM! Please resolve the merge conflict that resulted from merging Delvin's PR first (you don't have to worry about combining the files here.)

cklamann:
Conflict resolved in [61b2a8c](https://github.com/ccmbioinfo/stager/pull/502/commits/61b2a8c51ba1bb14ad9faa0e40e7c9a0a8fdbd54).

PR #506, opened 2021-04-14
Flask: /api/groups fix-up related to CodeQL warning

kevinlul:
There is no XSS risk as responses are JSON, but the location header was not escaped (not a problem specifically for group code as it has to be a valid domain name).

- Add type hints and docstrings to all endpoints
- Change group not found error code, regardless of authorization, to 404
- Validate the typing of the users array for the create and update endpoints
- Update a query's syntax for SQLAlchemy 1.4
- Escape the group code in the location header anyway
- Log some exceptions

### DIFF
--- a/flask/app/groups.py
+++ b/flask/app/groups.py
@@ -86,13 +86,24 @@ def get_group(group_code) -> Response:
 @validate_json
 def update_group(group_code) -> Response:
     """
-    Change the display name or user membership of a group. Administrator-only.
+    Change the display name or user membership of a group. Both request body fields
+    are optional and they will only be changed if provided. Administrator-only.
 
     Potential problems: MinIO operations failing result in an inconsistent state.
     """
+    strlist_users = (
+        request.json.get("users")  # Ignores the empty list
+        and type(request.json["users"]) is list
+        and all([type(user) is str for user in request.json["users"]])
+    )
+    if type(request.json) is not dict:
+        abort(400, description="Expected object")
+    if "users" in strlist_users and not strlist_users:
+        abort(400, description="users should be a string array")
+
     group = models.Group.query.filter_by(group_code=group_code).first_or_404()
 
-    if request.json.get("group_name"):
+    if request.json.get("group_name", type=str):
         # Check if display name is in use, 422
         conflicting_group = models.Group.query.filter_by(
             group_name=request.json["group_name"]
@@ -108,11 +119,7 @@ def update_group(group_code) -> Response:
     )
 
     # Check if a user to be added doesn't exist, 404
-    if (
-        "users" in request.json
-        and type(request.json["users"]) is list
-        and all([type(user) is str for user in request.json["users"]])
-    ):
+    if strlist_users:
         users = models.User.query.filter(
             models.User.username.in_(request.json["users"])
         ).all()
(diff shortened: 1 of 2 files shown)


flask/app/groups.py
```diff
@@ -99,7 +108,11 @@ def update_group(group_code):
     )
 
     # Check if a user to be added doesn't exist, 404
-    if "users" in request.json:
+    if (
+        "users" in request.json
+        and type(request.json["users"]) is list
+        and all([type(user) is str for user in request.json["users"]])
+    ):
```
delvinso, line 115: the group name update occurs regardless of whether this check succeeds or not - is this intended behaviour?

flask/app/groups.py
```diff
@@ -170,43 +187,52 @@ def create_group():
     except:
         abort(422, description="Invalid bucket name")
 
-    group_obj = models.Group(group_code=group_code, group_name=group_name)
+    group = models.Group(group_code=group_code, group_name=group_name)
 
     # Add users to the db if they were given
-    if request.json.get("users"):
+    have_users = (
```
delvinso, line 193: similar to the above, it seems like if the `users` field is passed in with the intention to add users to the group but the check fails, there is no message informing that the users have not been added to the group

flask/app/groups.py
```diff
@@ -170,43 +187,52 @@ def create_group():
     except:
```
delvinso, line 187: playing around with this and if the db constraint fails on group_code it will throw a 422 here

flask/app/groups.py
```diff
@@ -121,20 +134,20 @@ def update_group(group_code):
...
 @login_required
 @check_admin
 @validate_json
 def create_group():
+    """
+    Create a new permission group. Administrator-only.
 
+    Potential problems: MinIO operations failing result in an inconsistent state.
+    """
     group_name = request.json.get("group_name")
```
delvinso, line 151: minor - the use of this endpoint seems to be for one group at a time but if an array is passed in, the query will throw a db error

flask/app/groups.py
kevinlul: This is intended. The `users` field is optional. If not provided, no change is made.

flask/app/groups.py
kevinlul: I'm guessing the error was `'list' object has no attribute 'get'` and I think all of our endpoints are affected due to lack of detailed input validation.

flask/app/groups.py
kevinlul: This updates the behaviour so we only attempt if `users` is of the correct format. Before it would have borked with a strange error while already creating the bucket, so the name wouldn't be usable without manually removing the bucket.

flask/app/groups.py
kevinlul: I'm sorry, I'm not sure what the problem is here. The database constraint on `group_code` is a valid domain name, which matches the S3 requirements for a bucket. We try to make the bucket first to be confident before committing it to MySQL.

flask/app/groups.py
delvinso: what i meant (similar to below) is that if `users` is supplied but not in the correct format (eg. not in a list) then the group name update will still occur and there is no indication that the supplied `users` field was invalid. i don't think it's an issue, i'm just wondering if it would be helpful to at least log some kind of message indicating there was an attempt to do so

delvinso (approved):
looks good to me!

flask/app/groups.py
kevinlul: I've accounted for this now too. Maybe another time we can look at the input validation problem in depth.

delvinso: nice - i agree, it would also help in reducing repetitive code across the endpoints

delvinso (approved):
looks good to me

PR #507, opened 2021-04-14
Epic: OAuth authentication support

frewmack:
Lots of changes in this PR so I've included a checklist of all the notable changes.

**Flask changes:**
  - [ ] Add `authlib` library for OAuth/OIDC support. Add `requests` library for interfacing with Keycloak Admin API.
  - [ ] Update User model to include `issuer`, `subject` fields for uniquely identifying a user with OpenID Connect.
  - [ ] Add GET /api/login, GET /api/authorize endpoints for handling login flow.
    - [ ] GET /api/login - Return the [Authorization Endpoint URL](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) for the OAuth Provider that Stager uses.
    - [ ] GET /api/authorize - Take in parameters from a successful authorization request, exchange the provided Authorization Code for ID and Access Tokens at the OAuth Provider's [Token Endpoint](https://openid.net/specs/openid-connect-core-1_0.html#TokenEndpoint), and logs in the current user if the token exchange is successful. Return same user details as POST /api/login.
  - [ ] Update POST /api/login endpoint to return 401 if OIDC support is enabled, unless the current user is already signed in.
  - [ ] Add Keycloak Admin API support in ~~`manage_oidc.py`~~ `manage_keycloak.py` for setting up Keycloak in `manage.py` when running the backend stack in Docker from `docker-compose.oidc.yaml`.
  - [ ] Update GET /api endpoint to include `oauth` field in response json, for frontend to know which login page to render.

**React changes:**
  - [ ] Change `Login.tsx` to include different forms depending on whether OAuth support is enabled or disabled.

**Docker / Environment changes:**
  - [ ] Add environment variables `ENABLE_OIDC`, `OIDC_CLIENT_ID`, `OIDC_CLIENT_SECRET`, `OIDC_WELL_KNOWN`, `OIDC_PROVIDER`, `REACT_APP_API_ENDPOINT` whose uses are explained in `docs/oidc.md`.
  - [ ] Add `docker-compose.oidc.yaml` for launching Stager backend stack with a dedicated Keycloak instance for testing OAuth endpoints.
    - [ ] Add custom entrypoint script (`init/01-database.sh`) for `mysql` for creating a database for Keycloak in the same MySQL container as for Stager, as opposed to having two separate MySQL containers.
  - [ ] Update `docker-compose.yaml` to include the above environment variables for Flask if they are provided. Running the backend stack with `docker-compose up --build` without setting `ENABLE_OIDC` in `.env` should be functionally identical to master.


docker-compose.oidc.yaml
```diff
@@ -0,0 +1,86 @@
...
+      ST_DATABASE_URI: "mysql+pymysql://${MYSQL_USER}:${MYSQL_PASSWORD}@mysql/${MYSQL_DATABASE}"
+      MINIO_ENDPOINT: minio:9000
+      MINIO_ACCESS_KEY: "${MINIO_ACCESS_KEY}"
+      MINIO_SECRET_KEY: "${MINIO_SECRET_KEY}"
+      ENABLE_OIDC: "true"
+      OIDC_CLIENT_ID: "ccm-stager"
+      OIDC_CLIENT_SECRET: "$$not_a_secret"
+      OIDC_WELL_KNOWN: "http://keycloak:8080/auth/realms/ccm/.well-known/openid-configuration"
+      OIDC_PROVIDER: "keycloak"
+      KEYCLOAK_HOST: "http://keycloak:8080"
```
cklamann, line 61: Might need to add KEYCLOAK_USER and PW here, as they are referenced in `manage_oidc.py` (I needed to add them manually to use that function outside of the initialization code).

flask/app/manage_oidc.py
```diff
@@ -0,0 +1,284 @@
...
+            if len(found_users) != 1:
+                app.logger.error(f"ERROR: Added user {user.username} not found ???")
+                return False
+
+            added_user = found_users[0]
+            print(f"Found user in Keycloak: {added_user}")
+            endpoints = fetch_admin_well_known("ccm")
+            issuer = endpoints["issuer"]
+            subject = added_user["id"]
+            user.set_oidc_fields(issuer, subject)
```
cklamann, line 241: I needed to commit this via a db session for the changes to take.

cklamann (commented):
I was able to add users to keycloak, toggle the authorization mode, log in, log out (though I had to destroy my session through the keycloak admin page), and login as a new user, all while seeing what I expected to see. I didn't exhaustively test every route, but everything seems to be functioning normally. :rocket:

flask/app/manage_oidc.py
frewmack: The issue here is that `add_keycloak_user` is used in seed functions in `manage.py`, which handles committing the db session. I can get around this by adding click commands for these actions that handle the db session logic separately.

kevinlul (commented):
Okay, I finally have this working in some form. How does signing out work? I signed out of Stager but I was still signed-in to Keycloak, so signing into Stager again would automatically sign in without an option to switch accounts.

frewmack:
> 
> 
> Okay, I finally have this working in some form. How does signing out work? I signed out of Stager but I was still signed-in to Keycloak, so signing into Stager again would automatically sign in without an option to switch accounts.

Right now, signing out only removes your user session from Flask. Keycloak manages your session separately. 

There is a revocation endpoint in the OAuth 2.0 standard for revoking tokens on logout, but I haven't implemented this yet. I can take a look in a future PR.

init/01-database.sh
```diff
@@ -0,0 +1,27 @@
...
+    docker_process_sql --database=mysql <<-EOSQL
+        CREATE DATABASE \`$1\` CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
+        GRANT ALL PRIVILEGES ON \`$1\`.* TO '$2'@'%';
+EOSQL
+}
+
+_create_user "${KEYCLOAK_DB_USER}" "${KEYCLOAK_DB_PASSWORD}"
+_create_database "${KEYCLOAK_DATABASE}" "${KEYCLOAK_DB_USER}"
+
+mysql_note "Finished initializing!"
```
kevinlul, line 27: Trailing newline

react/sample.env
```diff
@@ -0,0 +1,3 @@
+# frontend
+REACT_APP_NAME=Stager
+REACT_APP_API_ENDPOINT=http://localhost:5000
```
kevinlul, line 3: Trailing newline

react/src/Login.tsx
```diff
@@ -56,53 +138,132 @@ export default function LoginForm({
...
+        <BrowserRouter>
+            <Switch>
+                <Route path="/oidc_callback" render={() => OAuthCallback} />
+                <Route
+                    path="/"
+                    render={() => (
+                        <Box className={classes.root}>
+                            <Container maxWidth="sm">
+                                <Paper component="form" className={classes.form}>
+                                    {FormDisplay}
```
kevinlul, line 260: Shouldn't this just be a component or a ternary here?

kevinlul (commented):
I'm okay with merging this to mainline and continuing to work on it. Sorry for taking so long. Some merge conflicts.

react/src/Login.tsx
```diff
@@ -56,53 +138,132 @@ export default function LoginForm({
...
+        () => (props.oauth ? <OIDCRedirectHandler {...props} /> : <Redirect to="/" />),
+        [props]
+    );
+
+    const FormDisplay: JSX.Element = useMemo(() => {
+        if (props.oauth === true) {
+            return <OauthLoginForm />;
+        }
+        return <LoginForm {...props} />;
+    }, [props]);
```
kevinlul, line 248: I see, this isn't a proper functional component because it's actually useMemo. Is there a reason for using that over just the body of this as a component (or inlined below?)

frewmack: I figured memoizing it made sense since `props.oauth` should never change for the duration that React is being served. I can move this to inline instead.

kevinlul (approved):
🚀

PR #508, opened 2021-04-15
Flask: add test data and tests for variant summary

delvinso:
- adds test variant data to datasets 2 and 3, as part of analysis 2
- tests in test_gene.py pass, but delete tests (`test_delete_*`) are failing in other tests due to foreign key constraints/relationships


flask/tests/conftest.py
```diff
@@ -5,6 +5,8 @@
 from app.config import Config
 from app.models import *
 from flask.testing import FlaskClient
+import numpy as np
```
delvinso, line 8: will remove these unnecessary imports

delvinso (commented):
I think you hit the nail on the head - tests are fixed in 021a3c589ff34d57227950207ce0b5da46325aaf and 491d6bc4ad372d6880184f8069e0a24041b2c95b. One thing I’m not sure of (and did not include) is whether the analysis delete needs to be committed before the dataset delete since it seems like pre-variant schema this was handled successfully in a single transaction.

cklamann (approved):
LGTM!

kevinlul (approved):
This does show that we might have problems with using the admin-only delete endpoints in general once report data is added to the database without delete cascades, but that's out of scope for this PR.

PR #509, opened 2021-04-16
Flask: /api/users fix-up related to CodeQL warning

kevinlul:
There is no XSS risk as responses are JSON, but the location header was not escaped, and usernames may contain arbitrary characters. This fixes that so the resulting location header is always usable, and also addresses usernames with slashes.

- Add type hints and docstrings to all endpoints
- Refactor: move get_minio_admin to utils
- Log some exceptions
- Add a simple assertion for request body typing
- Fix create_user not returning the Location header for the actual conflicting user

### DIFF
--- a/flask/app/groups.py
+++ b/flask/app/groups.py
@@ -6,8 +6,8 @@ from minio import Minio
 
 from . import models
 from .extensions import db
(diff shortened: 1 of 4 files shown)


delvinso (approved):
LGTM

PR #510, opened 2021-04-16
React: CSV template download in Add Datasets

### LINKED ISSUE #504: Add a template CSV for add datasets
This may or may not be based on the columns currently chosen in the user interface.

frewmack:
Closes #504 
Adds a button that downloads the contents of the Data Entry Table in it's current state in .csv format.

### DIFF
--- a/react/src/AddDatasets/components/utils.tsx
+++ b/react/src/AddDatasets/components/utils.tsx
@@ -1,13 +1,4 @@
-import {
-    DataEntryHeader,
-    DataEntryRow,
-    DataEntryRowBase,
-    DataEntryRowOptional,
-    DataEntryRowRNASeq,
-    Family,
-    Option,
-    Participant,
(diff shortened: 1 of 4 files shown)


react/src/AddDatasets/components/utils.tsx
```diff
@@ -224,3 +224,38 @@ function snakeToTitle(str: string): string {
...
+            if (Array.isArray(value)) value = value.join("|");
+            else if (!value) value = "";
+            else if (typeof value !== "string") value = "" + value;
+            values.push(value as string);
+        }
+        csv += values.join(",") + "\n";
+    }
+
+    return csv;
+}
```
kevinlul, line 261: This method doesn't handle escaping, so you'll get funny outputs when the field to be serialized contains a comma.

kevinlul (approved):
LGTM!

PR #518, opened 2021-04-19
connect autocomplete to variant csv endpoint

cklamann:
* Connect variant/csv endpoint to front-end variant viewer
* return csv and automatically download when user selects gene from autocomplete list
* cache csv as blob/filename in browser via react-query
* handle 400 errors if no variants returned for selected gene

### DIFF
--- a/react/src/SearchVariants/components/SearchVariantsPage.tsx
+++ b/react/src/SearchVariants/components/SearchVariantsPage.tsx
@@ -23,6 +23,7 @@ const useStyles = makeStyles(theme => ({
 const SearchVariantsPage: React.FC<SearchVariantsPageProps> = () => {
     const [selectedGenes, setSelectedGenes] = useState<string[]>([]);
     const [error, setError] = useState(false);
+    const [csvFetchEnabled, setCsvFetchEnabled] = useState(false);
 
     const { enqueueSnackbar } = useSnackbar();
 
@@ -34,8 +35,8 @@ const SearchVariantsPage: React.FC<SearchVariantsPageProps> = () => {
         }
     };
 
-    const { data: blob, refetch } = useVariantsQuery({ panel: selectedGenes.join(";") }, "csv", {
-        enabled: false,
+    const { data: blob } = useVariantsQuery({ panel: selectedGenes.join(";") }, "csv", {
+        enabled: csvFetchEnabled,
         onError: response => {
             if (response.status === 400) {
                 setError(true);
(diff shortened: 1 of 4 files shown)


react/src/SearchVariants/components/SearchVariantsPage.tsx
```diff
@@ -21,11 +22,32 @@ const useStyles = makeStyles(theme => ({
...
-    /* eslint-disable @typescript-eslint/no-unused-vars */
-    //const { data: variants } = useVariantsQuery({ search: selectedGene }, "csv", !!selectedGene);
+    const { data: blob } = useVariantsQuery({ panel: selectedGene?.hgnc_gene_name }, "csv", {
+        enabled: !!selectedGene,
+        onError: e => {
+            //I'm not sure how common this scenario will be
+            if (e.status === 400) {
+                setError(true);
+            } else {
+                throw e;
```
frewmack, line 34: Maybe create a Snackbar error message here for potential 500 errors

frewmack (commented):
Seems to work on my end! Only issue is with error handling for 500 errors

react/src/SearchVariants/components/SearchVariantsPage.tsx
cklamann: Can you give me steps to reproduce a 500? Thanks!

frewmack: Under `variant_summary` in `flask/app/gene_viewer.py`, add `abort(500, "Internal server error")` right below the function header.

In my case specifically I got a 500 because I tried importing Delvin's sqldump without migrating properly, so line 225 `sql_df = pd.read_sql(query.statement, query.session.bind)` failed because `analyses.priority` doesn't exist in my db. This was an issue specific to me, but showed that the frontend doesn't display 500 errors at all.

kevinlul (commented):
How does this work with multiple genes?

cklamann:
@kevinlul It doesn't allow multiple genes at this point. One option would be to allow a kind of multiselect off the autocomplete box that shows the user chips for the genes they have selected, then add a "Submit" button and that will send the request back.

react/src/SearchVariants/components/SearchVariantsPage.tsx
cklamann: updated in [b23a369](https://github.com/ccmbioinfo/stager/pull/518/commits/b23a369cb7de0f6db02749e40c81265bc1cee641)

cklamann:
(basic) multiple gene selection enabled in [d68a11d](https://github.com/ccmbioinfo/stager/pull/518/commits/d68a11d671a0a96eab774e3d142ecbbf9c07106e)

frewmack (approved):
Looks good to me!

react/src/SearchVariants/components/SearchVariantsPage.tsx
```diff
@@ -20,12 +21,44 @@ const useStyles = makeStyles(theme => ({
...
+        if (error) {
+            setError(false);
+            setSelectedGenes([]);
+        }
+    };
+
+    useEffect(() => {
+        if (blob && selectedGenes) {
+            downloadCsvResponse(blob);
+            setSelectedGenes([]);
```
kevinlul, line 59: I don't think we should be wiping out the user's selection on download, but otherwise this seems good as a basis.

cklamann: updated in [6362962](https://github.com/ccmbioinfo/stager/pull/518/commits/636296232ef268ace401e3890a4b8ecd76d774bc)

kevinlul (approved):
LGTM

Issue #505, opened 2021-04-13
Epic: priority flags for analyses


kevinlul:
Currently, the workflow for infrequent high-priority analyses involves a chain of emails. As a collaborator requesting analyses, I want to label requested analyses as "clinical priority" or "research priority" if needed to streamline this process.

Tasks:
- [x] Database: add a nullable enum column to the analysis table for priority. This can go in the "initial migration" if done before Stager rollout, and definitely before the variant viewer migration if possible.
- [ ] Backend: support this feature in all analysis endpoints once the database change is made
- [ ] Frontend: add a radio button to the analysis request modal to apply a priority flag if desired (clinical, research, none)
- [ ] Frontend: highlight analyses in the main table to distinguish their priority. Add filters or chip buttons for priority. Maybe sort priority analyses to appear first if they are not completed yet.

PR #520, opened 2021-04-20
Flask: add priority column to analysis table

Madelinehazel:
Part of #505 
Making a PR in case we want to include this in the initial migration before rollout

### DIFF
--- a/flask/app/models.py
+++ b/flask/app/models.py
@@ -341,6 +341,11 @@ class AnalysisState(str, Enum):
     Cancelled = "Cancelled"
 
 
+class PriorityType(str, Enum):
+    ClinicalPriority = "ClinicalPriority"
+    ResearchPriority = "ResearchPriority"
+
+
 @dataclass
 class Analysis(db.Model):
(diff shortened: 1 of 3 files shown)


flask/app/models.py
```diff
@@ -341,6 +341,11 @@ class AnalysisState(str, Enum):
     Cancelled = "Cancelled"
 
 
+class PriorityType(str, Enum):
+    ClinicalPriority = "ClinicalPriority"
+    ResearchPriority = "ResearchPriority"
```
kevinlul, line 346: Including `Priority` in the value of the enum seems kind of redundant when both the column and the enum type are labelled as such.

kevinlul (changes requested):

flask/app/models.py
kevinlul: This also changes the number of enumerations available in `/api/enums`

Madelinehazel: fixed in e635324d3c3643d525731b5628c95a5dcd529723

kevinlul (approved):
LGTM!

PR #521, opened 2021-04-20
Flask: minor bug fixes to /api/datasets

kevinlul:
Fix error 500 due to accessing username of None assignees
Fix error 500 due to incorrect syntax for ordering on updated_by
Fix list endpoint returning tissue sample created_by/updated_by instead of that of the dataset


delvinso (approved):
LGTM!

PR #522, opened 2021-04-20
Add analysis priority flag to fe components and process attribute in backend routes

### LINKED ISSUE #505: Epic: priority flags for analyses

cklamann:
close #505 
* Allow users to specify priority in run analysis modal
* Highlight rows in analysis table with Clinical or Research priority
* Filter by priority in analysis table
* Validate priority value in patch/post routes and add priority filter to index route
* Update tests

### DIFF
--- a/flask/app/analyses.py
+++ b/flask/app/analyses.py
@@ -319,6 +319,9 @@ def create_analysis():
         datasets=found_datasets,
     )
 
+    if request.json.get("priority"):
+        analysis.priority = request.json.get("priority")
+
     db.session.add(analysis)
     transaction_or_abort(db.session.commit)
 
--- a/flask/app/analyses.py
+++ b/flask/app/analyses.py
@@ -39,6 +39,7 @@ def list_analyses(page: int, limit: int) -> Response:
         "pipeline_id",
         "assignee",
         "requester",
+        "priority",
     ]
     assignee_user = aliased(models.User)
     requester_user = aliased(models.User)
@@ -74,6 +75,15 @@ def list_analyses(page: int, limit: int) -> Response:
     notes = request.args.get("notes", type=str)
     if notes:
         filters.append(func.instr(models.Analysis.notes, notes))
+    priority = request.args.get("priority", type=str)
+    if priority:
+        filters.append(
+            filter_in_enum_or_abort(
+                models.Analysis.priority,
+                models.PriorityType,
+                priority,
+            )
+        )
     result_path = request.args.get("result_path", type=str)
     if result_path:
         filters.append(func.instr(models.Analysis.result_path, result_path))
@@ -236,6 +246,8 @@ def create_analysis():
     if not models.Pipeline.query.get(pipeline_id):
         abort(404, description="Pipeline not found")
 
+    priority = request.json.get("priority")
+
     if app.config.get("LOGIN_DISABLED"):
         user_id = request.args.get("user")
         requester_id = updated_by_id = user_id or 1
@@ -303,6 +315,12 @@ def create_analysis():
         updated_by_id=updated_by_id,
         datasets=found_datasets,
     )
+
+    enum_error = mixin(analysis, request.json, ["priority"])
+
+    if enum_error:
+        abort(400, description=enum_error)
+
     db.session.add(analysis)
     transaction_or_abort(db.session.commit)
 
@@ -410,6 +428,11 @@ def update_analysis(id: int):
         "notes",
     ]
 
+    if request.json.get("priority") and request.json.get("priority") != "None":
+        editable_columns.append("priority")
+    elif request.json.get("priority") == "None":
+        analysis.priority = None
+
     if "assignee" in request.json:
         if not request.json["assignee"]:
             analysis.assignee = None
@@ -425,7 +448,7 @@ def update_analysis(id: int):
     enum_error = mixin(analysis, request.json, editable_columns)
 
     if enum_error:
-        abort(400, description=enum_error)  # check if this works
+        abort(400, description=enum_error)
 
     if request.json.get("analysis_state") == "Running":
         analysis.started = datetime.now()
--- a/flask/app/analyses.py
+++ b/flask/app/analyses.py
@@ -1,6 +1,5 @@
 from dataclasses import asdict
 from datetime import datetime
-from typing import Container
 
 from flask_login import current_user, login_required
 from sqlalchemy import distinct, func
@@ -12,6 +11,7 @@ from . import models
 from .extensions import db
 from .utils import (
     check_admin,
+    enum_validate,
     filter_in_enum_or_abort,
     filter_updated_or_abort,
     mixin,
@@ -246,7 +246,10 @@ def create_analysis():
     if not models.Pipeline.query.get(pipeline_id):
         abort(404, description="Pipeline not found")
 
-    priority = request.json.get("priority")
+    enum_error = enum_validate(models.Analysis, request.json, ["priority"])
+
+    if enum_error:
+        abort(400, description=enum_error)
 
     if app.config.get("LOGIN_DISABLED"):
         user_id = request.args.get("user")
@@ -315,11 +318,6 @@ def create_analysis():
         updated_by_id=updated_by_id,
         datasets=found_datasets,
     )
-
-    enum_error = mixin(analysis, request.json, ["priority"])
-
-    if enum_error:
-        abort(400, description=enum_error)
 
     db.session.add(analysis)
     transaction_or_abort(db.session.commit)
--- a/flask/app/analyses.py
+++ b/flask/app/analyses.py
@@ -312,15 +312,13 @@ def create_analysis():
     analysis = models.Analysis(
         analysis_state="Requested",
         pipeline_id=pipeline_id,
+        priority=request.json.get("priority"),
         requester_id=requester_id,
         requested=now,
         updated=now,
         updated_by_id=updated_by_id,
         datasets=found_datasets,
     )
-
-    if request.json.get("priority"):
-        analysis.priority = request.json.get("priority")
 
     db.session.add(analysis)
     transaction_or_abort(db.session.commit)
@@ -427,12 +425,8 @@ def update_analysis(id: int):
         "started",
         "finished",
         "notes",
+        "priority",
     ]
-
-    if request.json.get("priority") and request.json.get("priority") != "None":
-        editable_columns.append("priority")
-    elif request.json.get("priority") == "None":
-        analysis.priority = None
 
     if "assignee" in request.json:
         if not request.json["assignee"]:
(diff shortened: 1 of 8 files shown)


flask/app/analyses.py
```diff
@@ -303,6 +315,12 @@ def create_analysis():
         updated_by_id=updated_by_id,
         datasets=found_datasets,
     )
+
+    enum_error = mixin(analysis, request.json, ["priority"])
```
delvinso, line 319: Overall the backend looks great and works for me! Only issue I came across during testing was that in the POST request, when an invalid enum for priority is supplied in the request, an error 400 is thrown but after looking into this, it's not because of `enum_error` (the priority field is actually successfully updated) but because it appears that the instance is failing to be added to the database.
```
curl -i -X POST \
    --data '{"datasets" : ["1"],
            "pipeline_id" : 1,
            "priority" : "Non"}'  localhost:5000/api/analyses
HTTP/1.0 400 BAD REQUEST
Content-Length: 63
Date: Wed, 21 Apr 2021 12:43:43 GMT

{
  "error": "Data truncated for column 'priority' at row 1"
}
```
The correct function should be `enum_validate` (for this function the first argument should be the the model itself rather than an instance of the model). The two functions do look alike and I'll admit are not named very well! I believe the difference is that `mixin` is used for allowing only specific fields to be edited, whereas `enum_validate` is for determining whether an instance of a model to be added to the database has valid enums. It also looks like some logic to account for 'None' will also need to be added like what you did for the PATCH endpoint. The logic for enum validation could also be moved higher up to say before the query so the work from the query is not wasted if priority is not a valid enum.

cklamann: Nice catch. Updated in [5b4fd5b](https://github.com/ccmbioinfo/stager/pull/522/commits/5b4fd5bef2a15a6f722cb30e842a448f9cf522ef) and [7ff4505](https://github.com/ccmbioinfo/stager/pull/522/commits/7ff4505207db53a84ebefd3aa4850a6a3bae5892). For the "None" value, I'm not sure we'll need it in the POST route? My reasoning is that when POSTing, json null would be sufficient b/c the column is nullable, so just passing nothing or null should take care of that. It would probably be better to pass `null` on the PATCH route too (rather than `"None"`) -- that might require updating the `mixin` function to check whether the column is nullable and allow null values to pass through.

flask/app/analyses.py
```diff
@@ -303,6 +318,10 @@ def create_analysis():
         updated_by_id=updated_by_id,
         datasets=found_datasets,
     )
+
+    if request.json.get("priority"):
```
delvinso, line 322: I think to keep things consistent with the other POST endpoints, you can add to 312-319 (where an analysis object is instantiated), `priority = request.json.get('priority')` which will default to none thanks to the get method if no priority field is supplied. Since the enum check occurs before any value must be valid.

flask/app/analyses.py
```diff
@@ -410,6 +429,11 @@ def update_analysis(id: int):
         "notes",
     ]
 
+    if request.json.get("priority") and request.json.get("priority") != "None":
```
delvinso, line 432: Unless there's a reason I might be missing, priority can be added to `editable_columns` by default since all other editable fields are contained in the list regardless of whether there is intent to edit?

cklamann: My thinking here was that "None" isn't a value in the enum, so an additional check is needed so that "None" doesn't get passed to the `mixin` function, which will throw a validation error. If the user wanted to change, say, priority="Clinical" to priority=`null`, either by passing in "None" or `null` as the value, the request would fail without some kind of additional check, or at least it appears that way from my testing. One solution may be to adjust the `mixin` function to accept null as a valid value for nullable enum fields? That way we wouldn't have to pass back 'None' which I agree is not great. But maybe I'm missing a step?

flask/app/analyses.py
cklamann: :+1: updated in [e4b9593](https://github.com/ccmbioinfo/stager/pull/522/commits/e4b9593025382ef80a15513aeece84099c059f1e#diff-dfde89d18ccd5b96d88d26c96d8893c69183bcb28d98a401425b603b891d7701R315)

flask/app/analyses.py
cklamann: Took a stab at this in [e4b9593](https://github.com/ccmbioinfo/stager/pull/522/commits/e4b9593025382ef80a15513aeece84099c059f1e#). I got around the enum check in `mixin` for nullable values by adding a check to see if the column is nullable [here](https://github.com/ccmbioinfo/stager/pull/522/commits/e4b9593025382ef80a15513aeece84099c059f1e#diff-31ba4ef1c8b9ff11c45c7f693530599eeb901ba5bd496a2ace1f5698752d2312R34). It also seems that `mixin` is able to determine column type from the instance only if the field already has a value for that column in the db. I wasn't able to make it work using what I know of introspection functionality, so I just separated the [validation logic](https://github.com/ccmbioinfo/stager/pull/522/commits/e4b9593025382ef80a15513aeece84099c059f1e#diff-dfde89d18ccd5b96d88d26c96d8893c69183bcb28d98a401425b603b891d7701R443) from the [update logic](https://github.com/ccmbioinfo/stager/pull/522/commits/e4b9593025382ef80a15513aeece84099c059f1e#diff-dfde89d18ccd5b96d88d26c96d8893c69183bcb28d98a401425b603b891d7701R448).

react/src/Analyses/index.tsx
```diff
@@ -355,6 +407,8 @@ export default function Analyses() {
                         exportAllData: true,
                         exportButton: { csv: true, pdf: false },
                         exportCsv: (columns, data) => exportCSV(columns, data, "Analyses"),
+                        rowStyle: data =>
+                            getHighlightColor(theme, data.priority, data.analysis_state),
```
kevinlul, line 421: I'm not exactly sure why but this is only working for the first row.

cklamann: Weird... it will only highlight the first row that has a priority flag?

flask/app/utils.py
```diff
@@ -25,10 +25,13 @@ def mixin(
 ) -> Union[None, str]:
     for field in columns:
         if field in json_mixin:
-            column = getattr(entity, field)
+            column = getattr(entity, field)  # will be None if no value for field in db
             value = json_mixin[field]
             if isinstance(column, Enum):
-                if not hasattr(type(column), str(value)):
+                is_valid_null = (
```
kevinlul, line 31: `is_null_valid` ?

react/src/Analyses/index.tsx
```diff
@@ -301,6 +332,37 @@ export default function Analyses() {
...
+                                            )
+                                        }
+                                        fullWidth
+                                    >
+                                        {enums?.PriorityType.map(p => (
+                                            <MenuItem key={p} value={p}>
+                                                {p}
+                                            </MenuItem>
+                                        ))}
+                                        <MenuItem value={"None"}>None</MenuItem>
```
kevinlul, line 361: No braces required on string property

react/src/Analyses/index.tsx
```diff
@@ -301,6 +332,37 @@ export default function Analyses() {
...
+                            field: "priority",
+                            type: "string",
+                            width: "8%",
+                            lookup: priorityLookup,
+                            editComponent: ({ onChange, value }) => {
+                                //react doesn't like default null values in forms so we have to workaround
+                                const cast = value ? value : "None";
+                                return (
+                                    <Select
+                                        value={cast}
```
kevinlul, line 346: You should be able to pass `value || "None"` here instead then

react/src/Datasets/components/AnalysisRunnerDialog.tsx
```diff
@@ -67,27 +74,61 @@ export default function AnalysisRunnerDialog({
...
-                </FormControl>
+                <Grid container direction="column">
+                    <Grid item>
+                        <FormControl component="fieldset">
+                            <FormLabel
+                                component="legend"
+                                className={pipeline ? classes.text : classes.warningText}
+                            >
+                                Pipelines*:
+                            </FormLabel>
```
kevinlul, line 85: What's the reason for distinguishing this label this way?

kevinlul (changes requested):
I wondering if it's because of the nullable definition but I noticed that after deselecting the options, the frontend still sends an empty `priority=` in the query string to the backend, which doesn't happen for other controls of this form. It doesn't affect functionality.

react/src/Analyses/index.tsx
cklamann: updated in [4618fc9](https://github.com/ccmbioinfo/stager/pull/522/commits/4618fc9cbced57960989026ffdefaffd90100b79#diff-19e03fc8aab473e13bc8ddf811cf174a6b4c23b5c0740147da368a1478089cc6R344)

react/src/Analyses/index.tsx
cklamann: updated in [4618fc9](https://github.com/ccmbioinfo/stager/pull/522/commits/4618fc9cbced57960989026ffdefaffd90100b79#diff-19e03fc8aab473e13bc8ddf811cf174a6b4c23b5c0740147da368a1478089cc6R359)

react/src/Datasets/components/AnalysisRunnerDialog.tsx
cklamann: I was trying to indicate that the priority field is required by adding the asterisk to the label, coloring the label red if the value is missing, and disabling the button if the value is missing. I can take another approach or remove if this is undesirable!

cklamann:
for the lingering `priority` filter in the request, I noticed that this is happening with the lookup filters in the datasets table as well. It seem to be a quirk of material table? Maybe a custom component is the only way to clear them?

react/src/Datasets/components/AnalysisRunnerDialog.tsx
kevinlul: It's not really indicated locally what this means. I don't think that this is really needed.

kevinlul (approved):
If it's a general quirk of material-table then let's not worry about it here. The last thing aside this looks good to me!

react/src/Datasets/components/AnalysisRunnerDialog.tsx
cklamann: removed in [1cde926](https://github.com/ccmbioinfo/stager/pull/522/commits/1cde9269aa0eddce800c5605a045ad99a9999b30)

delvinso (approved):
back-end looks good to me!

react/src/Datasets/components/AnalysisRunnerDialog.tsx
```diff
@@ -166,7 +159,6 @@ export default function AnalysisRunnerDialog({
                     Cancel
                 </Button>
                 <Button
-                    disabled={!pipeline}
```
kevinlul: I should be more clear: I think the button disabling is a good idea, but the asterisk and colouring shouldn't be here unless the dialog explains what they mean.

PR #523, opened 2021-04-22
Flask: accept trailing slashes for endpoints with children

kevinlul:
GET /api
GET, POST /api/users
GET, POST /api/groups
GET, POST /api/families
GET, POST /api/participants
POST /api/tissue_samples
GET, POST /api/datasets

e.g. you can now GET `/api/` instead of just `/api` and getting a 404 on `/api/`

### DIFF
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -41,7 +41,7 @@ datasets_blueprint = Blueprint(
 )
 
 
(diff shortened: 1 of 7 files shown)


cklamann (approved):
Looks good to me! Confirmed that trailing slashes were causing 404 w/ indicated routes/methods and that they are now returning the same result as the path w/o the slash.

PR #524, opened 2021-04-22
React: support larger page sizes and retrieving all rows

kevinlul:
This was quite the adventure. At least it wasn't hair-ripping or frustrating.

Changes the default page size to 20 and the available page sizes to 20, 50, 100, and all. 
Implements support for fetching all rows at once, since it was benchmarked to take at most five seconds on my internet (for now) with the datasets table. Requires a MaterialTablePaginationOverride and an `any` cast to work correctly.
If there are fewer rows available than the current page size, the space will no longer be filled with blanks. This happens to be needed for the above to work correctly.

Refactor: add MaterialTablePrimary for common props between the three tables
Refactor: remove unused Analyses/components/DatasetTable

The rest should be documented in comments. (Some based on research in mbrn/material-table#1685 and the references there.)


react/src/hooks/utils.tsx
```diff
@@ -103,7 +103,10 @@ export async function queryTableData<RowData extends object>(
     }
     // page information
     searchParams.append("page", `${query.page}`);
```
cklamann, line 105: When page is 0 a `page=0` param is added to the query string, which I don't think has an effect on the request.

cklamann (approved):
Looks good to me! The 'all' query itself was very quick on my machine, though the rendering time (~4k rows) was lengthy in some cases (around 2 minutes for participants), freezing up the ui in the process. If this becomes an issue,  some kind of progressive rendering approach might be needed.

react/src/hooks/utils.tsx
kevinlul: It doesn't, I guess I can group this under the condition as well.

PR #525, opened 2021-04-23
Flask: GET /api/summary/participants

### LINKED ISSUE #482: Variant viewer: GET /api/summary/participants?panel=

Required query string parameter `panel` is a semicolon-delimited list of genes to filter for in this summary report. The gene names can be exact `hgnc_gene_name`, HGNC IDs in the form `hgnc:#`, or Ensembl IDs in the form `ensg###########`. 

In JSON mode, responds with an array of participants, each with a `gene` property that matches the output of __/api/summary/variants__, though variants no longer need to identify the participant. 

In CSV mode, the results should be similar to the old sample-wise report. 

We may need to restrict the results returned so that only variants detected in analyses that the user has permission to access are returned. Due to the potentially large volume of data returned here, we may need to implement paging in JSON mode like in our other list endpoints. Some form of caching may be required if there are performance problems.

delvinso:
closes #482 
- renamed `get_variant_wise_df` to `get_report_df` to account for both variant and sample wise forms of the output report
- rename gene_viewer blueprint to variant
- like `/api/summary/variants` does not allow for querying through hgnc or ensembl id
- uses the same Accept header logic as GET /api/summary/variants (defaults to json unless text/csv specified)
- sample report has been validated against one generated from the old report aggregation script
- `/api/summary/participants`  outputs either a tidy report where each row is a participant’s variant, variant annotations and the genotype or json response with variants for requested genes under that participant’s tissue samples datasets, eg. 
```
curl  -H “Accept: application/json” localhost:5000/api/summary/participants\?panel=LOXL4\;RTEL1
[….
  {
    "dataset": [
      {
        "variants": [
          {
            "alt_allele": "T", 
            "alt_depths": 27, 
            "analysis_id": 6145, 
            "burden": 1, 
            "cadd_score": null, 
            "conserved_in_20_mammals": null, 
            "dataset_id": 8311, 
            "depth": 232, 
            "gene": "RTEL1", 
            "gene_id": 5203, 
            "gnomad_af": 4.087e-06, 
            "polyphen_score": null, 
            "position": "20:62324490", 
            "reference_allele": "C", 
            "sift_score": null, 
            "variant_id": 48815, 
            "variation": "splice_region_variant", 
            "zygosity": "Het"
          }, 
          {
            "alt_allele": "A", 
            "alt_depths": 57, 
            "analysis_id": 6145, 
            "burden": 1, 
            "cadd_score": 34.0, 
            "conserved_in_20_mammals": 0.953, 
            "dataset_id": 8311, 
            "depth": 311, 
            "gene": "LOXL4", 
            "gene_id": 7350, 
            "gnomad_af": 0.00028506, 
            "polyphen_score": 0.964, 
            "position": "10:100016632", 
            "reference_allele": "G", 
            "sift_score": 0.03, 
            "variant_id": 48193, 
            "variation": "missense_variant", 
            "zygosity": "Het"
          }
        ]
      }
    ], 
    "family_codename": "804", 
    "participant_codename": "TR0183"
  }
...
]
```


flask/app/variants.py
```diff
@@ -0,0 +1,424 @@
...
+        return jsonify(
+            [
+                {
+                    "participant_codename": dataset.tissue_sample.participant.participant_codename,
+                    "family_codename": dataset.tissue_sample.participant.family.family_codename,
+                    "dataset": [
+                        {
+                            "variants": [
+                                {
+                                    **asdict(genotype),
```
cklamann, line 230: I see the semantic utility of combining properties from the gene, variant, and genotype models here, but I also wonder what it means for programmatic consumers of the response. For instance, the front-end might need a transformer and intermediate Typescript definitions to 'rebuild' the underlying models so that they  work with existing code. But maybe these concerns are irrelevant for `summary/` endpoints, which by definition give a 'flattened' view of the data.

kevinlul: I came up with this idea but it's open to change later as we don't really know what a frontend consumer might do with these yet.

flask/app/variants.py
```diff
@@ -0,0 +1,424 @@
...
+            ]
+        )
+    elif request.accept_mimetypes["text/csv"]:
+
+        app.logger.info("text/csv Accept header requested")
+
+        try:
+            sql_df = pd.read_sql(query.statement, query.session.bind)
+        except:
+            abort(500, "Unexpected error")
```
kevinlul, line 249: In these cases we should really try to log the exception (`app.logger.error("message", exc_info=True)`) in case

flask/app/variants.py
```diff
@@ -0,0 +1,424 @@
...
+                    "participant_codename": dataset.tissue_sample.participant.participant_codename,
+                    "family_codename": dataset.tissue_sample.participant.family.family_codename,
+                }
+                for dataset in q
+            ]
+        )
+        return jsonify(
+            [
+                {
+                    "participant_codename": dataset.tissue_sample.participant.participant_codename,
```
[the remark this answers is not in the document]
delvinso: Done, also added in family ID

cklamann (approved):
LGTM!

PR #526, opened 2021-04-23
React: Add row-based editing for assignee, status on Analysis table

frewmack:
Also adds a filter component for filtering by status, and fixed query invalidation for analyses after mutations.

Known bug: 500 server error when updating analysis.assignee with valid, non-blank username; `'int' object has no attribute '_sa_instance_state'`. Seems to be related to line 421 in `analyses.py`. Blank usernames (removing assignee) works as expected. Fixed in #536.

Row-edits send all editable fields on update so due to the above backend bug, editing status can be tested by first setting the assignee field to be blank.

### DIFF
--- a/react/src/Analyses/components/SelectPipelineStatus.tsx
+++ b/react/src/Analyses/components/SelectPipelineStatus.tsx
@@ -1,0 +1,23 @@
+import React, { useMemo } from "react";
+import { MenuItem, Select } from "@material-ui/core";
+import { EditComponentProps } from "material-table";
+import { Analysis, PipelineStatus } from "../../typings";
+import { checkPipelineStatusChange } from "../../functions";
+
+export default function SelectPipelineStatus(props: EditComponentProps<Analysis>) {
+    const options = useMemo(
+        () =>
+            Object.values(PipelineStatus)
+                .filter(
+                    state => props.value === state || checkPipelineStatusChange(props.value, state)
+                )
+                .map(value => <MenuItem value={value}>{value}</MenuItem>),
+        [props.value]
+    );
+
+    return (
+        <Select value={props.value} onChange={e => props.onChange(e.target.value)}>
+            {options}
+        </Select>
+    );
+}
--- a/react/src/Analyses/components/SelectPipelineStatus.tsx
+++ b/react/src/Analyses/components/SelectPipelineStatus.tsx
@@ -1,19 +1,44 @@
 import React, { useMemo } from "react";
-import { MenuItem, Select } from "@material-ui/core";
+import { MenuItem, Select, ListSubheader } from "@material-ui/core";
 import { EditComponentProps } from "material-table";
 import { Analysis, PipelineStatus } from "../../typings";
 import { checkPipelineStatusChange } from "../../functions";
+import { useUserContext } from "../../contexts";
 
 export default function SelectPipelineStatus(props: EditComponentProps<Analysis>) {
-    const options = useMemo(
-        () =>
-            Object.values(PipelineStatus)
-                .filter(
-                    state => props.value === state || checkPipelineStatusChange(props.value, state)
-                )
-                .map(value => <MenuItem value={value}>{value}</MenuItem>),
-        [props.value]
-    );
+    const userClient = useUserContext();
+    const options = useMemo(() => {
+        if (userClient.user.is_admin) {
+            const [valid, invalid] = Object.values(PipelineStatus).reduce(
+                (prev, curr) => {
+                    if (curr === props.value || checkPipelineStatusChange(props.value, curr)) {
+                        prev[0].push(curr);
+                        return prev;
+                    } else {
+                        prev[1].push(curr);
+                        return prev;
+                    }
+                },
+                [[], []] as string[][]
+            );
+
+            return (
+                <>
+                    <ListSubheader>Valid States</ListSubheader>
+                    {valid.map(value => (
+                        <MenuItem value={value}>{value}</MenuItem>
+                    ))}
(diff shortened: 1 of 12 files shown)


react/src/components/SelectFilterComponent.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+    return (
+        <Select
+            value={selected}
+            onChange={e => updateFilter(e.target.value as string)}
+            displayEmpty
+            inputProps={{ "aria-label": `${props.columnDef.title} filter` }}
+        >
+            {props.options.map(label => (
+                <MenuItem value={label}>{label}</MenuItem>
```
cklamann, line 29: Browser complains about a missing `key` attribute here.

react/src/functions.tsx
```diff
@@ -376,3 +376,17 @@ export const updateTableFilter = (
         }
     }
 };
+
+export const checkPipelineStatusChange = (fromState: PipelineStatus, toState: PipelineStatus) => {
+    const endStates = [PipelineStatus.COMPLETED, PipelineStatus.ERROR, PipelineStatus.CANCELLED];
+    return (
+        // End states cannot change to another state
+        !(fromState in endStates) &&
```
cklamann, line 384: This looks like a Python-JavaScript mixup :grin:. ` !(endStates.includes(fromState))` gave me the results I expected.

cklamann (commented):
Besides the syntax issue in the `checkPipelineStatusChange` function, this looks good to me. I was able to update the fields as expected, and the validation/display logic for the statuses seems to be functioning as intended.

react/src/Analyses/components/SelectPipelineStatus.tsx
```diff
@@ -0,0 +1,23 @@
...
+export default function SelectPipelineStatus(props: EditComponentProps<Analysis>) {
+    const options = useMemo(
+        () =>
+            Object.values(PipelineStatus)
+                .filter(
+                    state => props.value === state || checkPipelineStatusChange(props.value, state)
+                )
+                .map(value => <MenuItem value={value}>{value}</MenuItem>),
+        [props.value]
+    );
```
kevinlul, line 16: I'm noticing a side effect here that if you choose the wrong option, you have to back out of the entire edit to try again.

react/src/components/SelectFilterComponent.tsx
```diff
@@ -0,0 +1,35 @@
+import React, { useState } from "react";
+import { Column } from "material-table";
+import { MenuItem, Select } from "@material-ui/core";
+
+export default function SelectFilterComponent<RowData extends object>(props: {
+    columnDef: Column<RowData>;
+    onFilterChanged: (columnID: string, value: any) => void;
+    options: string[];
+}) {
+    const [selected, setSelected] = useState("");
```
kevinlul, line 10: What's the difference between this component, the one used for the priority column, and the one used for the pipeline column?

frewmack: Good point. A `lookup` field for the status column would suffice, similar to the priority column. The pipeline column needed its own filter component because the value that it provides to the table differs from the name that it displays in the list of options.

react/src/Analyses/components/SelectPipelineStatus.tsx
```diff
@@ -1,19 +1,44 @@
...
+                (prev, curr) => {
+                    if (curr === props.value || checkPipelineStatusChange(props.value, curr)) {
+                        prev[0].push(curr);
+                        return prev;
+                    } else {
+                        prev[1].push(curr);
+                        return prev;
+                    }
+                },
+                [[], []] as string[][]
```
kevinlul: You can put this typing between angle brackets at `.reduce<string[][]>(`

react/src/Analyses/components/SelectPipelineStatus.tsx
```diff
@@ -0,0 +1,55 @@
...
+                [[], []] as PipelineStatus[][]
+            );
+
+            return [
+                <ListSubheader>Valid States</ListSubheader>,
+                valid.map(value => (
+                    <MenuItem value={value}>{value === oldValue ? <b>{value}</b> : value}</MenuItem>
+                )),
+                <ListSubheader>Invalid States</ListSubheader>,
+                invalid.map(value => <MenuItem value={value}>{value}</MenuItem>),
```
kevinlul, line 32: Needs key=

react/src/Analyses/components/SelectPipelineStatus.tsx
```diff
@@ -0,0 +1,55 @@
...
+            ].flat();
+        }
+        return Object.values(PipelineStatus)
+            .filter(
+                state =>
+                    props.value === state ||
+                    state === oldValue ||
+                    checkPipelineStatusChange(oldValue, state)
+            )
+            .map(value => <MenuItem value={value}>{value}</MenuItem>);
```
kevinlul, line 42: Needs key=

react/src/components/index.tsx
```diff
@@ -34,3 +34,4 @@ export { default as Note } from "./Note";
 export { default as Notification } from "./Notification";
 export { default as NotificationPopover } from "./NotificationPopover";
 export { default as SecretDisplay } from "./SecretDisplay";
+export { default as SelectFilterComponent } from "./SelectFilterComponent";
```
kevinlul, line 37: Should we still keep this file?

frewmack: I can remove it for now.

kevinlul (approved):
LGTM!

kevinlul:
Can you merge? I think #539 caused the conflicts/

PR #536, opened 2021-04-26
Flask: Fix setting assignee for analysis

frewmack:
Fixes 500 Server Error when attempting to PATCH /api/analyses/<id> with a non-blank assignee username, as mentioned in #526.


kevinlul (approved):
oops

PR #537, opened 2021-04-26
CI: login to registry with GITHUB_TOKEN

kevinlul:
https://docs.github.com/en/packages/guides/using-github-packages-with-github-actions#authenticating-to-github-container-registry GITHUB_TOKEN now supported over PATs, access permitted here https://github.com/orgs/ccmbioinfo/packages/container/stager/settings/actions_access


delvinso (approved):
LGTM

PR #538, opened 2021-04-28
React: add catch-all route for 404s

cklamann:


kevinlul (changes requested):
We need another Route in the other Switch on the page so the appbar components still get pushed to the right like the usual pages. I think `NotFound` can just be in its `index.tsx` as there's not much else going on here.

kevinlul (approved):
LGTM!

PR #539, opened 2021-04-28
React: apply new ESLint rules

cklamann:
New eslint rules:
* no useless path segments on imports
* no cyclical imports
* enforce import order (react, external, parent, sibling, index).
* no unused classes (mui)
* no unnecessary curly braces for props 
* no missing key props
* sort modules alphabetically when importing multiple exports from a module


flask/.gitignore
```diff
@@ -1,3 +1,4 @@
 __pycache__/
 venv/
 *.pyc
+.devcontainer.json
```
kevinlul, line 4: Is this from the VSCode extension?

react/package.json
```diff
@@ -1,66 +1,114 @@
 {
-  "name": "ccm-stager",
```
kevinlul, line 2: I'm not sure what happened here but the indentation shouldn't have changed. Did prettier overwrite this file?

flask/.gitignore
cklamann: Yes -- it's the settings for vscode while it's attached to a docker container (extensions to load, paths, etc.).

react/package.json
cklamann: Yeah I ran prettier to format the json. Do you want me to revert? (or adjust prettier json settings)?

flask/.gitignore
kevinlul: I think you need to activate the EditorConfig for VSCode extension as this file is missing a trailing newline but the editorconfig takes care of that.

react/package.json
kevinlul: Yeah, I don't think prettier should touch this. editorconfig has json and yaml at two-space indents.

react/package.json
```diff
@@ -41,11 +43,56 @@
...
+        {
+          "ignoreDeclarationSort": true,
+          "ignoreCase": true
+        }
+      ]
+    }
   },
   "prettier": {
     "printWidth": 100,
-    "tabWidth": 4,
```
kevinlul, line 48: Is this the only way to make it work?

kevinlul (commented):
Everything else seems fine. It seems like the workflow is failing because even though the dependencies changed, a cache hit is still being registered

react/package.json
cklamann: Nope -- a separate formatter could be set for json and prettier wouldn't see the json files anymore. But if editorconfig is handling the tab settings for all files already, prettier will obey them as long as there isn't an explicit override. Removing this setting tells prettier to do what editorconfig says, which means it can format json too.

kevinlul (approved):
Confirmed that everything is working locally 🚀

PR #540, opened 2021-04-28
React: small tweaks to column order, remove analysis chip buttons

kevinlul:
`width` isn't correctly applied so I removed it.
Enable the material-table column hiding feature.
Material-table also has a resizable columns feature but it's not really reliable. Using the re-resizable component would seemingly require reimplementing MTableHeader as the individual cell is just another table cell.
Remove redundant chip buttons on analysis table.


frewmack (approved):
LGTM! Just handle the merge with master and we're good.

PR #551, opened 2021-04-29
CI: fix stale yarn caches from always using the same key

kevinlul:
Apparently, hashFiles is relative to GITHUB_WORKSPACE, not the working directory
https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#hashfiles

I'm surprised this didn't come up earlier


cklamann:
LGTM!

cklamann (approved):

PR #552, opened 2021-04-30
React: Add progressive loading to Add Datasets suggestions

frewmack:
Families are fetched on a per-row basis depending on that row's `family_codename` field, as opposed to being fetched once at table level.

TODO: ~~Fetch families based on `inputValue` (what the user types in the Autocomplete box) as opposed to `row.family_codename` (what the user selects from the Autocomplete selection box).~~ Finished. May require some form of debounce function to prevent fetching on every keystroke.

Functionality for Add Datasets should appear identical to master. Only difference should be in how families are fetched.


frewmack:
Ready for review!

Issue #547, opened 2021-04-28
Epic: CSV export button should download all


kevinlul:
Instead of using material-table's built-in, this should download all rows matching the frontend filters directly from the backend. This will likely require a new endpoint or a modification of the endpoint for each.

PR #553, opened 2021-04-30
React: Replace material-table dependency with @material-table/core

cklamann:
* Change dependency to active fork of material-table
* This should help with issues like #547 as well as bugs related to disappearing filters


kevinlul (approved):
LGTM!

PR #567, opened 2021-05-04
Epic: Export all Data from Tables

cklamann:
refs  #483, #547
* Create useDownloadCsv react hook to share csv download functionality across components and consolidate code
* Memoize material-table columns
  - should mitigate remaining issues with disappearing filters (including those that would be caused by useDownloadCsv hook)
* Add functionality to return csv or json to analyses, datasets, and participants endpoints
  - duplicated code abstracted into functions in `utils.py`
  - add unit tests for `query_results_to_csv` function and move unit tests into subdirectory
* Add radio buttons to variant-viewer page, allowing users to select sample-wise or variant-wise reports


flask/app/utils.py
```diff
@@ -154,6 +165,73 @@ def get_minio_admin() -> MinioAdmin:
...
+    """ filter list item mappings and key order according to list of keys """
+    return [{key: row[key] for key in keys} for row in data]
+
+
+def expects_json(req: Request):
+    return req.headers.get("Accept") in ["application/json", "*/*", None]
+
+
+def expects_csv(req: Request):
+    return req.headers.get("Accept") == "text/csv"
```
kevinlul, line 232: I think `request.accept_mimetypes` was fine because Flask took care of figuring out that `*/*` means you accept `application/json`.

flask/tests/unit/test_query_results_to_csv.py
```diff
@@ -0,0 +1,42 @@
...
+    """ can we get a csv that csv.reader can process and whose contents are accurate? """
+    csv = query_results_to_csv([ModelFake(string_col="foo", integer_col=2)])
+    results_reader = reader(csv.split("\n"))
+    cols = ModelFake.metadata.tables["faker"].columns.keys()
+    for i, row in enumerate(list(results_reader)[:-1]):  # last row is empty
+        if i == 0:
+            for item in row:
+                assert item in cols
+        else:
+            assert len(row) == len(cols)
```
kevinlul, line 29: This is a really interesting fake that uses the ORM's API while avoiding any database access at all!

flask/tests/test_genes.py
```diff
@@ -33,7 +33,6 @@ def test_fetch_genes_csv(test_database, client, login_as):
     login_as("user")
     response = client.get("/api/summary/genes", headers={"Accept": "text/csv"})
     assert response.status_code == 200
-    assert response.headers["Content-Type"] == "text/csv"
```
kevinlul, line 36: What happened here?

react/src/Analyses/index.tsx
```diff
@@ -264,98 +364,7 @@ export default function Analyses() {
...
-                        },
-                        {
-                            title: "ID",
-                            field: "analysis_id",
-                            type: "string",
-                            editable: "never",
-                            defaultFilter: paramID,
-                        },
-                    ]}
+                    columns={columns || []}
```
kevinlul, line 367: Is `|| []` actually needed here?

flask/app/utils.py
cklamann: I did it this way b/c the integration test requests don't have the Accept header set, so the existing json-expecting tests for analyses, datasets, participants will fail unless it's added (hence the `None` option). I can add the header to the tests and restore the mimetypes check (or do something like `if not request.accept_mimetypes or 'application/json' in request.accept_mimetypes`) if that makes more sense.

flask/tests/test_genes.py
cklamann: Ha good question. I think I was planning to test the`csv_response` function independently but then realized it would just be a test of flask code but forgot to restore this assertion. Restored in [61302ea](https://github.com/ccmbioinfo/stager/pull/567/commits/61302ead558ef578a390c4c18c058dea6f71a2c0#diff-93c465b7fd38919afba78a72d6b399c8e754f8dec3afe5b4b56d51b0c0969a70R36)

react/src/Analyses/index.tsx
cklamann: Nope! This is the relic of an earlier approach. Fixed in [61302ea](https://github.com/ccmbioinfo/stager/pull/567/commits/61302ead558ef578a390c4c18c058dea6f71a2c0)

flask/app/utils.py
kevinlul: I see. I think that's more a problem with how we do tests, since the `request.accept_mimetypes` object takes care of a lot of things for us with its convenient syntax.

cklamann: Sounds good. I'll add the headers to the tests.

kevinlul: I think you might be able to cheese it with what the proposed `if not request.accept_mimetypes or 'application/json' in request.accept_mimetypes`

react/src/Participants/components/ParticipantTable.tsx
```diff
@@ -20,29 +28,95 @@ export default function ParticipantTable() {
...
+            {
+                title: "Participant Codename",
+                field: "participant_codename",
+                defaultFilter: paramID,
+            },
+            {
+                title: "Family Codename",
+                field: "family_codename",
+                editable: "never",
+            },
```
kevinlul, line 52: I changed this in one PR but family codename should come first.

react/src/hooks/participants/useParticipantsPage.tsx
```diff
@@ -3,9 +3,11 @@ import { useQueryClient } from "react-query";
 import { Participant } from "../../typings";
 import { queryTableData } from "../utils";
 
+export const GET_PARTICPANTS_URL = "/api/participants";
+
 async function fetchParticipants(query: Query<Participant>) {
     // fetch
-    const queryResult = await queryTableData<Participant>(query, "/api/participants");
+    const queryResult = await queryTableData<Participant>(query, GET_PARTICPANTS_URL);
```
kevinlul, line 10: Typo. I'm also not sure what this change accomplishes — is it to be consistent with the analyses hook?

react/src/SearchVariants/components/SearchVariantsPage.tsx
```diff
@@ -105,14 +120,36 @@ const SearchVariantsPage: React.FC<SearchVariantsPageProps> = () => {
...
+                            row
+                            aria-label="Pipelines"
+                            name="pipelines"
+                            value={downloadType}
+                            onChange={event =>
+                                setDownloadType(event.target.value as "variant" | "participant")
+                            }
+                        >
+                            <FormControlLabel
+                                label="Variant Wise Report"
```
kevinlul, line 142:
```suggestion
                                label="variant-wise"
```

react/src/SearchVariants/components/SearchVariantsPage.tsx
```diff
@@ -105,14 +120,36 @@ const SearchVariantsPage: React.FC<SearchVariantsPageProps> = () => {
...
+                                value="variant"
+                                control={<Radio color="primary" />}
+                            />
+                            <FormControlLabel
+                                label="Sample Wise Report"
```
kevinlul, line 147:
```suggestion
                                label="participant-wise"
```

react/src/Participants/components/ParticipantTable.tsx
```diff
@@ -205,7 +220,7 @@ export default function ParticipantTable() {
                     {
                         tooltip: "Copy combined codename",
                         icon: FileCopy,
-                        onClick: copyToClipboard,
+                        onClick: CopyToClipboard,
```
kevinlul, line 223: I think the casing got reverted from a merge conflict here?

react/src/hooks/participants/useParticipantsPage.tsx
kevinlul: Nevermind, this gets used in components now for the download csv hook

cklamann: Yes -- it's so we can export it and the hook can use it. The hook needs only a url -- the params are supplied by the caller (since they might come from a table or the autocomplete component, we can't totally abstract that piece) and the useQuery hook is built and managed by the useDownloadCsv hook. I put the url in a constant for DRY purposes in case the the url gets changed.

kevinlul (approved):
🚀

Issue #542, opened 2021-04-28
Reorganize gene table to account for the identifiers in use


kevinlul:
Our pipelines use Ensembl IDs, so these can be used as a primary key. (Note that multiple Ensembl IDs could refer to the same gene across releases, but we expect the user to understand that.)

We need to support aliased names to these IDs.

Ensembl IDs and HGNC IDs should be search options in the frontend alongside the "common names".

kevinlul:
Problem domain:
- some old reports only contain HGNC IDs
- current reports contain either Ensembl IDs or NCBI IDs in the `ensembl_id` column depending on how the variant was annotated
- reports also contain a common gene name column, which may or may not be filled out
- some genes may be referred to by multiple common names
- some names refer to multiple Ensembl IDs across releases

Solution:
Retrieve a mapping from https://www.genenames.org/download/custom/. (We can also look at FTP downloads from Ensembl and the NCBI if there are problems.) Filter out rows that are HGNC-exclusive (have blanks for other database identifiers). Use this to populate both of the below tables. When inserting reports, if we encounter Ensembl IDs that aren't in the mapping, we should also insert them separately. If we encounter NCBI IDs that aren't in the mapping, fetch the corresponding Ensembl ID from https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?db=gene&id= (ASN.1) OR https://eutils.ncbi.nlm.nih.gov/entrez/eutils/efetch.fcgi?retmode=xml&db=gene&id= (XML).

_Gene table_
- **ensembl_id**: integer primary key. Truncate the `ENSG` since we're only analyzing humans. These are a stable identifier, so we can use it as a natural key. This replaces the surrogate key across the database.
- **hgnc_id**: nullable unique integer. Truncate the `HGNC:` part. Only populated if we have it on hand.
- **ncbi_id**: nullable unique integer. Only populated if we have it on hand.

_Alias table_
- **ensembl_id**: foreign key to the gene table
- **name**: string. Not unique in case a single name corresponds to multiple IDs.

kevinlul:
Closed in #568, see #621 for remainder

PR #568, opened 2021-05-05
Flask: support gene aliases, use Ensembl IDs as gene primary key

kevinlul:
For #542


flask/app/models.py
```diff
@@ -419,14 +419,28 @@ class PipelineDatasets(db.Model):
...
+    alias_id = db.Column(db.Integer, primary_key=True)
+    ensembl_id: int = db.Column(
+        db.Integer,
+        db.ForeignKey("gene.ensembl_id", onupdate="cascade", ondelete="cascade"),
+    )
+    # Not unique in case one name corresponds to multiple ENSGs across releases
+    name: str = db.Column(db.String(20), nullable=False)
+    # No point in allowing dupes for the same identifier though
+    __table_args__ = (db.UniqueConstraint("ensembl_id", "name"),)
+
```
delvinso, line 442: We are limiting these to 'common' gene symbols (ie. HGNC) right? Is there any value in having a 'type' column (or enum) indicating what type of name is associated with the ensembl id? eg. current hgnc gene symbol, previous gene symbol, synonyms.

flask/app/models.py
```diff
@@ -419,14 +419,28 @@ class PipelineDatasets(db.Model):
...
+@dataclass
+class GeneAlias(db.Model):
+    # Autoincrement surrogate key
+    alias_id = db.Column(db.Integer, primary_key=True)
+    ensembl_id: int = db.Column(
+        db.Integer,
+        db.ForeignKey("gene.ensembl_id", onupdate="cascade", ondelete="cascade"),
+    )
+    # Not unique in case one name corresponds to multiple ENSGs across releases
+    name: str = db.Column(db.String(20), nullable=False)
```
delvinso, line 439: Increase length to 30 or even 50?

https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:52873
https://www.genenames.org/data/gene-symbol-report/#!/hgnc_id/HGNC:42360

flask/app/models.py
kevinlul: If we have that information, I suppose it couldn't hurt.

delvinso: Yes that can be extracted from the flatfile, other than the length changes and the type column it looks good to me!

kevinlul:
Going to review #567 first as that changes endpoint files that I'll need to change here

delvinso:
Ref allele, alt allele and refseq_change char length needs to be increased to at minimum:
- Ref and Alt allele: 150 -> 300 
- Refseq_change: 250 -> 500

kevinlul:
I think we should punt the frontend changes to a separate PR so this one doesn't drag on.

flask/app/variants.py
```diff
@@ -67,7 +66,6 @@ def get_report_df(df: pd.DataFrame, type: str):
                 "variation": "first",
```
delvinso, line 66: chromosome is present in participant-wise report but not variant-wise
```suggestion
                "variation": "first",
                "chromosome": "first",
```

flask/app/variants.py
```diff
@@ -67,7 +66,6 @@ def get_report_df(df: pd.DataFrame, type: str):
...
                 "refseq_change": "first",
                 "depth": list,
-                "gene": "first",
```
delvinso, line 70: i think it would be helpful to leave the gene in the reports similar to how it was done in the old reports ('long'/tidy style)

kevinlul: I think we would have to perform a separate query per gene in the panel since there's no explicit relationship. We could also generate a column in the results in SQL based on which position range matched, but I'm not sure if that's possible through the ORM.

flask/app/variants.py
```diff
@@ -95,62 +92,96 @@ def get_report_df(df: pd.DataFrame, type: str):
...
-    else:
-        user_id = current_user.user_id
-        app.logger.debug("User is regular with ID '%s'", user_id)
-
-    app.logger.debug("Parsing query parameters..")
-
-    genes = request.args.get("panel")
-
+def parse_gene_panel() -> List[Any]:
+    """
```
cklamann, line 101: This could maybe benefit from a more detailed function name or docstring -- it took me a minute to see that it's actually building the region query.

cklamann (approved):
This looks good to me! The queries/exports are bringing back the expected results (and are quite quick!).

delvinso (approved):
LGTM, works as expected!

PR #569, opened 2021-05-06
React: Update participant info dialog content and formatting

cklamann:
- add link to datasets table in participant info dialog
  * includes adding dataset_id filter to datasets endpoint 
- update datetimetext component to return empty string when date is invalid (e.g., `null`)
- allow for variable column count in detail section to reduce vertical space
- add fullwidth option to `Field` component
- hide analyses in participant dialogue by default and toggle with button
- remove unnecessary primary keys from participant info dialogue

### DIFF
--- a/react/src/Participants/components/ParticipantInfoDialog.tsx
+++ b/react/src/Participants/components/ParticipantInfoDialog.tsx
@@ -1,5 +1,14 @@
-import React, { useMemo } from "react";
-import { Dialog, DialogContent, Divider, makeStyles } from "@material-ui/core";
+import React, { useMemo, useState } from "react";
+import {
+    Box,
+    Button,
+    Collapse,
+    Dialog,
+    DialogContent,
+    Divider,
+    makeStyles,
+    Typography,
+} from "@material-ui/core";
 import { ShowChart } from "@material-ui/icons";
 import { DetailSection, DialogHeader, InfoList } from "../../components";
 import {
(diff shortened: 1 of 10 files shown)


kevinlul (commented):
I like where this is going! I'm not sure if hiding the analyses is the best right now as people might want to get to that faster. Lastly, there's an extraneous parenthesis introduced in b919264dfbb8efaa6b6c4f164181e4325f9ad0c6 that we should fix in one of these PRs

react/src/Participants/components/ParticipantInfoDialog.tsx
```diff
@@ -96,20 +125,31 @@ export default function ParticipantInfoDialog(props: DialogProp) {
...
+                        <Collapse in={analysisInfoListOpen}>
+                            {analyses && analyses.length ? (
+                                <InfoList
+                                    infoList={getAnalysisInfoList(analyses)}
+                                    enums={enums}
+                                    icon={<ShowChart />}
+                                    linkPath="/analysis"
+                                />
+                            ) : (
+                                <Typography>There are no analyses for this participant</Typography>
```
frewmack: I think this text should display instead of the button if there are no analyses associated to this participant. Alternatively, disable the button if there are no analyses.

react/src/Participants/components/ParticipantInfoDialog.tsx
```diff
@@ -96,20 +125,31 @@ export default function ParticipantInfoDialog(props: DialogProp) {
...
+
+                <>
+                    <Divider />
+                    <Box margin={3}>
+                        <Button
+                            variant="contained"
+                            size="small"
+                            onClick={() => setAnalysisInfoListOpen(!analysisInfoListOpen)}
+                        >
+                            {`${analysisInfoListOpen ? "Hide" : "Show"} Analyses`}
```
frewmack: We should display how many analyses there are in the button label (eg. "Show Analyses (3)")

frewmack (changes requested):
Overall looks good! The reduction in unused space in the detail section is great.

frewmack (approved):
My change requests are no longer relevant now that analyses are always displayed. LGTM

kevinlul (approved):
LGTM!

kevinlul:
Please make sure to include your PR message in the commit message when you squash! It doesn't happen automatically

Issue #550, opened 2021-04-28
Save material-table configurations to localStorage


kevinlul:
This should be similar to the Add Datasets page.

- [x] column order
- [x] sort order
- [x] active filters?
- [x] visible/hidden columns?

frewmack:
Update: setting `defaultFilter` on columns before passing them to material-table seems to work nicely. Still testing but it looks promising. Thanks @cklamann for the help!

PR #577, opened 2021-05-10
React: Save column order, orderBy column to localStorage

frewmack:
Part of #550

- Column order and sorting by column are loaded from localStorage on mount, and saved to localStorage every time order or sorting changes
- Loading sorting on mount triggers a second query with updated orderBy, orderDirection parameters. Unsure if there is a workaround to querying twice.

### DIFF
--- a/react/src/functions.tsx
+++ b/react/src/functions.tsx
@@ -1,3 +1,4 @@
+import React from "react";
 import dayjs from "dayjs";
 import LocalizedFormat from "dayjs/plugin/localizedFormat";
 import utc from "dayjs/plugin/utc";
@@ -372,6 +373,31 @@ export const updateTableFilter = (
             tableRef.current.onQueryChange();
         }
     }
+};
+
+/**
+ * Get the columns currently stored by material-table.
+ */
+export const getTableColumns = (tableRef: React.MutableRefObject<any>): any[] | null => {
+    if (tableRef.current) return tableRef.current.dataManager.columns;
+    return null;
+};
+
+/**
+ * Return a mapping of column ids to column order indexes currently stored by material-table.
+ */
+export const getTableColumnOrder = (
+    tableRef: React.MutableRefObject<any>
+): Record<number, number> | null => {
+    const cols = getTableColumns(tableRef);
+    const result: Record<number, number> = {};
+    if (cols) {
+        cols.forEach(col => {
+            result[col.tableData.id] = col.tableData.columnOrder;
+        });
+        return result;
+    }
+    return null;
 };
 
 export const checkPipelineStatusChange = (fromState: PipelineStatus, toState: PipelineStatus) => {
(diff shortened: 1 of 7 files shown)


react/src/hooks/useSortOrderCache.tsx
```diff
@@ -0,0 +1,70 @@
...
+            } else {
+                // cache hit
+                try {
+                    // simple format check
+                    const cachedArray = sortOrderCache.split(",");
+                    if (cachedArray.length !== 2) throw Error(`${cacheKey} is not array of size 2`);
+                    const orderBy = parseInt(cachedArray[0]);
+                    const orderDirection = cachedArray[1];
+
+                    tableRef.current.onChangeOrder(orderBy, orderDirection);
```
kevinlul, line 52: Does this also throw exceptions? Otherwise there's not much reason for the try-catch

react/src/hooks/useSortOrderCache.tsx
```diff
@@ -0,0 +1,70 @@
...
+        orderBy,
+        orderDirection
+    ) => {
+        localStorage.setItem(cacheKey, [orderBy, orderDirection].join(","));
+    };
+
+    useEffect(() => {
+        if (
+            tableRef.current &&
+            (dependencies === undefined || dependencies.find(dep => !dep) === undefined) &&
```
kevinlul, line 30: I think `dependencies.indexOf(false) === -1` may be more clear on intent

react/src/functions.tsx
```diff
@@ -374,6 +375,31 @@ export const updateTableFilter = (
     }
 };
 
+/**
+ * Get the columns currently stored by material-table.
+ */
+export const getTableColumns = (tableRef: React.MutableRefObject<any>): any[] | null => {
+    if (tableRef.current) return tableRef.current.dataManager.columns;
+    return null;
+};
```
kevinlul, line 384: I think this is only used below so you can just inline it with nullish coalescing: `tableRef.current?.dataManager.columns`

react/src/hooks/useColumnOrderCache.tsx
```diff
@@ -0,0 +1,73 @@
...
+            localStorage.setItem(cacheKey, JSON.stringify(columnOrderRecord));
+        } else {
+            console.error(`Failed to cache column order for ${cacheKey}.`);
+        }
+    };
+
+    useEffect(() => {
+        if (
+            tableRef.current &&
+            (dependencies === undefined || dependencies.find(dep => !dep) === undefined) &&
```
cklamann, line 37: I think this may have unintended consequences, since `[true,2,undefined]` would also return `undefined` (if `undefined` is a possible dep value).

cklamann (commented):
LGTM! -- I was able to save sorting and column order on the tables between page loads. As for avoiding the 2 queries with the sort hook, maybe it could be done by altering the appropriate column's `defaultSort` prop?  It seems like that may involve passing the whole column array to the hook, and I'm not sure how feasible/intrusive that would be, especially considering that those columns have to be memoized and have dependencies of their own. Maybe it's a moot point since this solution seems to work fine.

kevinlul (approved):
LGTM!

PR #579, opened 2021-05-11
React: updates to Add Datasets

cklamann:
* attach notes to dataset, not participant model in _bulk endpoint
* use defaults from session storage (if they exist) as initial search state
* create fresh defaults on form submission
* make notes field free text (though still technically autocomplete)
* update value of non-enum fields on blur
* add empty string default options where appropriate to suppress console warnings


kevinlul:
Does this address the problem with types disappearing from the autocomplete?

react/src/AddDatasets/components/TableCells.tsx
```diff
@@ -109,24 +105,54 @@ export function AutocompleteCell(
...
+        props.onEdit("");
+        //trigger dynamic option load
         if (props.onSearch) props.onSearch(value);
     };
 
-    const onEdit = (newValue: Option, autopopulate?: boolean) => {
-        onSearch(newValue.inputValue);
-        props.onEdit(newValue.inputValue, autopopulate);
-    };
+    const getIsFreeSolo = () => props.column.field === "notes";
```
kevinlul, line 126:
```suggestion
    const isFreeSolo = () => props.column.field === "notes";
```

kevinlul (changes requested):

kevinlul (approved):
LGTM!

frewmack (approved):
LGTM!

PR #587, opened 2021-05-12
React: Save default table filters to localStorage

frewmack:
Part of #550 
Sets default filters by setting defaultFilter values for columns before they are added to material-table, as opposed to injecting them into the table via ref as a side effect.

### DIFF
--- a/react/src/Participants/components/ParticipantTable.tsx
+++ b/react/src/Participants/components/ParticipantTable.tsx
@@ -20,6 +20,7 @@ import {
     useMetadatasetTypesQuery,
     useParticipantsPage,
     useSortOrderCache,
+    useTableFilterCache,
 } from "../../hooks";
 import { transformMTQueryToCsvDownloadParams } from "../../hooks/utils";
 import { Participant } from "../../typings";
@@ -47,8 +48,12 @@ export default function ParticipantTable() {
         [metadatasetTypesQuery.data]
     );
 
-    const columns: Column<Participant>[] = useMemo(() => {
-        return [
+    const { handleFilterChange, setInitialFilters } = useTableFilterCache<Participant>(
+        "participantTableDefaultFilters"
+    );
+
+    const columns = useMemo(() => {
+        const columns: Column<Participant>[] = [
             {
                 title: "Family Codename",
                 field: "family_codename",
@@ -105,7 +110,9 @@ export default function ParticipantTable() {
                 ),
             },
         ];
-    }, [datasetTypes, sexTypes, participantTypes, paramID]);
+        setInitialFilters(columns);
+        return columns;
+    }, [datasetTypes, sexTypes, participantTypes, paramID, setInitialFilters]);
 
     const tableRef = useRef<any>();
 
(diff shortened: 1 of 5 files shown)


react/src/hooks/useTableFilterCache.tsx
```diff
@@ -0,0 +1,48 @@
+import { Column, Filter } from "@material-table/core";
+
+/**
+ * Return functions for handling table filter caching.
+ *
+ * Instead of loading saved filters on mount via side effect,
+ * use `setInitialFilters` to set the default filter values
+ * for all columns before providing them to material-table as a prop.
+ */
+export function useTableFilterCache<RowData extends object>(cacheKey: string) {
```
cklamann, line 10: This may not necessarily need to be a hook. The React docs are kind of vague on this point, but if 'the rules of hooks' don't have to be followed (b/c the function isn't itself wrapping hooks), then 2 plain functions might suffice (and avoid eslint errors that target hook-specific rules).

react/src/Participants/components/ParticipantTable.tsx
```diff
@@ -168,7 +175,10 @@ export default function ParticipantTable() {
             <MaterialTablePrimary
                 tableRef={tableRef}
                 columns={columns}
-                data={dataFetch}
+                data={query => {
+                    if (query) handleFilterChange(query.filters);
```
cklamann, line 179: if you can get away with firing this only at component tear-down, you could return the cache function from a useEffect hook so it fires just once. You'd have to grab and store the filters from `tableRef.current.state.query.filters`. Not sure it's necessary but it's an option.

cklamann (commented):
The functionality worked for me in all the tables except for fields that have custom filter components (e.g., `affected` & `solved` on the participants table). The query runs perfectly but the column filter box stays blank. I think this might be a shortcoming of material-table, since it may not know how to set the value of the custom component.

react/src/hooks/useTableFilterCache.tsx
```diff
@@ -0,0 +1,48 @@
...
+                const filter = filterMapping.get(col.field);
+                if (filter !== undefined) col.defaultFilter = filter;
+            });
+        }
+    };
+
+    return {
+        handleFilterChange: handleFilterChange,
+        setInitialFilters: setInitialFilters,
+    };
```
kevinlul, line 47: If we're keeping this, you can actually use [shorthand notation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#new_notations_in_ecmascript_2015) here instead of repeating the property name.

react/src/Participants/components/ParticipantTable.tsx
frewmack: I'm not sure if I can grab the filters at tear-down. I ran into issues trying to do something similar in #577 because `tableRef.current` is null by that point.

react/src/hooks/useTableFilterCache.tsx
frewmack: Yeah I considered that this doesn't necessarily need to be a hook, but I think it still works as a convenient way to bundle both functions together, since both are required for full functionality. The `cacheKey` and typing information also get passed down to both functions, preventing us from having to write that info twice if the functions were made separate.

frewmack:
Looked into the issue regarding custom filterComponents not being updated. Apparently default filters are controlled, with the value `columnDef.tableData.filterValue`, which is supposed to be an internal value. Not sure why they let you override filterComponents and don't expose this value in types but :shrug:.

I'll handle this as a separate issue.

kevinlul (approved):
Looks good to me!

PR #588, opened 2021-05-12
React: Save hidden column states to localStorage

frewmack:
Last part of #550
Hidden columns are loaded similarly to #587.


kevinlul (approved):
LGTM!

kevinlul:
Ooh, doing this separately from #587 has resulted in merge conflicts

PR #589, opened 2021-05-13
React: Update custom material-table filters to be controlled components

frewmack:
Custom `filterComponent` needs to be controlled by value `props.columnDef.tableData.filterValue` and update function `props.onFilterChanged` in order for #587 to work.


kevinlul (approved):
Looks good to me, works once merged to master

PR #591, opened 2021-05-13
Docker: fix nginx resolving to stale DNS entries when containers are replaced

kevinlul:
This has proved effective on https://stager-dev.ccm.sickkids.ca for weeks now.


cklamann (approved):
Looks good to me!

Issue #582, opened 2021-05-11
Update dialogs for missing information


kevinlul:
Remove primary keys and improve compactness where possible (hide numbers where they don't help and use something human-readable)
Missing properties: permission group, aliases, etc.

Compact display if cluttered

PR #592, opened 2021-05-14
React & Flask: Add permission group column to datasets

### LINKED ISSUE #546: Show permission group column on the frontend
For users in multiple permission groups or admins, show the permission group in at least the dataset table

frewmack:
Closes #546.
- Datasets returned by GET /api/datasets each include an array of group_codes associated to each dataset if applicable (if current user is admin or belongs to multiple groups).
- Frontend adds column to DatasetTable for group_code iff the current user is an admin or belongs to multiple groups.
- Update frontend typing for Dataset to include group_code

### DIFF
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -199,9 +199,7 @@ def list_datasets(page: int, limit: int) -> Response:
             "family_codename": dataset.tissue_sample.participant.family.family_codename,
             "created_by": dataset.created_by.username,
             "updated_by": dataset.updated_by.username,
-            "group_code": [group.group_code for group in dataset.groups]
-            if current_user.is_admin or len(current_user.groups) > 1
-            else None,
+            "group_code": [group.group_code for group in dataset.groups],
         }
         for dataset in datasets
     ]
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -199,6 +199,9 @@ def list_datasets(page: int, limit: int) -> Response:
             "family_codename": dataset.tissue_sample.participant.family.family_codename,
             "created_by": dataset.created_by.username,
             "updated_by": dataset.updated_by.username,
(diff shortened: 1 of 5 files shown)


flask/app/datasets.py
```diff
@@ -199,6 +199,9 @@ def list_datasets(page: int, limit: int) -> Response:
             "family_codename": dataset.tissue_sample.participant.family.family_codename,
             "created_by": dataset.created_by.username,
             "updated_by": dataset.updated_by.username,
+            "group_code": [group.group_code for group in dataset.groups]
```
cklamann, line 202: #582 mentions showing permissions groups in dialogs. Will this conditional prevent that functionality or do we want to display the group codes in the dialogs only if the dataset is associated with more than one group?

frewmack: I'm actually pretty unsure about this. It might just be fine to just always send group code information, and then just selectively hide it on the frontend. I don't see any security concerns for people knowing who else has access to the datasets that they have access to. @kevinlul thoughts?

kevinlul: So `/api/groups` only allows regular users to read groups that they are a member of. This would make a user aware of other groups, though there would be ways to indirectly determine that anyway if changes occurred that were not because of any user in the group, and the username is revealed when it's edited. I think we are okay without the condition?

kevinlul (commented):
Trying to sort the column breaks the table. Should searching be in scope or out of scope for this PR?

frewmack:
I'll look into sorting. I think filtering by permission group should be a separate issue.

frewmack:
On second thought, it may be better to look at filtering and sorting at a different time. This issue also applies to the Dataset Types column in the Participants Table, as it can be neither sorted nor filtered. For now, I'll disable frontend filtering and sorting for both.

kevinlul (approved):

PR #601, opened 2021-05-18
React: Update dialogs for missing info

### LINKED ISSUE #582: Update dialogs for missing information

cklamann:
closes #582 
* Remove primary keys from modals
* make inputs full width by default and reduce column width to save space
* add group codes and aliases


kevinlul (changes requested):
So batch ID and sequencing ID aren't primary keys; they're external identifiers and just presently unused, but more relevant for RNA sequencing in the future. I still found a few more invalid dates when processing nulls, with library prep date, extraction date, and some timestamps related to analyses being affected.

PR #608, opened 2021-05-20
Flask: Fix test analyses to be compatible with respective pipelines

### LINKED ISSUE #604: Test database analyses are incompatible with their respective pipelines
Datasets included in the analyses in the test data are not compatible with the pipelines that those analyses use

frewmack:
Closes #604 
- Added test database overview to docs


kevinlul (changes requested):
Good idea! I think you need to update your local MinIO image. A couple of months ago, a new default `consoleAdmin` policy started being included, which is why the base number of policies is now 5 instead of 4.

kevinlul (approved):
LGTM!

PR #620, opened 2021-05-26
Flask: Update user endpoints to support OIDC fields

frewmack:
- Update POST /api/users to support OIDC fields
- Update PATCH /api/users/:username to support OIDC fields
- Update jsonify_user to return OIDC fields if permitted
Example:
```
curl --header "Content-Type: application/json" -X POST --data '{"username":"testuser","password":"secret","email":"user@example.com","issuer":"https://accounts.google.com","subject":"123456789"}' http://localhost:5000/api/users
{
  "deactivated": false, 
  "email": "user@example.com", 
  "groups": [], 
  "is_admin": false, 
  "issuer": "https://accounts.google.com", 
  "last_login": null, 
  "minio_access_key": "2d3a51503ce710ed", 
  "minio_secret_key": "8766837f5b275d17891bc982fc4aaf8a", 
  "subject": "123456789", 
  "username": "testuser"
}
```

```
curl --header "Content-Type: application/json" -X PATCH --data '{"issuer":"https://accounts.example.com","subject":"987654321"}' http://localhost:5000/api/users/testuser
{
  "deactivated": false, 
  "email": "user@example.com", 
  "groups": [], 
  "is_admin": false, 
  "issuer": "https://accounts.example.com", 
  "last_login": null, 
  "minio_access_key": "2d3a51503ce710ed", 
  "minio_secret_key": "8766837f5b275d17891bc982fc4aaf8a", 
  "subject": "987654321", 
  "username": "testuser"
}
```

### DIFF
--- a/flask/app/users.py
+++ b/flask/app/users.py
@@ -48,18 +48,24 @@ def jsonify_user(user: models.User) -> Response:
     Helper function for serializing individual users with their MinIO credentials
     for the remaining endpoints.
     """
-    return jsonify(
-        {
-            "username": user.username,
-            "email": user.email,
-            "is_admin": user.is_admin,
-            "last_login": user.last_login,
-            "deactivated": user.deactivated,
-            "groups": [group.group_code for group in user.groups],
-            "minio_access_key": user.minio_access_key,
-            "minio_secret_key": user.minio_secret_key,
-        }
-    )
+    user_dict = {
+        "username": user.username,
+        "email": user.email,
+        "is_admin": user.is_admin,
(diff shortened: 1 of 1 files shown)


kevinlul (commented):
Should these fields still be updatable when OAuth isn't enabled?

frewmack:
I'm not entirely sure if they should be. I left them open since the OAuth fields are always present in the User table regardless of whether OAuth is enabled or not, but I do hide them on the frontend (#623) when it's disabled.

kevinlul (approved):
LGTM!

PR #622, opened 2021-05-26
React/Flask: Add fields to analyses table and introduce multifield search

### LINKED ISSUE #585: How do I find analyses quickly given a family or participant?
This is the most frequent use of Stager. The autoincrement primary key has no semantic meaning, so it's unhelpful in both the notification pane and as the only general thing distinguishing rows in the table.

As a short identifier, we could use `family_participant_datasettype` as a non-unique display for singletons, `family_codename` if we determine an analysis to be a trio or similar, or a list of the non-unique identifiers otherwise (uncommon) that takes up more room.

cklamann:
closes #585 
* add `participant_codenames` and `family_codenames` to analysis index route json and csv payload along with eager loading and filtering
* add multified filter to analyses index route, encompassing `family_codenames`, `participant_codenames`, `participant_aliases`, and `family_aliases` fields and query from search box on analyses page
* truncate path prefix field on analyses table, show in full on hover and in dialog to save space
* use index route for all analyses-page fetches to ensure payload consistency


react/src/Analyses/index.tsx
```diff
@@ -257,6 +294,7 @@ export default function Analyses() {
                 title: "Path Prefix",
                 field: "result_path",
                 type: "string",
+                render: rowData => <TruncatedText text={rowData.result_path} maxLength={15} />,
```
frewmack, line 297: We have a `Note` component that also provides a similar functionality to this which we use for the Notes column.

frewmack (changes requested):

frewmack (approved):
LGTM!

PR #624, opened 2021-05-27
React: update variant viewer frontend to match #568 

### LINKED ISSUE #621: Variant viewer incompatible with new gene table
Variant viewer (mainly autocomplete box and typing) needs to be updated to match new gene table structure.

kevinlul:
Fix up error handling, store gene panel in localStorage
Closes #621 

Is it worth it to make the RadioGroup a controlled component?

### DIFF
--- a/flask/app/variants.py
+++ b/flask/app/variants.py
@@ -111,12 +111,12 @@ def parse_gene_panel() -> List[Any]:
         abort(400, description="No gene(s) provided")
     genes = genes.lower().split(",")
     app.logger.info("Requested gene panel: %s", genes)
-    ensgs = []
+    ensgs = set()
     errors = []
     for gene in genes:
         if gene.startswith("ensg"):
             try:
-                ensgs.append(int(gene[4:]))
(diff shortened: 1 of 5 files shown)


cklamann (commented):
This looks good to me as far as the refactoring goes. The only issue I'm running into seems to be an edge case related to the new schema. It happens when I search/select 2 aliases that map to the same ensembl_id (e.g., LMWPTP and ACP1). I get a 400 and then the ui becomes unstable w/ unpredictable re-fetches on chip removal or focus events. Some of this might be related to duplicate ensembl_id keys on the chips. One solution might be to group aliases with the same ensembl id on one chip and strip the dupes from the query?

For the radio component, I tend to make everything controlled by habit but if you see a better way please go ahead. Probably the RadioGroup aria-label and name props should be updated as well, since it looks like I made a copy-paste error there :grimacing:.

cklamann (approved):

frewmack (approved):
LGTM!

PR #631, opened 2021-05-31
Flask: Add MinIO permission group for Stager admins

### LINKED ISSUE #548: Give Stager admins full visibility on MinIO and delete permissions

frewmack:
Closes #548
Admins have readwrite access to all buckets in MinIO. Users with is_admin permissions should have access_keys belonging to the "admin" group in MinIO with this readwrite policy.

### DIFF
--- a/flask/app/__init__.py
+++ b/flask/app/__init__.py
@@ -1,7 +1,8 @@
 import logging
+from datetime import date, time, datetime
 from flask import Flask, logging as flask_logging
+from flask.json import JSONEncoder
 from .extensions import db, login, migrate, oauth
-from .utils import DateTimeEncoder
 
 from app import (
     buckets,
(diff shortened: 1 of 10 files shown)


frewmack:
I'm stumped as to why the tests are failing. Somehow a RuntimeError is escaping from the try-except block in `users.py::save_remove`, and I have no idea why that is...

kevinlul:
Might it be because we're trying to alter properties of users that actually don't exist in MinIO? As in, an access key is stored in MySQL but it doesn't exist in MinIO?

kevinlul (changes requested):
Overall I'm not sure about attaching the policy to an `admin` group vs admin users directly. It does scale out easier, since a user may only have one policy set but any number of groups, but this also requires reserving a special `admin` group in MinIO (who aren't MinIO admins, just full `readwrite`, a policy included by default).

cklamann (approved):
LGTM!

Issue #545, opened 2021-04-28
Pass region name to minio_client for bucket creation


kevinlul:
Bucket creation will fail if the region of the bucket is different than the default

PR #635, opened 2021-06-03
Flask: support MINIO_REGION_NAME environment variable

cklamann:
refs #545 
- add region name to all compose files and sample .env
- pass region name to `make_bucket` method to avoid default region if region has been specified


kevinlul (approved):
LGTM!

PR #649, opened 2021-06-07
Flask & React: Log user out of OAuth session as well as Stager

frewmack:
Users who sign in using either Auth0 or Keycloak can now sign out of their OAuth session by signing out of Stager. Other OAuth providers will need to be added on a case-by-case basis since the specs for OIDC single sign-out are not finalized.


cklamann (dismissed):
Looks good to me! I was able to log in and log out with keycloak and confirm that the keycloak session was destroyed on logout.

cklamann (approved):
Looks good! I discovered a small timing issue in the log in flow but it has nothing to do with this PR and can be addressed separately.

PR #658, opened 2021-06-08
Flask: Allow sorting by analysis_id, dataset_id

### LINKED ISSUE #656: BUG: Can't sort by dataset or analysis id
Sorting by id on the analyses and dataset tables returns a 400. I think we should make these sortable rather than disabling sorting on the table controls.

frewmack:
Closes #656


cklamann (approved):
LGTM!

PR #662, opened 2021-06-10
React: Add descriptions to variant report types

### LINKED ISSUE #654: Variant export UX improvements
- [ ] describe difference between variant-wise and participant-wise reports on front-end
- [x] make column order as similar as possible between exports
- [ ] show sample gene query
- [ ] update download name of participant-wise report (currently 'sample-wise')

frewmack:
~~Part of~~ Closes #654
Also fixes the filename for participant-wise reports.
Thanks to @delvinso for the descriptions.

### DIFF
--- a/react/src/SearchVariants/components/CardButton.tsx
+++ b/react/src/SearchVariants/components/CardButton.tsx
@@ -13,14 +13,19 @@ export interface CardButtonProps {
     description?: string;
     onClick?: () => void;
     selected?: boolean;
+    disabled?: boolean;
 }
 
 const useStyles = makeStyles<Theme, CardButtonProps>(theme => ({
     root: props => ({
-        outlineColor: theme.palette.primary.light,
-        outlineWidth: props.selected ? 3 : 0,
+        outlineColor: props.disabled ? theme.palette.action.disabled : theme.palette.primary.light,
+        outlineWidth: props.disabled ? 0 : props.selected ? 3 : 0,
         outlineStyle: "solid",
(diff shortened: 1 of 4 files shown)


react/src/SearchVariants/components/CardButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+    makeStyles,
+    Theme,
+    Typography,
+} from "@material-ui/core";
+
+export interface CardButtonProps {
+    title: string;
+    description?: string;
+    onClick?: (value: any) => void;
+    value?: any;
```
cklamann, line 15: Is this being used?

react/src/SearchVariants/components/CardButton.tsx
```diff
@@ -0,0 +1,47 @@
...
+    CardContent,
+    makeStyles,
+    Theme,
+    Typography,
+} from "@material-ui/core";
+
+export interface CardButtonProps {
+    title: string;
+    description?: string;
+    onClick?: (value: any) => void;
```
cklamann, line 14: I think we may not need an argument here, unless it is the click event itself.

cklamann (changes requested):
The cards look really good and the descriptions are clear. I think that if we are going to disable the download button if no genes have been selected, it makes sense to disable the card-buttons too, so the user has only one active control to start with (not sure how that would look -- maybe just suppress the outline and fade the colors until a gene is selected?).

cklamann (approved):
Looks good!

Issue #603, opened 2021-05-19
Sorting and Filtering for table columns with multiple values


frewmack:
See Dataset Types in Participant table, Permission groups in Datasets table

- [x] Filtering: string search for value in list (eg. `/api/participants?dataset_types=RGS` looks for RGS for each value in the list per row)

- [ ] Sorting: sorts by the column as though each list in each row is itself sorted in descending order. eg. `/api/datasets?order_by=permission_groups&order_dir=desc`
```
ach
ach,bcch
bcch
bcch,c4r
c4r
```

PR #664, opened 2021-06-11
Flask & React: Filter by multi-value columns

frewmack:
Part of #603 
Adds filtering by permission_group for datasets, and dataset_type for participants

### DIFF
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -186,15 +186,7 @@ def list_datasets(page: int, limit: int) -> Response:
     group_code = request.args.get("group_code", type=str)
     if group_code:
         subquery = (
-            models.Dataset.query.join(
-                models.groups_datasets_table,
-                models.Dataset.dataset_id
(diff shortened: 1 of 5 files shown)


flask/app/datasets.py
```diff
@@ -183,6 +183,25 @@ def list_datasets(page: int, limit: int) -> Response:
...
+        subquery = (
+            models.Dataset.query.join(
+                models.groups_datasets_table,
+                models.Dataset.dataset_id
+                == models.groups_datasets_table.columns.dataset_id,
+            )
+            .join(
+                models.Group,
+                models.Group.group_id == models.groups_datasets_table.columns.group_id,
+            )
```
cklamann, line 197: I think this can be simplified by joining direction on the relationship like:

```
models.Dataset.query.join(
                models.Dataset.groups
            )
```

cklamann (changes requested):

cklamann (approved):
Looks great!

PR #665, opened 2021-06-11
Flask: changes, re-factoring endpoints for adding requested gene(s) to JSON and CSV

### LINKED ISSUE #653: Add gene identifiers to variant exports
Depending on performance, this may require a change in how we link genes/gene aliases to variants.

delvinso:
closes #653 

- merged participant and variant endpoints into one endpoint since both now use the same query
- description below, from endpoint docstrings
```

    The same sqlalchemy query is used for both endpoints as the participant-wise report is the precursor to the variant-wise report.

    The JSON response for participants is de-normalized such that each object is a participant and a variant,
    whereas for the variant JSON response, each object is a variant, the annotations, and an array of genotypes for the involved participants.

    Similarly, the csv output for the participants is de-normalized such that each row is a participant's variant. If the requested genes span similar coordinates duplicated variants will be returned, for each gene.
    The variant csv output is a summary - each row is a unique variant with various columns collapsed and ';' delimited indicating for example, all participants that had such a variant.

```

### DIFF
--- a/flask/app/variants.py
+++ b/flask/app/variants.py
@@ -59,11 +59,10 @@ def get_report_df(df: pd.DataFrame, type: str, relevant_cols=relevant_cols):
     # some columns are duplicated eg. dataset_id, is there a way to query so that this doesn't happen?
 
     df = df[~df["zygosity"].str.contains("-|Insufficient")]
-
-    df = df.fillna('')
+    df = df.fillna("")
     df = df.astype(str)
-    df['ensembl_id'] = df['ensembl_id'].apply(lambda x: 'ENSG' + x.rjust(11, '0'))
-    
+    df["ensembl_id"] = df["ensembl_id"].apply(lambda x: "ENSG" + x.rjust(11, "0"))
+
     if type == "participants":
         return df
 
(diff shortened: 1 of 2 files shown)


cklamann (commented):
Looks good and seems fast! A few small things: 1) it appears that `nan` in the csv corresponds to `null` in the JSON? Is it possible for the csv to display `null` in these cases or maybe just an empty string (empty string might be better for sorting)? 2) the ensembl id in the variant summary csv resembles a serialized Python set (e.g., `{'157193'}`). Can it be shown in the same format as the participant ensembl id? 3) related to the last point: can you display the full ensembl id rather than our internal version, since our users are probably expecting to see the complete identifier rather than the shortened form?

delvinso:
> Looks good and seems fast! A few small things: 1) it appears that `nan` in the csv corresponds to `null` in the JSON? Is it possible for the csv to display `null` in these cases or maybe just an empty string (empty string might be better for sorting)? 2) the ensembl id in the variant summary csv resembles a serialized Python set (e.g., `{'157193'}`). Can it be shown in the same format as the participant ensembl id? 3) related to the last point: can you display the full ensembl id rather than our internal version, since our users are probably expecting to see the complete identifier rather than the shortened form?

Good points, fixed in 4f279ff! This may be niche, but in variant csv mode, ensembl id is now also ';' delimited. eg. if a variant is found in multiple requested genes it will indicate as so in the report.

cklamann (approved):
LGTM!

PR #675, opened 2021-06-15
Flask: Fix GET /api/datasets returning fewer rows than expected

### LINKED ISSUE #661: BUG: [GET] datasets results truncated
Some queries to the datasets index route return fewer than the number of results specified by the `limit` query param. This seems related to outer joins on related fields (permission groups, files). When multiple related records are located, a row is created for each relationship, meaning that 20 rows may actually map only to (say) 10 unique datasets by the orm. A solution may lie in changing the way the relationships are eager-loaded.

frewmack:
Closes #661 
Also fixes a frontend bug where empty array filters in material-table (eg. for enum columns) would be passed on as empty filter parameters when querying.

The problem seems to be fixed, but I definitely need help determining whether this approach (using multiple `joinedloads`) is efficient, or if we should use `subqueryload` at all.


flask/app/datasets.py
```diff
@@ -163,31 +170,15 @@ def list_datasets(page: int, limit: int) -> Response:
...
-        )
+        query = query.options(joinedload(models.Dataset.groups)).filter(*filters)
 
     total_count = query.with_entities(
         func.count(distinct(models.Dataset.dataset_id))
     ).scalar()
 
-    datasets = query.order_by(order).limit(limit).offset(page * (limit or 0)).all()
+    datasets = (
+        query.distinct().order_by(*order).limit(limit).offset(page * (limit or 0)).all()
```
delvinso, line 180: i'm still a bit confused about the distinct call as (unless i'm misunderstanding) it seems to me the resulting rows from the query would already be unique? i tried omitting the distinct call from the query and all the tests still pass.

frewmack: I think I figured it out. The distinct call was put in-place first when we were doing an outer join by groups. This reduced the number of rows in the query such that only 1 of each dataset_id occurred.

Since then, we've replaced the outer join by groups with a joinedload by groups, eliminating the need for the distinct call at all since groups aren't part of the query. So we accidentally found 2 ways to solve the problem: outer-joining groups and using distinct, or eager-loading groups.

flask/app/datasets.py
```diff
@@ -137,24 +140,28 @@ def list_datasets(page: int, limit: int) -> Response:
...
+            contains_eager(models.Dataset.files),
+            contains_eager(models.Dataset.updated_by),
+        )
+        .join(models.Dataset.tissue_sample)
+        .join(models.TissueSample.participant)
+        .join(models.Participant.family)
+        .outerjoin(models.Dataset.files)
+        .join(models.Dataset.updated_by)
+    )
+
```
cklamann, line 161:
```suggestion
    query = (
        models.Dataset.query.options(
            #tell the ORM that our join contains models we'd like to eager load, so it doesn't try to load them lazily
            #in my test, this sped up loading time by ~3x
            contains_eager(models.Dataset.tissue_sample)
            .contains_eager(models.TissueSample.participant)
            .contains_eager(models.Participant.family),
            #eager load groups and files for speed, since we're not joining here, we can't order or search by these fields
            #ordering is probably not a great loss b/c one-to-many makes it tough,
            # but if we want to search, a solution might be a subquery (something similar to the filter in analyses.py:194-208)
            # FE will have to be adjusted to disallow sorting/filtering on these fields
            selectinload(models.Dataset.groups),
            selectinload(models.Dataset.files),
        )
        # join with these since they are 1-to-1 and we want to order/filter on some of their fields
        .join(models.Dataset.tissue_sample)
        .join(models.TissueSample.participant)
        .join(models.Participant.family)
        .join(models.Dataset.updated_by)
    )

```

flask/app/datasets.py
```diff
@@ -163,31 +170,13 @@ def list_datasets(page: int, limit: int) -> Response:
...
-                contains_eager(models.Dataset.updated_by),
-            )
-            .join(models.Dataset.tissue_sample)
-            .outerjoin(models.Dataset.groups)
-            .join(models.TissueSample.participant)
-            .join(models.Participant.family)
-            .outerjoin(models.Dataset.files)
-            .join(models.Dataset.updated_by)
-            .filter(*filters)
-        )
```
cklamann, line 184: :+1:

flask/app/datasets.py
```diff
@@ -208,6 +197,20 @@ def list_datasets(page: int, limit: int) -> Response:
...
+        total_count,
+    )
+
+    if len(datasets) != min(limit or total_count, total_count):
+        app.logger.warning(
+            "Datasets to be returned is incorrect; expected: '%d', actual: '%d'",
+            min(limit or total_count, total_count),
+            len(datasets),
+        )
+
```
cklamann, line 213: I don't think we need the warning b/c our tests should fail at merge time if there's an issue here. Is the debugging statement for development purposes or logging? (TBH, I'm still not 100% familiar with logging strategy).

cklamann (changes requested):
I like how the code is consolidating! Right now I'm seeing an issue with the `files` relationship when I test on the front-end. It seems that every dataset comes back with only one linked file, even if it has several. I think this is related to the remaining outer join. AFAIK, `files` is like `groups` -- both are one-to-many and can lead to weird result sizes with certain joins. I think the easiest solution would be to eager load both and disable sorting. If we need filtering, we can do it with a subquery.

cklamann (approved):
:rocket: :rocket: :rocket:

Issue #676, opened 2021-06-16
Backend: Some datasets have mismatched pipelines


cklamann:
Following data migration, some datasets are linked to the wrong pipelines. For example a dataset with `dataset_type` of `WES` is attached to an analysis that is attached to the pipeline with name `crg2`. We should write a flask click command or a SQL script to update the pipelines to the correct value. I lean toward a click command b/c we can write a test for it and it will be preserved in source code.

cklamann:
Closed after #677 was merged and run in production.

PR #677, opened 2021-06-17
Flask/DB: Click command to reassociate analyses with correct pipelines

cklamann:
refs #676 
- this click command locates analyses associated with incorrect pipelines based on related `dataset.dataset_type` and updates them to the correct value.
- the script can be run via docker with `docker exec <app-container> flask update-analysis-pipelines` or in a shell inside the app container.
- I can provide a sql query to check for misassociated pipelines via DM that can be run before and after script execution to verify results.


delvinso (approved):
Looks good! I can confirm it produces the expected results through the SQL statement you provided and consequently the ORM.

PR #678, opened 2021-06-18
React: Align variant cards vertically

cklamann:
Depending on screen width, cards on variant search screen might fall of out of vertical alignment if they aren't flex containers.


frewmack (approved):
LGTM!

PR #687, opened 2021-06-23
React: Variant export enhancements for frontend

frewmack:
Part of #699 

- Adds a dropdown to the Autocomplete component for choosing a 'search category' including variant position, region, and RSID
- Adds a select box for selecting columns to include in the report. Columns are saved to localStorage 
- Render the above features only in development until the backend endpoint is updated to support these features
- Slightly update the styling for the Autocomplete for end adornments


react/src/SearchVariants/components/Autocomplete.tsx
```diff
@@ -1,20 +1,112 @@
...
+                variant="menu"
+            >
+                {searchCategoryList.map(category => (
+                    <MenuItem
+                        key={category}
+                        onClick={() => props.onSelect(category)}
+                        selected={props.value === category}
+                    >
+                        {"Search by " + snakeCaseToTitle(category)}
+                        <div style={{ flex: 1 }} />
```
cklamann, line 69: is this needed?

react/src/SearchVariants/components/Autocomplete.tsx
```diff
@@ -1,20 +1,112 @@
...
+    OutlinedInput,
+    useTheme,
+} from "@material-ui/core";
+import { ArrowDropDown, Check, Search } from "@material-ui/icons";
 import { Autocomplete } from "@material-ui/lab";
+import { snakeCaseToTitle } from "../../functions";
 import { useGenesQuery } from "../../hooks/genes";
 import { GeneAlias } from "../../typings";
 
+type SearchCategory = "gene" | "region" | "variant_position" | "rsld";
```
cklamann, line 19: rsid (here and elsewhere, display value would be rsID or refSNP ID, we should check for the exact preference)

react/src/SearchVariants/components/Autocomplete.tsx
```diff
@@ -1,20 +1,112 @@
...
+            </Button>
+            <Divider className={classes.divider} orientation="vertical" flexItem />
+            <Menu
+                anchorOrigin={{ horizontal: "left", vertical: "bottom" }}
+                transformOrigin={{ horizontal: "left", vertical: "top" }}
+                anchorEl={anchorEl}
+                keepMounted
+                open={!!anchorEl}
+                onClose={() => setAnchorEl(null)}
+                variant="menu"
```
cklamann, line 60: Was getting a cryptic console warning about these settings, evidently you need to manually set                 `getContentAnchorEl={null}` to pacify it.

react/src/SearchVariants/components/SearchVariantsPage.tsx
```diff
@@ -152,6 +245,37 @@ const SearchVariantsPage: React.FC<SearchVariantsPageProps> = () => {
                                 disabled={disableControls}
                             />
                         </Grid>
+                        {process.env.NODE_ENV === "development" && (
```
cklamann, line 248: This is cool and gives a good baseline for thinking about how this is all going to work once the pieces are hooked up. A few ideas from purely UX perspective:
1. can we see how this looks in a (big) modal? It might make more efficient use of real estate to put the values in columns.
2. can you sort the column names alphabetically?
3. select all / unselect all buttons
4. all columns selected by default
5. remove the chips and replace with some kind of indicator text, like "22 out of 46 columns selected" with a button next to it that says something like "add/remove columns" that pulls up the modal.

react/src/SearchVariants/components/Autocomplete.tsx
frewmack: Oh nice, this change actually doubles as a way to make the popover appear below the Autocomplete bar so that you can actually see the changes made.

frewmack:
Please take a look at the new modal and see if there's any changes needed.

frewmack:
This branch should be ready for review again. The frontend variant export components should now be compatible with the changes made in #717: users can now search for variants using positions and regions instead of gene names, and they can select columns which will be passed onto the variant summary endpoint.

Question: Should we incorporate a text mask library like [imaskjs](https://imask.js.org/) to help the user with entering regions/positions in the correct format? Right now the frontend performs a simple regex check that tells them that the format is wrong until they get it right.

cklamann:
Great! I think the regex is probably sufficient for now. We can collect feedback from users and make changes as necessary.

cklamann (approved):
Looks great! :rocket:

PR #688, opened 2021-06-24
React: Update OAuth login flow for untracked users

### LINKED ISSUE #683: Keycloak: handle users who successfully authenticate but are not STAGER users
- don't start a STAGER session, even if users successfully authenticate via Keycloak, if they do not have a STAGER user record
- update UI to advise these users on getting a STAGER login
- automatically link to Keycloak login (remove 'login with Keycloak link') [not exactly related]

frewmack:
Closes #683

### DIFF
--- a/flask/app/manage.py
+++ b/flask/app/manage.py
@@ -76,7 +76,7 @@ def update_analysis_pipelines() -> None:
             .pipeline_id
         )
 
-        """  
+        """
         This is not the fastest or nicest way to do this, but the only way to override the onupdate setting on models.Analysis
         seems to be passing in the the model's own current `updated` value. Otherwise sqlalchemy would touch the `updated` timestamp for each analysis.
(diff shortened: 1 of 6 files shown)


cklamann (changes requested):
This is coming along! The various redirects are definitely adding complexity, though. Some things I noticed:
- I get a 404 (`/oidc_callback` route not found) on the front end after logging in, I think b/c the routing logic for `/oidc_callback` is still in the LoginPage.
- there's a loop that traps non-stager users when they fail to log in ('go back to log in' brings them back to where they already are). I think we  need either to drop the button or else destroy the keycloak session so the user can get routed back to the login screen successfully. 
- there is some flashing of intermittent pages that happens during login/logout, and in some cases after login the 'landing' page starts to load data and then aborts after the last redirect. Is it possible to use a loading modal or dummy page or otherwise prevent premature or intermittent renders?

cklamann (approved):

PR #690, opened 2021-06-25
Flask/Minio: Remove Minio group policies from admins to prevent overrides

### LINKED ISSUE #689: BUG: Minio individual readwrite policy being overridden by group policy
Currently, users with individual `readwrite` policies are unable to perform actions explicitly denied by the policies of their associated groups. According to Amazon's IAM docs, ["An explicit deny in any policy overrides any allows"](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_evaluation-logic.html). So we either need to rework the group permissions so that they don't have any explicit denies, or remove admins from the groups altogether. The latter seems simplest.

cklamann:
Closes #689


frewmack (approved):
LGTM!

PR #691, opened 2021-06-28
Flask/db: update schema to expand report column coverage

### LINKED ISSUE #685: Flask/Database: Expand gene-pull column coverage
Sub-issue related #655. 

- There are many columns in the reports that are currently not in the database.
- Some of these columns are static, some are analysis-specific. 
- When adding these columns to the database, some questions to consider:
  - does the column belong on the `variant` table or `genotype` table or somewhere else?
  - should data be in 'long' or 'wide' format?
  - where should our update scripts live?

Sub-tasks include:

- [ ]  exploring report dictionary, modifying the ORM and creating the migration scripts
- [ ]  modifying existing report insertion scripts for c4r to click command(s), and modify to account for new fields

delvinso:
closes #685 

Changes include modifying variant and genotype tables and corresponding alembic migrations to accommodate the remaining fields not initially included.

- table consists of both ‘variant-analysis’ and ‘variant-annotation’ fields
- variant-analysis level (from vcf) - exists in variant table but ideally be specific to an analysis
- variant-annotation level - downstream vcf annotations, also exists in variant table, but ideally would be specific to a variant
- field goes in genotype table if it can be decomposed and assigned to a specific dataset

Previous scripts used to map and insert c4r reports are now a click command (`map-insert-c4r-reports `) and have been modified to accommodate insertion of the remaining fields into the db. The command can be run on the local machine as `docker-compose exec app flask map-insert-c4r-reports report_path` , where report_path is now a volume mount in the docker-compose yaml.

Three files are dumped on completion of the command, a csv and two pickles, the latter two are dictionary (key: family codename) of dictionaries (key: participant codename) with metadata on the identified families, participants, analysis ids, dataset ids, and report paths for each report. They can be extracted through `docker cp` or some variation. 

Using the 06-28-2021 Stager production dump, I was able to map 1115/1515 reports. 

ps. The variant summary report is now an eyesore but we don't expect users to select all fields given the option.


frewmack (commented):
Looks good! Are these new columns the only ones that will be provided in variant reports? I'm wondering if I should update #687 to match.

delvinso:
> Looks good! Are these new columns the only ones that will be provided in variant reports? I'm wondering if I should update #687 to match.

Yes, they should be the only new ones added (for now 😅 )

cklamann (changes requested):
Looks good! I ran the alembic migration and click command locally and everything seems to be functioning as expected. A couple minor requests: 
- can you add the migration directory to the .gitignore? 
- can the migration directory go in the flask directory?
- can the artifacts (pickle, etc) be saved to the volume rather than the root directory, possibly as an archive with a timestamp in the filename? If we end up running more of these down the line, it would be good to know which artifacts came from where/when.

delvinso:
> Looks good! I ran the alembic migration and click command locally and everything seems to be functioning as expected. A couple minor requests:
> 
>     * can you add the migration directory to the .gitignore?
> 
>     * can the migration directory go in the flask directory?
> 
>     * can the artifacts (pickle, etc) be saved to the volume rather than the root directory, possibly as an archive with a timestamp in the filename? If we end up running more of these down the line, it would be good to know which artifacts came from where/when.

some notes 
- the migration folder has been moved into the flask folder, which is already mounted, meaning we longer need to separately mount the report folder
- since artifacts are saved in the mounted volume within the container, they are also saved on our local host
- tested locally, i end up with the same results as before (1115/1515 reports mapped and inserted)

cklamann (approved):

PR #698, opened 2021-06-30
React/Flask: Add support for multiplex files

### LINKED ISSUE #684: React/Flask: Add flag for files that can be attached to more than one dataset
- Some files contain multiplexed data that can be linked to several datasets.  We should allow users to indicate that their data is multiplexed during the file-linking process, and this information should be stored in the database so that that these files are not filtered out as "already linked." 
- This (potentially?) also changes the dataset-file relationship to many-to-many. Progress on this issue should address the need for a migration and updates to the ORM.

cklamann:
Closes #684 

Allow files marked `multiplexed` to be associated with multiple datasets

* migration to create new `file` table and linking table
  * migration does not drop old table (for now), so switching between branches should be ok
  * however... it seems we have a few migrations in this sprint, so beware that the 'head' id stored in the `alembic_versions` table might become incomprehensible to  the CLI if it refers to a migration in another branch.
* update ORM mappings to use new relationships
* update `_bulk` and `[PATCH] datasets` endpoint validation, insert, and delete logic
* update/add tests
* update unlinked files endpoint to return files as JSON (pseudo file models, basically)
* update FE file linking component and dependencies to use new file model
* add toggle feature to selected files, so that the popover wrapping the selected filename allows user to toggle on/off `multiplexed` field
* Allow for filtering by file path on dataset table

### DIFF
--- a/flask/app/buckets.py
+++ b/flask/app/buckets.py
@@ -55,21 +55,23 @@ def get_unlinked_files():
 
     files = []
 
-    linked_files = [
-        f.path
+    linked_files = {
+        f.path: True
         for f in models.File.query.filter(
             or_(models.File.multiplexed == None, models.File.multiplexed == False)
         ).all()
-    ]
+    }
 
-    linkable_files = [
-        f.path for f in models.File.query.filter(models.File.multiplexed == True).all()
-    ]
+    linkable_files = {
+        f.path: True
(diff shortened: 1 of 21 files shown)


frewmack (changes requested):

frewmack (approved):
LGTM!

PR #700, opened 2021-07-05
Flask: Add gene alias column to export

### LINKED ISSUE #686: Flask: Include gene aliases in gene pull export

cklamann:
Closes #686 

- Contains a migration for adding an index to alias table. It shares a `head` with other migrations in this sprint, so a rollback might be necessary when switching branches.

### DIFF
--- a/flask/app/variants.py
+++ b/flask/app/variants.py
@@ -4,8 +4,8 @@ from typing import Any, List
 from flask import Blueprint, Response, abort, current_app as app, jsonify, request
 from flask_login import current_user, login_required
 from sqlalchemy import distinct, func
-from sqlalchemy.orm import contains_eager, selectinload
-from sqlalchemy.sql import and_, or_
+from sqlalchemy.orm import contains_eager
+from sqlalchemy.sql import and_
 
 import pandas as pd
 from . import models
--- a/flask/app/variants.py
+++ b/flask/app/variants.py
@@ -3,14 +3,13 @@ from typing import Any, List
 
 from flask import Blueprint, Response, abort, current_app as app, jsonify, request
 from flask_login import current_user, login_required
+import pandas as pd
 from sqlalchemy import distinct, func
-from sqlalchemy.orm import contains_eager
+from sqlalchemy.orm import aliased, contains_eager
 from sqlalchemy.sql import and_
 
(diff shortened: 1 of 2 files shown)


flask/migrations/versions/184f5edd4719_add_index_to_gene_alias_ensembl_id.py
```diff
@@ -0,0 +1,24 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.create_index("gene_alias_ensembl_id_IDX", "gene_alias", ["ensembl_id"])
+
+
+def downgrade():
+    op.drop_index("gene_alias_ensembl_id_IDX")
```
delvinso, line 24: missing the table to drop the index from
```suggestion
    op.drop_index("gene_alias_ensembl_id_IDX", "gene_alias")
```

flask/app/variants.py
```diff
@@ -185,7 +186,12 @@ def summary(type: str):
 
     ensgs = parse_gene_panel()
 
-    # app.logger.debug(ensgs)
+    alias_subquery = aliased(
```
delvinso, line 189: Cool solution to an ensembl id not always having an HGNC gene symbol! I think it'd be helpful if the rationale for this approach was explained, maybe in the docstrings or a comment here

flask/app/variants.py
```diff
@@ -196,7 +202,8 @@ def summary(type: str):
             .contains_eager(models.Analysis.datasets)
             .contains_eager(models.Dataset.tissue_sample)
             .contains_eager(models.TissueSample.participant)
-            .contains_eager(models.Participant.family)
+            .contains_eager(models.Participant.family),
+            contains_eager(models.Gene.aliases.of_type(alias_subquery)),
```
delvinso, line 206: The common name should be serialized for the variant endpoint, around lines ~255

delvinso (changes requested):
Overall looks good to me except for a few minor things. I played around with keeping the usual eager loading with a filter on the alias table but seems like your solution is the best way keeping in mind that an ensembl id may not always have an hgnc gene symbol.

delvinso (approved):
👍 🚀

Issue #699, opened 2021-06-30
React/Flask: Variant export UX improvements, pt.2


cklamann:
- Allow users to select/deselect columns included in report and save preferences
- Allow users to search for variants by region or rsID

PR #717, opened 2021-07-13
Flask: Search for Variants by position, region; select columns

frewmack:
Part of #699, Blocking #687 

Updates `/api/summary/variants`, `/api/summary/participants` to accept region and position. 
- ~~Adds `search_by` parameter for specifying how to interpret the `panel` parameter.~~
- Removes `panel` parameter, instead adds parameters `genes`, `regions`, `positions` for querying by different criteria (only 1 may be specified at a time).

Adds `columns` parameter to allow specifying specific columns to include in CSV report. Defaults to including all columns if not specified.


flask/app/variants.py
```diff
@@ -358,6 +471,21 @@ def summary(type: str):
...
+            # filter out faulty columns
+            valid_columns = {col for col in agg_df.columns.values}
+            fixed_columns = [col for col in columns if col in valid_columns]
+            if len(fixed_columns) < len(columns):
+                app.logger.warn(
+                    "Ignoring invalid columns from request: {}".format(
+                        set(columns) ^ set(fixed_columns)
+                    )
+                )
+            agg_df = agg_df.loc[:, fixed_columns]
```
cklamann, line 487: This looks good for the csv, but there's also the question of the JSON representation (which I don't think is being used anywhere at the moment). Is it possible to move the filter into the SQL/ORM as a select statement so that only the requested columns are returned from the database? I'm not sure how this would impact current functionality, esp. re: Pandas, but in theory it might take care of the csv and JSON responses in one go.

flask/app/variants.py
```diff
@@ -227,17 +226,109 @@ def parse_gene_panel() -> List[Any]:
...
+    return position_filter
 
 
 @variants_blueprint.route("/api/summary/<string:type>", methods=["GET"])
 @login_required
 def summary(type: str):
     """
-    GET /api/summary/participants\?panel=ENSG00000138131
-    GET /api/summary/variants\?panel=ENSG00000138131
+    GET /api/summary/participants\?panel=ENSG00000138131&search_by=gene
```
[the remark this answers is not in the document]
frewmack: I was originally going to go with this, but decided on the current design to indicate that you can only search by one category at a time. I'll go ahead with this change though because I agree that it looks more intuitive

flask/app/variants.py
```diff
@@ -227,17 +226,109 @@ def parse_gene_panel() -> List[Any]:
         app.logger.error("Not all requested genes were found.")
         abort(400, description="Not all requested genes were found.")
 
-    genes = query.all()
+    genes_list = query.all()  # TODO: is this line necessary?
```
delvinso, line 229: Doesn't look like it's being used downstream at all

flask/app/variants.py
```diff
@@ -227,17 +226,109 @@ def parse_gene_panel() -> List[Any]:
...
+    Parses a given list of ranges (eg. "chr1:0500-0509,chr5:0600-0630,chr18:0440-0330").
+    Returns a filter for use against the variant table to find corresponding variants
+    based on gene start and end positions.
+    """
+    CHR, START, END = 0, 1, 2
+    # format check
+    # 1: chromosome, 2: start position, 3: end position
+    region_pattern = re.compile("chr([\da-zA-Z]+):([\d]+)-([\d]+)")
+    matches = [region_pattern.match(r) for r in region.split(",")]
+
```
delvinso, line 245: This isn't necessary, but I think input validation such that the requested end position is > start position would be nice.

flask/app/variants.py
```diff
@@ -227,17 +226,109 @@ def parse_gene_panel() -> List[Any]:
...
+            )
+            for tup in position_set
+        ]
+    )
+
+    query = models.Variant.query.filter(position_filter)
+    found_variants = query.with_entities(
+        func.count(distinct(models.Variant.variant_id))
+    ).scalar()
+    if found_variants == 0:
```
delvinso, line 318: If no variants are returned, can we prevent the subsequent query (line 381 onwards;which seems to be expensive, at least for region search) from being fired off? There could be a better way but I was thinking of something like returning `found_variants` along with the filter and then performing the check with an abort in the endpoint.

flask/app/variants.py
delvinso: Interested to see how this goes as I believe duplicate columns end up being selected as a result from the eager loading

delvinso (commented):
Looks great so far! Functionality is there and the pattern matching is elegant - just some minor requests

flask/app/variants.py
frewmack: The parsing function is already agnostic to the start-end order. Lines 253-257 correct the order if the start position is higher than the end position.

delvinso: Ah I see, in that case then eg. `http://localhost:5000/api/summary/participants?panel=chr10:100028007-100007446&search_by=region` and `http://localhost:5000/api/summary/participants?panel=chr10:100007446-100028007&search_by=region` should return the same results? If so I get 0 results back on the former

flask/app/variants.py
frewmack: The way the JSON result is formatted for the participant-wise report also adds to the complexity. Currently hoping to find a way to get a set of columns that are present in the query, and then only accept columns that are a subset of those columns, similar to the CSV case.

flask/app/variants.py
delvinso: It looks like the set itself isn't being modified with the corrected order, so the set that's iterated over to generate the filters isn't actually corrected. A solution might be to convert the set to a list and index into the list for assignment eg. `for i, r in enumerate(region_list)`

flask/app/variants.py
```diff
@@ -227,17 +226,108 @@ def parse_gene_panel() -> List[Any]:
...
+
+    app.logger.info("Requested region: %s", region)
+
+    if not all(match is not None and len(match.groups()) == 3 for match in matches):
+        app.logger.error("Invalid region format: %s", region)
+        abort(400, description="Invalid region format")
+
+    # prepare set of regions
+    region_set = set([match.groups() for match in matches])
+    for r in region_set:
```
delvinso, line 254: It looks like the set itself isn't being modified with the corrected order, so the set that's iterated over to generate the filters isn't actually corrected. A solution might be to convert the set to a list and index into the list for assignment eg. for i, r in enumerate(region_list)

flask/app/variants.py
frewmack: Not sure how to proceed with the column selection issue. Currently, I have a way to generate a set of model entities using the requested list of column names, however I've tried a couple of different approaches for reducing the query down to the requested columns:

- `query.options(load_only(...))` allows you to only load a specific subset of columns/entities, and will "defer" loading for any other entities. This doesn't really address the issue, since it effectively just delays loading other columns which we don't want to load to begin with.
- `db.session.query(...)` is how you specify what to SELECT. However, this seems to cause issues with relationship properties (in particular, `Variant.genotype`). This is the approach that is most similar to what Conor suggested, but I haven't yet found a way to make it work due to this relationship property issue.
- `query.with_entities(...)` allows you to return a subset of the columns selected as a list of tuples in the same order as they're provided to the function. This approach seems the most like what we're looking for, however it means that we can no longer use `asdict` to convert the query output, since the output is now a list of tuples and not a list of gene/variant class instances. It also doesn't seem to play nicely with pandas, but I haven't looked into that aspect yet.

Going to work on the frontend side of this issue for a bit and come back to this later.

flask/app/variants.py
```diff
@@ -228,17 +227,136 @@ def parse_gene_panel() -> List[Any]:
...
+
+    app.logger.info("Requested region: %s", region)
+
+    if not all(match is not None and len(match.groups()) == 3 for match in matches):
+        app.logger.error("Invalid region format: %s", region)
+        abort(400, description="Invalid region format")
+
+    # prepare set of regions
+    region_list = []
+    for i, r in enumerate(set([match.groups() for match in matches])):
```
delvinso, line 257: my initial thought of using enumerate was to loop through the items in the list with a counter and modifying the list in place, but this works too! how you currently have it means we can omit the enumerate function along with the `i` counter

flask/app/variants.py
```diff
@@ -369,6 +519,21 @@ def summary(type: str):
...
+            # filter out faulty columns
+            valid_columns = {col for col in agg_df.columns.values}
+            fixed_columns = [col for col in columns if col in valid_columns]
+            if len(fixed_columns) < len(columns):
+                app.logger.warn(
+                    "Ignoring invalid columns from request: {}".format(
+                        set(columns) ^ set(fixed_columns)
+                    )
+                )
+            agg_df = agg_df.loc[:, fixed_columns]
```
delvinso, line 535: the order the columns are specified in the query params will determine the order of the columns in the csv, overriding the order specified by `relevant_cols`. I think, if this is not accounted for in the front end, the order of the columns should be preserved in the report and possibly fixed, at least for chromosome, position, ref allele, and alt allele. thoughts?

frewmack: Should be fixed now. `fixed_columns` is now generated from `relevant_cols` while being compared to the requested columns and `valid_columns`.

flask/app/variants.py
```diff
@@ -1,12 +1,13 @@
...
+import re
 from typing import Any, List
 
 from flask import Blueprint, Response, abort, current_app as app, jsonify, request
 from flask_login import current_user, login_required
 import pandas as pd
 from sqlalchemy import distinct, func
-from sqlalchemy.orm import aliased, contains_eager
-from sqlalchemy.sql import and_
+from sqlalchemy.orm import aliased, contains_eager, load_only
```
delvinso, line 9: unused `load_only` import

flask/app/variants.py
```diff
@@ -228,17 +227,151 @@ def parse_gene_panel() -> List[Any]:
...
+                models.Variant.position == tup[POS],
+            )
+            for tup in position_set
+        ]
+    )
+
+    return position_filter
+
+
+def parse_rsid(rsid: str):
```
delvinso, line 314: Not an issue with the PR itself but unfortunately a query with a single rsid tied to >1000 variants takes over 10 minutes to execute for a ptp-wise report since it's a text search. I don't see a way around this without decomposing the rsid column into a 1:n table. At the same time this does bring us back to the problem where we might be re-inventing the creation of the reports.

flask/app/variants.py
```diff
@@ -228,17 +227,151 @@ def parse_gene_panel() -> List[Any]:
...
+        return parse_position(param)
+    if search_type == "rsids":
+        return parse_rsid(param)
+
+    # This shouldn't happen
+    app.logger.error("Invalid search type '%s'", search_type)
+    abort(400, description="Invalid parameter")
+
+
+def jsonify_variant_wise(tup, columns):
```
delvinso, line 344: for book keeping purposes, do you mind leaving a comment that it's unused as i don't see this being referenced, please ignore this if it's already mentioned somewhere else, eg. design docs!

delvinso (commented):
As if this pr isn't already massive enough, could you add an index to `variant.position` in the models and the resulting alembic migration? It speeds up the position based query up by ~90% (from minutes to seconds) on my end. If not I think it would be easy enough (and maybe cleaner) to implement in a separate pr. Asides from that and the two minor changes, this looks amazing and good to go on my end 👍

flask/app/variants.py
frewmack: This function is left over from trying to implement column selection in JSON format. It's unused and I don't think it even works, so I'll just remove it.

frewmack:
> 
> 
> As if this pr isn't already massive enough, could you add an index to `variant.position` in the models and the resulting alembic migration? It speeds up the position based query up by ~90% (from minutes to seconds) on my end. If not I think it would be easy enough (and maybe cleaner) to implement in a separate pr. Asides from that and the two minor changes, this looks amazing and good to go on my end 👍

I think adding the index for `Variant.position` would be best saved for a separate PR. Maybe the performance issue with retrieving rsIDs could be included as well?

delvinso (approved):

PR #722, opened 2021-07-14
Flask/Docker: minor fixes and changes

delvinso:
- correct the datetime import in `manage.py` as an alternate import of the datetime package was breaking the dev db seed 
- remove unused mount in `docker-compose.yaml`


cklamann (approved):
LGTM!

PR #723, opened 2021-07-14
Changes in styling for Login page

hannahle:
- Changed font, color theme, and layout of Login Page. 
- Works with both dark / light mode

**Further improvements to be made:**
- [ ] Create design system with proper fonts and color themes in App.tsx to ensure consistent font and coloring


frewmack (changes requested):
Wow this looks great! Really professional! My only concern right now is responsiveness: it looks like the login form begins to go off-screen at low resolutions (~700px wide). Could we have the form centered on-screen and the double-helix graphic hidden at lower window sizes? MUI [Breakpoints ](https://material-ui.com/customization/breakpoints/) would probably be helpful for this as well as Grid components for the form.

frewmack (approved):
Looks good to me! 🚀

cklamann (approved):

PR #732, opened 2021-07-19
React: Create loading indicator for SearchVariants page

### LINKED ISSUE #726: React: Add loading indicator to report download page
Sometimes reports can take a little while to generate, and we'd like to let users know that their request is still being processed while they wait with a modal or other loading indicator. Since material-table uses the material-ui spinner, we should probably do the same for the sake of uniformity. One interesting option is the ["interactive integration"](https://material-ui.com/components/progress/#interactive-integration) variant. Once the behavior and design are settled, we'll probably want to use it elsewhere in the app (such as the Add Datasets page or the user/group update sections). For the SearchVariants page, it will probably require mounting the indicator as part of the [csv download function](https://github.com/ccmbioinfo/stager/blob/master/react/src/SearchVariants/components/SearchVariantsPage.tsx#L86) and unmounting it using the [onError](https://github.com/ccmbioinfo/stager/blob/master/react/src/SearchVariants/components/SearchVariantsPage.tsx#L68-L84)/onSuccess callbacks in the react-query hooks.

hannahle:
Closes #726

### DIFF
--- a/react/src/SearchVariants/components/SearchVariantsPage.tsx
+++ b/react/src/SearchVariants/components/SearchVariantsPage.tsx
@@ -38,8 +38,8 @@ const useStyles = makeStyles(theme => ({
         paddingRight: theme.spacing(1),
     },
     wrapper: {
-        margin: theme.spacing(1),
-        minWidth: 166,
+        margin: theme.spacing(1, 0.5, 1, 1),
+        // minWidth: 166,
         justifyContent: "space-between",
         position: "relative",
         display: "flex",
@@ -48,7 +48,11 @@ const useStyles = makeStyles(theme => ({
     buttonProgress: {
         color: green[500],
         position: "absolute",
-        marginLeft: 50,
+        // marginLeft: 50,
+        top: "50%",
+        left: "50%",
+        marginLeft: -12,
+        marginTop: -12,
     },
 }));
 
--- a/react/src/SearchVariants/components/SearchVariantsPage.tsx
+++ b/react/src/SearchVariants/components/SearchVariantsPage.tsx
@@ -1,25 +1,24 @@
 import React, { useEffect, useState } from "react";
-import { useQueryClient } from 'react-query';
-
-import clsx from "clsx";
 import {
     Button,
     Chip,
+    CircularProgress,
     Container,
+    Fab,
     Grid,
     makeStyles,
     Typography,
-    CircularProgress,
-    Fab,
 } from "@material-ui/core";
-import { Check, CloudDownload } from "@material-ui/icons";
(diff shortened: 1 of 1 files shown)


react/src/SearchVariants/components/SearchVariantsPage.tsx
```diff
@@ -75,16 +122,35 @@ const SearchVariantsPage: React.FC<SearchVariantsPageProps> = () => {
...
     });
 
     const downloadParticipantwiseCsv = useDownloadCsv(GET_VARIANTS_BY_PARTICIPANTS_SUMMARY_URL, {
         onError,
+        onSuccess,
     });
 
     const downloadCsv = () => {
         const panel = selectedGenes.map(gene => `ENSG${gene.ensembl_id}`).join(",");
+        const key: QueryKey = [
```
cklamann, line 142: This is a nice solution to the problem of a lack of cache-access callbacks! Could you add a brief comment explaining its purpose, in case someone comes to the code later and doesn't remember the specific problem it solves?

cklamann (changes requested):
This is a nice implementation of the interactive integration pattern! Looking at it in action, I think that a simpler version might work best for us, since users might interpret the persistent 'success' state of the button as having semantic value. Do you mind dropping the `FAB` and the success state/indicators and seeing how it looks?

react/src/SearchVariants/components/SearchVariantsPage.tsx
```diff
@@ -26,6 +37,19 @@ const useStyles = makeStyles(theme => ({
...
+        justifyContent: "space-between",
+        position: "relative",
+        display: "flex",
+        alignItems: "center",
+    },
+    buttonProgress: {
+        color: green[500],
+        position: "absolute",
+        marginLeft: 50,
+    },
```
frewmack, line 52: The download button appears to have a wide space from the side margin. 

image

I figured out how to fix it while maintaining the circular progress overlay, based on https://material-ui.com/components/progress/#interactive-integration.

```suggestion
    wrapper: {
        margin: theme.spacing(1, 0.5, 1, 1),
        // minWidth: 166,
        justifyContent: "space-between",
        position: "relative",
        display: "flex",
        alignItems: "center",
    },
    buttonProgress: {
        color: green[500],
        position: "absolute",
        // marginLeft: 50,
        top: "50%",
        left: "50%",
        marginLeft: -12,
        marginTop: -12,
    },
```

frewmack (changes requested):
Some styling suggestions but otherwise looks good!

frewmack (approved):
LGTM! 🚀

cklamann (approved):

Issue #752, opened 2021-07-28
Chore: Prepare for MUI 5.0.


hannahle:
In anticipation of Material-UI 5.0., maintainers have added deprecation warnings to latest 4.x releases. We need to update our code so that warnings are cleared. Migration guide is here: https://next.material-ui.com/guides/migration-v4/.

PR #755, opened 2021-07-28
React: Remove deprecated material-ui code

hannahle:
Resolved Material-UI related deprecation warnings

### DIFF
--- a/react/src/components/ListItemRouterLink.tsx
+++ b/react/src/components/ListItemRouterLink.tsx
@@ -23,7 +23,7 @@ export default function ListItemRouterLink(props: ListItemRouterLinkProps) {
                     {...linkProps}
                     onMouseEnter={() => setOpenTooltip(true)}
                     onMouseLeave={() => setOpenTooltip(false)}
-                ></Link>
+                />
             )),
         [props.to]
     );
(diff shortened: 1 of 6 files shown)


react/src/components/ListItemRouterLink.tsx
```diff
@@ -23,7 +23,7 @@ export default function ListItemRouterLink(props: ListItemRouterLinkProps) {
                     {...linkProps}
                     onMouseEnter={() => setOpenTooltip(true)}
                     onMouseLeave={() => setOpenTooltip(false)}
-                />
+                ></Link>
```
cklamann, line 26: Is this change necessary?

hannahle: I think there was a deprecation warning that requires Link to have children nodes. It's totally fine if we remove it though since we don't have any children to begin with!

cklamann: OK sounds good -- the children should be passed through the props, so hopefully it doesn't complain if you revert!

frewmack (commented):
Is this PR for issue #752? The migration guide seems to provide some [codemod scripts](https://next.material-ui.com/guides/migration-v4/#run-codemods) that we can run on the `/react` folder to automatically make changes:

```
npx @material-ui/codemod@next v5.0.0/preset-safe <folder>
npx @material-ui/codemod@next v5.0.0/variant-prop <folder>
npx @material-ui/codemod@next v5.0.0/link-underline-hover <folder>
```

frewmack (approved):
Looks good to me! No warnings on my end

PR #757, opened 2021-07-28
Flask/Keycloak: Enable token-based authentication

### LINKED ISSUE #735: Keycloak/Flask: setup OAuth client credentials flow for accessing stager
- immediate purpose is to allow client applications (e.g., pipeline execution services) to post data (e.g., status updates) to Stager
- it seems that the most straightforward way of implementing this would be via the existing keycloak infrastructure. Each client would have its own grant. We'll need to research the best way to use keycloak's auth logic to protect Flask endpoints.

cklamann:
closes #735 

- ~creates a decorator that wraps `login_required` to enable a second check for a valid Bearer token. If the token is valid and contains the role stipulated by the decorator, the action is authorized.~
- uses request_loader callback to resolve user based on Bearer token in Authorization header if no session found
- ~sample implementation in [PATCH] analyses would allow a client (say, a pipeline execution service) to update analysis status.~
- ~unit tests provided~
- docs updated with instructions for how to add a client via the keycloak admin module


samuelkimuli:
Regarding client policies, how crucial is it to confirm to a required security standard like FAPI ? to make it easy to secure client applications. Take a look at this https://github.com/keycloak/keycloak-community/blob/master/design/client-policies.md

frewmack (commented):
Some small changes for now. I'm having issues using the `userinfo` endpoint with a `client_credentials` token, so I'll have a more thorough review once I figure this out

frewmack:
A comment on Keycloak roles: client roles may be more appropriate for our use-case than realm roles, since roles like "analysis-management" only have significance to Stager. They can be added by going to ccm-stager in Clients, then the Roles tab to add a new client role. Then other clients with service accounts (eg. pipeline execution service) can have client roles added in their Service Account Roles tab as shown below
image

I also think that using policies/permissions could be held off to if/when we decide to switch Stager authorization to Keycloak completely. The kind of fine-grain control that policies would provide would be more worthwhile if we could apply that access control scheme to all Stager users. For now, I think roles are sufficient.

delvinso:
here is a possible, less hackier alternative, to injecting the user model into the controller: 
- use a request_loader (https://flask-login.readthedocs.io/en/latest/#login-using-authorization-header) which seems to be designed for multiple authentication. as a result, both the typical login and token based authorization will then occur on any route decorated with `login_required` 
- this may require the `User()` model to be inherited and potentially a method for performing the ‘sub’ check, or/property for the ‘sub’ check to occur on

delvinso (commented):
here is a possible, less hackier alternative, to injecting the user model into the controller:

   - use a request_loader (https://flask-login.readthedocs.io/en/latest/#login-using-authorization-header) which seems to be designed for multiple authentication. as a result, both the typical login and token based authorization will then occur on any route decorated with `login_required`
 - this may require modifications on the current user model (oops of course the user class is already inherited) ~~either a method for performing the 'sub' check, or property to be added `User()` model to be inherited and potentially for either a method to be added for performing the ‘sub’ check, or/property added for the ‘sub’ check to occur on~~

cklamann:
> * use a request_loader (https://flask-login.readthedocs.io/en/latest/#login-using-authorization-header) which seems to be designed for multiple authentication. as a result, both the typical login and token based authorization will then occur on any route decorated with `login_required`

This seems like a much cleaner way of retrieving the user model from the request. My only concern is that the `role` attribute that we use to validate the user's access to the route is actually not part of the Stager User model, which would prevent us from writing a role-based decorator like `@authorize('some-keycloak-role-name`)` to selectively open routes to client-users. For this to work, I think, we'd need to move to a role-based auth system on the Stager side and sync it with the keycloak roles so the Stager-side user model mirrors what's in keycloak. This seems like a pretty disruptive change. Another option (besides the current one) might be to drop keycloak-role-based authorization altogether and instead assign the token-user to a Stager group and treat them like any other user. In other words, the token-user wouldn't be restricted to a whitelist of endpoints but could hit all the same routes on the backend that a regular user with that group membership has.

PR #787, opened 2021-08-23
Flask: Consolidate user resolution and datasets access code

cklamann:
- consolidate user resolution logic to function `get_current_user` in `utils.py`
- consolidate dataset-group permissions logic to function `filter_datasets_by_user_groups` in `utils.py`
- update tests

### DIFF
--- a/flask/app/utils.py
+++ b/flask/app/utils.py
@@ -18,11 +18,12 @@ from flask.json import JSONEncoder
 from flask_login import current_user
 from flask_sqlalchemy import Model
 from sqlalchemy import exc
+from sqlalchemy.orm.query import Query
 from werkzeug.exceptions import HTTPException
 
 from .extensions import db
 from .madmin import MinioAdmin
-from .models import User
+from .models import User, Group, Dataset
 
 
 def handle_error(e):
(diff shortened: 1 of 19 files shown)


flask/app/utils.py
```diff
@@ -306,3 +307,46 @@ def find(collection: Iterable[Mapping[str, Any]], pred: Callable):
...
+    if app.config.get("LOGIN_DISABLED") or current_user.is_admin:
+        user_id = request.args.get("user")
+        if user_id:
+            user = User.query.filter(User.user_id == user_id).first()
+            if not user:
+                abort(400, description="Provided user does not exist!")
+
+    if not user:
+        user = current_user
+
```
delvinso, line 327: user lacks the `is_admin` property when we login anonymously (via setting LOGIN_DISABLED = True in the config), resulting in endpoints 500ing. From my testing this can be fixed by setting the property to be true if it doesn't exist

```
curl localhost:5000/api/families/1
{
  "error": "'AnonymousUserMixin' object has no attribute 'is_admin'"
}
```

flask/app/utils.py
```diff
@@ -306,3 +307,46 @@ def find(collection: Iterable[Mapping[str, Any]], pred: Callable):
...
+    app.logger.debug("user_id: '%s'", getattr(user, "user_id", None))
+
```
delvinso, line 329:
```suggestion
    if not hasattr(user, 'is_admin'):
        user.is_admin = False
```

PR #799, opened 2021-09-02
Fixed missing fields bug

hannahle:
- Added missing fields (specifically family_aliases and participant_aliases) to `"/api/analyses/<int:id>", methods=["GET"]`


cklamann (approved):
LGTM!

PR #805, opened 2021-09-09
Flask: add query params to perform exact search by codename in GET datasets/participants

### LINKED ISSUE #804: Flask: Unable to search by exact codename in GET participants and datasets
Add functionality to both endpoints so an exact match by either participant or family codename can be performed

delvinso:
closes #804

example : `curl 'localhost:5000/api/datasets?user=1&family_codename=200&family_codename_exact_match=True'`

- Family codenames for the test database have also been changed from 'A' and 'B' to 'Aa' and 'Bb' respectively so the `*_exact_match` query parameter can be tested

### DIFF
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -133,8 +133,7 @@ def list_datasets(page: int, limit: int) -> Response:
         )
         if participant_codename_exact_match:
             filters.append(
-                models.Participant.participant_codename
-                == func.binary(participant_codename)
+                models.Participant.participant_codename == participant_codename
             )
         else:
             filters.append(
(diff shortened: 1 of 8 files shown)


flask/app/datasets.py
```diff
@@ -120,13 +122,43 @@ def list_datasets(page: int, limit: int) -> Response:
...
+
     if family_codename:
-        filters.append(func.instr(models.Family.family_codename, family_codename))
+        app.logger.debug(
+            "family_codename_exact_match: {}".format(family_codename_exact_match)
+        )
+        if family_codename_exact_match:
+            filters.append(
+                models.Family.family_codename == func.binary(family_codename)
+            )
```
kevinlul, line 158: Why is `func.binary` used?

flask/app/datasets.py
```diff
@@ -120,13 +122,43 @@ def list_datasets(page: int, limit: int) -> Response:
         filters.append(models.Dataset.dataset_id == dataset_id)
 
     participant_codename = request.args.get("participant_codename", type=str)
+    participant_codename_exact_match = request.args.get(
+        "participant_codename_exact_match", type=str_to_bool, default=False
+    )
```
kevinlul, line 127: I think instead of the str_to_bool thing you might be able to simply check for the existence of the query string parameter and use that.

delvinso: Didn't think about that! I thought it would be more explicit this way but I'm open to changing it

kevinlul (approved):

PR #806, opened 2021-09-09
Flask/ React: Fixed delete bug

### LINKED ISSUE #797: Bug: Delete dataset
Currently the "delete" button on the datasets table isn't sending a request to the server. The front-end components should be so that the request goes through, and we should probably also confirm that associated models are handled appropriately (including analyses and their related models, such as `genotype` and `variant`). Currently, users are prevented from deleting datasets that have associated analyses. 

To be completed in this issue: 

- [x] Add delete request from client to backend and refresh table
- [x] Remove the corresponding row with the dataset ID from `group_datasets` table
- [x] Remove automatically tissue from `tissue_sample` table using the `tissue_sample_id` from `datasets`.

hannahle:
closes #797 
- Added request to backend
- Added delete functionality for the appropriate rows in `groups_datasets` and `tissue_sample` tables. 
- It seems like we're not sure whether we should delete the associated analysis yet, so I'm leaving it as is.

### DIFF
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -344,9 +344,15 @@ def delete_dataset(id: int):
     # When a row from the dataset table with dataset_id is deleted, we also want to cascade the deletion to the corresponding row from groups_datasets.
     dataset.groups = []
 
-    tissue_sample = models.TissueSample.query.filter(
-        models.TissueSample.tissue_sample_id == dataset.tissue_sample_id
-    ).first_or_404()
+    tissue_sample = (
+        models.TissueSample.query.filter(
+            models.TissueSample.tissue_sample_id == dataset.tissue_sample_id
+        )
+        .options(joinedload(models.TissueSample.datasets))
+        .first_or_404()
+    )
+
+    tissue_sample.datasets = []
 
     if not dataset.analyses:
         try:
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -356,6 +356,10 @@ def delete_dataset(id: int):
 
     if not dataset.analyses:
         try:
+            app.logger.debug('DATASET IS')
+            app.logger.debug(dataset)
+            app.logger.debug('TISSUE SAMPLE IS')
+            app.logger.debug(tissue_sample)
             db.session.delete(dataset)
             db.session.delete(tissue_sample)
             db.session.commit()
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -51,7 +51,10 @@ EDITABLE_COLUMNS = [
     "discriminator",
 ]
 
-datasets_blueprint = Blueprint("datasets", __name__,)
+datasets_blueprint = Blueprint(
+    "datasets",
+    __name__,
+)
 
 
 @datasets_blueprint.route("/api/datasets", methods=["GET"], strict_slashes=False)
@@ -483,7 +486,8 @@ def update_dataset_linked_files(
             # insert new files
             dataset.linked_files.append(
                 models.File(
-                    path=file_model["path"], multiplexed=file_model.get("multiplexed"),
+                    path=file_model["path"],
+                    multiplexed=file_model.get("multiplexed"),
                 )
             )
 
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -51,10 +51,7 @@ EDITABLE_COLUMNS = [
     "discriminator",
 ]
 
-datasets_blueprint = Blueprint(
-    "datasets",
-    __name__,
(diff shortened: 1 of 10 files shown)


flask/app/datasets.py
```diff
@@ -340,12 +340,30 @@ def delete_dataset(id: int):
...
+    # Remove permission groups with the corresponding dataset ID in groups_datasets table
+    dataset.groups = []
+
+    tissue_sample = (
+        models.TissueSample.query.filter(
+            models.TissueSample.tissue_sample_id == dataset.tissue_sample_id
+        )
+        .options(joinedload(models.TissueSample.datasets))
+        .first_or_404()
+    )
```
kevinlul, line 356: This should be retrieved by a joinedload on the dataset query. All the relationships are configured so the reverse side also exists, even if it's not in the class definition of the reverse side.

flask/app/datasets.py
```diff
@@ -340,12 +340,30 @@ def delete_dataset(id: int):
...
+
     if not dataset.analyses:
         try:
             db.session.delete(dataset)
             db.session.commit()
+
+            # Only delete tissue sample if it doesn't have any associated datasets left.
+            if len(tissue_sample.datasets) == 0:
```
kevinlul, line 364: Does this field update without the prior commit? That would be preferable as this operation is achievable in one transaction.

kevinlul (changes requested):
I'm not quite sure what's going on with the reformatting in unrelated files.

flask/app/datasets.py
hannahle: > I'm not quite sure what's going on with the reformatting in unrelated files.

I ran prettier --write on the react folder, which led to these changes. I probably should run it only on the files I change in the future!

kevinlul (changes requested):
Can you add a unit test to verify the new behaviour? The existing ones do not cover it completely.

hannahle:
> Can you add a unit test to verify the new behaviour? The existing ones do not cover it completely.

Sure!

flask/tests/test_datasets.py
```diff
@@ -210,10 +211,45 @@ def test_delete_dataset_admin(client, test_database, login_as):
...
+        .one_or_none()
+    )
+
+    participant_tissue_samples = [
+        s
+        for s in participant.tissue_samples
+        if s.tissue_sample_id == tissue_sample.tissue_sample_id
+    ]
+    assert (len(participant_tissue_samples)) == 0
+
```
kevinlul, line 252: I suggest making another test function instead to make the intent more clear. For the assertions, you only need to save the tissue_sample_id primary key before the request and then query for that after and assert its nonexistence

flask/tests/conftest.py
```diff
@@ -338,6 +338,26 @@ def test_database(client):
...
+
+    # Creating a tissue sample with only one dataset
+    sample_4 = TissueSample(
+        tissue_sample_id=4,
+        tissue_sample_type=TissueSampleType.Blood,
+        created_by_id=admin.user_id,
+        updated_by_id=admin.user_id,
+    )
+    sample_4.datasets.append(dataset_5)
+    participant_3.tissue_samples.append(sample_4)
```
kevinlul, line 359: Does dataset_5 belong to sample_3 or sample_4?

kevinlul (changes requested):

flask/tests/conftest.py
hannahle: Dataset 5 should belong to sample 4. The sample 3 line there is not really necessary. I will remove.

flask/tests/test_datasets.py
hannahle: Ahh I see. Didn't see that we have a GET request for tissue_sample also. I'll clean it up.

flask/tests/test_datasets.py
```diff
@@ -210,10 +211,35 @@ def test_delete_dataset_admin(client, test_database, login_as):
...
+    login_as("admin")
+
+    # Deleting a dataset without an analysis but associated with tissue sample
+    dataset_5 = (
+        models.Dataset.query.filter(models.Dataset.dataset_id == 5)
+        .options(
+            joinedload(models.Dataset.tissue_sample).joinedload(
+                models.TissueSample.datasets
+            ),
+        )
```
kevinlul, line 234: Last thing: you don't need this join here since you're only looking for `tissue_sample_id`, which is already a column in the dataset table.

hannahle: That's right!

kevinlul (approved):
LGTM 🚀

PR #818, opened 2021-09-15
React/Flask: Add RNAseq fields to dataset entry flow

### LINKED ISSUE #796: Update RNAseq fields
Stager currently has the option to enter RNAseq datasets on the AddDatasets page. But in order to enable this feature, the input fields need to be updated.  So far, it doesn't appear that any of the RNAseq-specific fields currently in the AddDatasets form are being collected by researchers, we might be able to remove them after confirming with users. Below is a list of fields that are currently being collected for RNAseq datasets and their equivalents in Stager:

- sample name: `participant_codename`
- family ID: `family_codename`
- RNA fastq location: `linked_files`
- tissue type: `tissue_sample_type`
- candidate genes (optional): currently not implemented, would require a new field, could reuse gene search autocomplete to fetch 
- condition: `participant_type` 

It seems that the minimal implementation would require:
1. removing unused fields from RNAseq form
2. creating a new `candidate_genes` field and displaying it conditionally for RNAseq data only

If significant work is needed to update the data entry form or refactoring is otherwise called for, it would be good to reference #781 and the patterns laid out in [this branch](https://github.com/ccmbioinfo/stager/tree/frewmack/data-entry-table-refactor)

cklamann:
closes #796
- refactor AddDataset FE components
  - lift column state management functionality to top-level component
  - allow for mixed RNA/DNA seq entries on the input table
  - create candidate gene lookup control
  - remove unused code
- update datasets table and relevant modals to show RNAseq data
- add new RNAseq columns to DB and update name of RNAseq pipeline in migration
- update endpoints to handle RNAseq and mixed input and filter fields
- update tests

### DIFF
--- a/flask/app/buckets.py
+++ b/flask/app/buckets.py
@@ -70,7 +70,6 @@ def get_unlinked_files():
         if linkable_files.get(file_name):
             files.append({"path": file_name, "multiplexed": True})
         elif not linked_files.get(file_name):
-            app.logger.debug(file_name)
             files.append({"path": file_name, "multiplexed": False})
 
     app.logger.debug("Returning JSON array..")
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -178,6 +178,12 @@ def list_datasets(page: int, limit: int) -> Response:
     if updated:
         filters.append(filter_updated_or_abort(models.Dataset.updated, updated))
 
+    candidate_genes = request.args.get("candidate_genes", type=str)
+    if candidate_genes:
+        filters.append(
+            func.instr(models.RNASeqDataset.candidate_genes, candidate_genes)
+        )
+
     user = get_current_user()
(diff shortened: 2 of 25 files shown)


kevinlul:
Okay, so far what I've found is that when RRS is selected as a dataset type in the submission form, a huge number of columns no longer appear, including ones related to participants!

cklamann:
> A huge number of columns no longer appear, including ones related to participants!

The intention was to show only columns indicated by researchers as pertinent to RNAseq. But if we need to add more fields it should be pretty straightforward to do so by updating the classes [here](https://github.com/ccmbioinfo/stager/blob/cklamann/update-rna-seq/react/src/typings.tsx#L190-L227).

kevinlul (approved):
Sorry for taking so long. This looks good to me then!

PR #821, opened 2021-09-17
React: Added aggregate notes popover on Analysis page

hannahle:
- There was no way to see notes from other tabs (such as Participants or Datasets) on the Analysis page. 
- I added an Info icon in the `actions` property of the table and attached a popover that displays the notes (if there are any). 
- Modified the `get_analyses` code in `analyses.py` to also return notes from the participants tab.

### DIFF
--- a/react/src/Analyses/components/AnalysisNotes.tsx
+++ b/react/src/Analyses/components/AnalysisNotes.tsx
@@ -53,10 +53,12 @@ export default function AnalysisNotes(props: AnalysisNotesProps) {
                     {analysisWithNotes.map(dataset => (
                         <>
                             <Divider className={classes.divider} />
-                            <Typography className={classes.comment}>
-                                Comment by {dataset.updated_by} for dataset{" "}
-                                {dataset.participant_codename}/{dataset.tissue_sample_type}/
-                                {dataset.dataset_type}
+                            <Typography className={classes.comment} gutterBottom>
+                                Last updated {dataset.updated} by {dataset.updated_by}
+                            </Typography>
+                            <Typography variant="button">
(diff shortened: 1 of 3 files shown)


react/src/Analyses/components/AnalysisNotes.tsx
```diff
@@ -0,0 +1,80 @@
...
+            }}
+        >
+            {analysisWithNotes.length > 0 ? (
+                <>
+                    <Typography variant="h6">Notes</Typography>
+                    {analysisWithNotes.map(dataset => (
+                        <>
+                            <Divider className={classes.divider} />
+                            <Typography className={classes.comment}>
+                                Comment by {dataset.updated_by} at {dataset.updated}
```
kevinlul, line 57: This could be misleading as the updated timestamp is for any field.

react/src/Analyses/components/AnalysisNotes.tsx
```diff
@@ -0,0 +1,80 @@
...
+                            </Typography>
+                            <Typography variant="subtitle1">
+                                Dataset {dataset.dataset_id}
```
kevinlul, line 60: The primary key is actually not very helpful to a person. We have done something like `family_codename/participant_codename/dataset_type` for the dataset delete button to be more identifying.

kevinlul (changes requested):

react/src/Analyses/components/AnalysisNotes.tsx
kevinlul: The same is true of the updated_by field, so I think it's easiest to just label them for what they are: Last updated WHEN why WHO

kevinlul (approved):
LGTM!

PR #822, opened 2021-09-17
React/Flask: Added sequencing ID field in Dataset table

jennyziyi-xu:
Added a column `sequencing_id` to the dataset table. The column can be sorted and filtered. Added test cases in test_datasets.py. Added field `sequencing_id` to the global test cases in conftest.py. 

Comment: Currently, only 4 datasets for pytest are documented in flask.md, should we update flask.md? There are 5 datasets in conftest.py.

### DIFF
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -124,7 +124,6 @@ def list_datasets(page: int, limit: int) -> Response:
     if dataset_id:
         filters.append(models.Dataset.dataset_id == dataset_id)
 
-    # sequencing ID.
     sequencing_id = request.args.get("sequencing_id", type=str)
     if sequencing_id:
         filters.append(func.instr(models.Dataset.sequencing_id, sequencing_id))
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -77,6 +77,7 @@ def list_datasets(page: int, limit: int) -> Response:
         "participant_codename",
         "family_codename",
         "dataset_id",
+        "sequencing_id",
     ]
     if order_by is None:
         order = None  # system default, likely dataset_id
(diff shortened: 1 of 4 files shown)


hannahle:
> Comment: Currently, only 4 datasets for pytest are documented in flask.md, should we update flask.md? There are 5 datasets in conftest.py.

This is because I updated conftest.py recently. I will update the flask.md. Thanks for pointing that out!

flask/app/datasets.py
```diff
@@ -94,6 +95,8 @@ def list_datasets(page: int, limit: int) -> Response:
         # Since this is an elif clause, we know special cases are already handled above.
         # order_by has been restricted to a known list, so we can safely use getattr
         order = getattr(models.Dataset, order_by)
+    elif order_by == "sequencing_id":
+        order = models.Dataset.sequencing_id
```
kevinlul, line 99: This is a field of the dataset table, so it is already handled by the prior elif by being included in the list and does not need to be explicitly done here.

flask/tests/conftest.py
```diff
@@ -218,6 +218,7 @@ def test_database(client):
         condition=DatasetCondition.Somatic,
         created_by_id=admin.user_id,
         updated_by_id=admin.user_id,
+        sequencing_id=5,
```
kevinlul, line 221: This column is a string, so it shouldn't be an int here

kevinlul (changes requested):
Please enter a pull request title instead of the automatically-generated one!

kevinlul:
> Please enter a pull request title instead of the automatically-generated one!

jennyziyi-xu:
> Please enter a pull request title instead of the automatically-generated one!

Done. Sorry, was a bit confused.

kevinlul (approved):
LGTM!

PR #825, opened 2021-09-20
React: Exact search toggle for family and participant codenames

hannahle:
- material-table stores an internal query object in its state (source code [here](https://github.com/mbrn/material-table/blob/42d966eacdbe47cd176f9c5c68e3d97b126b489e/src/material-table.js#L28)). This query object is internally updated by material-table whenever table properties are changed, such as when the user is filtering. 
- The approach I took was to add an additional property to this query object called `searchType` to programmatically enable exact match search when user toggles a switch. When the state of the switch changes, `onQueryChange()` is called on the table ref to requery.

### DIFF
--- a/react/src/components/ExactMatchFilterToggle.tsx
+++ b/react/src/components/ExactMatchFilterToggle.tsx
@@ -1,0 +1,70 @@
+import React, { useEffect, useState } from "react";
+import { Column } from "@material-table/core";
+import {
+    FormGroup,
+    FormHelperText,
+    InputAdornment,
+    makeStyles,
+    Switch,
+    TextField,
+} from "@material-ui/core";
+import { FilterList } from "@material-ui/icons";
+import { updateTableFilter } from "../functions";
+import { Dataset } from "../typings";
+
+const useStyles = makeStyles(theme => ({
+    root: {
+        marginBottom: theme.spacing(1.5),
+    },
+}));
+
+interface ExactMatchFilterToggleProps {
+    MTRef: React.MutableRefObject<any>;
+    columnDef: Column<Dataset>;
+    onFilterChanged: (rowId: string, value: any) => void;
+}
+
+export default function ExactMatchFilterToggle(props: ExactMatchFilterToggleProps) {
+    const classes = useStyles();
+    const [exactMatch, setExactMatch] = useState<boolean>(false);
+
+    useEffect(() => {
+        updateTableFilter(
+            props.MTRef,
+            `${props.columnDef.field}_exact_match`,
+            exactMatch.toString()
+        );
+    }, [props.MTRef, props.columnDef.field, exactMatch]);
+
+    return (
+        <>
+            <FormGroup className={classes.root}>
+                <FormHelperText>Exact match</FormHelperText>
+                <Switch size="small" onChange={event => setExactMatch(event.target.checked)} />
+            </FormGroup>
+            <TextField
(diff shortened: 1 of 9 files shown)


kevinlul (commented):
I find that the UI is really bulky. This makes the filter control for these columns take up much more vertical space than the rest and it doesn't fit in as well as the chevron for the date filter.

hannahle:
> I find that the UI is really bulky. This makes the filter control for these columns take up much more vertical space than the rest and it doesn't fit in as well as the chevron for the date filter.

We can also put the toggle inline with the text field. Any thoughts? Maybe something like this: 
Screen Shot 2021-09-20 at 2 34 59 PM

react/src/functions.tsx
```diff
@@ -389,6 +390,24 @@ export const updateFiltersAndRequery = (tableRef: React.MutableRefObject<any>) =
...
+ */
+
+export const updateSearchTypeAndRequery = (
+    tableRef: React.MutableRefObject<any>,
+    searchTypeOptions: SearchType
+) => {
+    const oldQuery = tableRef.current.state.query;
+    if (oldQuery.searchType) {
+        tableRef.current.state.query.searchType.push(searchTypeOptions);
+    }
```
kevinlul, line 406: This doesn't do anything because you are overwriting the property in the following line.

react/src/components/ExactMatchFilterToggle.tsx
```diff
@@ -0,0 +1,67 @@
...
+            updateSearchTypeAndRequery(props.MTRef, {
+                column: props.columnDef.field,
+                exact: exactMatch,
+            });
+        }
+    }, [props.MTRef, props.columnDef.field, exactMatch]);
+
+    return (
+        <TextField
+            autoFocus
```
kevinlul, line 41: This component is used multiple times on the page, so only the last one actually receives focus. I think we don't need this.

react/src/components/ExactMatchFilterToggle.tsx
```diff
@@ -0,0 +1,67 @@
...
+                ),
+                endAdornment: (
+                    <Tooltip title="Only show exact match">
+                        <Checkbox
+                            color="primary"
+                            size="small"
+                            onChange={event => setExactMatch(event.target.checked)}
+                        />
+                    </Tooltip>
+                ),
```
kevinlul, line 62: Make the Tooltip'd Checkbox the startAdornment and remove the endAdornment to remove clutter and match the appearance of the date filter.

react/src/components/ExactMatchFilterToggle.tsx
```diff
@@ -0,0 +1,67 @@
...
+    return (
+        <TextField
+            autoFocus
+            className={classes.input}
+            id="input-with-icon-textfield"
+            onChange={event => {
+                props.onFilterChanged((props.columnDef as any).tableData.id, event.target.value);
+            }}
+            InputProps={{
+                autoFocus: true,
```
kevinlul, line 48: Again, autofocus here is messy when multiple inputs request it.

kevinlul (changes requested):

react/src/components/ExactMatchFilterToggle.tsx
hannahle: Oh that's right. Forgot to delete this one. Will do - thanks!

kevinlul (approved):
LGTM!

PR #837, opened 2021-09-27
CI: rewrite frontend workflow to use separate environments for security

kevinlul:
I specifically added a branch protection rule to this branch to allow the deployment. When merged, we will need to change production to reverse proxy the new buckets and remove the old secrets and accounts.

For #835


cklamann (approved):
LGTM!

Issue #814, opened 2021-09-14
React/Flask: Fix Datasets editing modal


cklamann:
- [x] Currently the datasets modal will trigger a 500 on PATCH calls due to `library_prep_date` being an empty string by default. It should contain a valid date or `null`.
- [x] Currently the datasets modal allows users to edit non-dataset fields, which the datasets endpoint cannot update. We should  disable these fields so users can't try to update them.
- [x] The `dataset.discriminator` field is for internal purposes and shouldn't be shown to the user
- [x] For fields that have length constraint, uesrs cannot type in anymore when the limit is hit. 
- [x] Fields that have type and length constraints in the database should be validated before being sent back to the server (e.g. `read length`)
- [x] `library_prep_date`, if already present in the dataset model, will vanish in editing mode and, evidently, be replaced with an invalid display string, which will throw a 500 when the update request is sent back. We should make sure that, in editing view, the date is passed to the component in the valid format.

Issue #836, opened 2021-09-27
React: Implement drop down menu for LinkedFiles in Dataset modal


jennyziyi-xu:
In the `dataset` modal, implement a drop down menu for `LinkedFiles` field. Existing files can only be used once. We would not allow users to select files that are already in the `dataset` table. In other words, once the files are in the database, they can't be used. `FileLinkingComponent` can help with implementing this.

PR #843, opened 2021-09-29
React/Flask: Fix datasets editing modal (#814)

jennyziyi-xu:
Subtask 4: For the following fields, when the limit is hit, the user cannot type into the field anymore. Sequencing Center (string max len 100) 
Batch ID (string max len 50) 
Extraction Protocol (string max len 100) 
Capture Kit (string max len 50) 
Library Prep Method (string max len 50) 
Read length (string max len 10, length of the maximum integer)

( Notes (TEXT object, do not have a length limit).) 

Subtask 5: Implement type validation for `read_length` only, `read_length` must only contain digits, as other fields are either objects or can be alphanumeric. My only concern is if making the type validation in `onChange()` would slow down the rendering. I hypothesize that this is not an issue because `error` and `helperText` are props for TextField in Material UI, so they should be already optimized. 

Subtask 6: `onUpdate() ` is a function that we want to refactor in this sprint. I think that the way we want to send `library_prep_date` to the backend may be different after refactoring `onUpdate() `. The code added may or may not be useful in the future.

### DIFF
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -1,6 +1,5 @@
 from dataclasses import asdict
 from typing import List
-
 from flask import (
     Blueprint,
     Response,
@@ -9,11 +8,9 @@ from flask import (
     jsonify,
     request,
 )
-from sqlalchemy.sql.expression import update
 from flask_login import current_user, login_required
 from sqlalchemy import distinct, func
-from sqlalchemy.orm import contains_eager, joinedload, selectinload, with_polymorphic
-
+from sqlalchemy.orm import contains_eager, joinedload, selectinload
 from . import models
 from .extensions import db
 from .utils import (
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -9,6 +9,7 @@ from flask import (
     jsonify,
     request,
 )
+from sqlalchemy.sql.expression import update
 from flask_login import current_user, login_required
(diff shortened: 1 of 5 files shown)


kevinlul (approved):
Yeah that's all I have within the scope of this PR, since there's more work to do later

LGTM!

cklamann (changes requested):
There's a 500 that gets thrown when I try to update a dataset with linked files. It seems that this is due to the same issue that's causing problems with `library_prep_date` updates--the modal is sending back the view representation (array of file path strings) rather than the model (array of `File` objects). I think for the time being it might be best to just disable this field, since the user can use the in-table editing controls anyway, until we have the modal refactor done and can add the `FileLinkingComponent`.

jennyziyi-xu:
> There's a 500 that gets thrown when I try to update a dataset with linked files. It seems that this is due to the same issue that's causing problems with `library_prep_date` updates--the modal is sending back the view representation (array of file path strings) rather than the model (array of `File` objects). I think for the time being it might be best to just disable this field, since the user can use the in-table editing controls anyway, until we have the modal refactor done and can add the `FileLinkingComponent`.

Yes, good point. I am aware of this error and I think another ticket #836 addresses this issue. I will disable the field for now.

cklamann (approved):
:rocket:

PR #845, opened 2021-09-30
Flask: fix mixin column coverage

### LINKED ISSUE #844: Flask: Bug -- `mixin` iteration over table columns misses polymorphic attributes
The current `mixin` validation function discovers a model's fields by accessing the `__table__.columns` property. For polymorphic models like `RNASeqDataset`, this will cause issues, since some of its properties (e.g., `condition`) are actually on the `dataset` table and won't appear in the `RNASeqDataset` column list. An alternative might be to use a [mapper inspector](https://docs.sqlalchemy.org/en/14/orm/mapping_styles.html#runtime-introspection-of-mapped-classes-and-mappers), which has a `columns` attribute that contains all fields in the model. Issue can be reproduced by making a `PATCH` request to `datasets` with an rnaseq payload.

delvinso:
closes #844

- ~~for `inspect` to return a mapper (and therefore return all fields associated with polymorphic models), it must be called on a `Model` class. `mixin` is modified to accept said class and iterate through the columns~~ Conor has found out the mapper can in fact be accessed on an inspected model instance
- a test was added to verify enums can be modified for `RNASeqDataset` under the test database and also as a stand alone unit test 
- mixin should now throw a more helpful error message than `error: <field>` 
- small fix for PATCH /api/analyses/<int:id> where `enum_validate` was redundantly called before `mixin`. 

may split off into another PR - (try to) consolidate `mixin` and `enum_validate` as the logic is largely the same

### DIFF
--- a/flask/app/analyses.py
+++ b/flask/app/analyses.py
@@ -656,7 +656,7 @@ def update_analysis(id: int):
             else:
                 abort(400, description="Assignee not found")
 
-    enum_error = mixin(models.Analysis, analysis, request.json, editable_columns)
+    enum_error = mixin(analysis, request.json, editable_columns)
 
     if enum_error:
         abort(400, description=enum_error)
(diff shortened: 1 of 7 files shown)


kevinlul (approved):
LGTM!

Issue #847, opened 2021-09-30
Flask: add dataset.sequencing_id to csv payload from /api/analyses endpoint


cklamann:
The value does not need to be included in the JSON response.

PR #852, opened 2021-10-04
Flask: fix sequencing_id in _bulk and GET analyses

### LINKED ISSUE #846: Flask: add dataset.sequencing_id to _bulk endpoint
`sequencing_id` is currently not being uploaded to the database with the rest of the dataset fields. `routes.py::bulk_update` should be updated to include it.

delvinso:
closes #846 and #847 

I didn't think it was necessary to update tests, but as a sanity check for both:

- GET analyses (csv) will return `sequencing_id` : ```curl -H "Accept: text/csv" localhost:5000/api/analyses\?user=1```
- _bulk now inserts `sequencing_id` into the database and returns it in the json : ```curl -XPOST -H 'Content-Type: text/csv' --data-binary @flask/tests/samplecsv.csv localhost:5000/api/_bulk\?user=1\&groups=cheo```

### DIFF
--- a/flask/app/analyses.py
+++ b/flask/app/analyses.py
@@ -246,6 +246,7 @@ def list_analyses(page: int, limit: int) -> Response:
     results = [
         {
             **asdict(analysis),
+            "sequencing_id": [d.sequencing_id for d in analysis.datasets],
             "participant_codenames": [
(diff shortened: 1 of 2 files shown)


cklamann (approved):
LGTM!

PR #862, opened 2021-10-08
CI: rewrite backend workflow to be push-based

kevinlul:


kevinlul:
In 57b9cf2, I demonstrate that we can run code on our self-hosted runner and start the container there. Now I've added the necessary secrets to SSH to the actual host to deploy the backend.

PR #868, opened 2021-10-12
React/Flask: refactor modals and DetailSection

jennyziyi-xu:
`DetailSection` component is refactored to be more generic. An `update` function is passed from `DatasetInfoDialog` and `ParticipantInfoDialog` to `DetailSection`. The `update` function handles the changes made in the modals. 

For `Participant` modal, a query is added in Flask to handle GET request. 
- If the `user` is an admin, the `user` can access all `participants` and all related `datasets`. 
- If the `user` is not an admin, the `user` only has access to certain participants who have `datasets` that share the same group as `user`. Inside a `Participant` modal, only `datasets` that share at least one common group as `user` are listed. The `tissue samples` of the participant are shown or not shown accordingly. 

Issues that will be addressed in other tickets: 
1, Some of the `Participant` modal fields do not work properly. 
2, Updating `linked_files` is not handled in `Dataset` modal.  
3, The snackbar displays "Dataset ID undefined updated successfully". It should instead show the specific Dataset ID.

### DIFF
--- a/flask/app/participants.py
+++ b/flask/app/participants.py
@@ -273,9 +273,9 @@ def get_participant(id: int):
             user,
         )
 
-    try:
-        participant = query.one()
-    except NoResultFound:
+    participant = query.one_or_none()
+
+    if not participant:
         abort(404)
 
     return jsonify(


cklamann:
> The same error occurs here as in #863 . Perhaps one of the weekend releases of MinIO broke something?

The problem seems to have been introduced in this release: `RELEASE.2021-10-10T16-53-30Z` . Tests pass locally with this one: `RELEASE.2021-10-08T23-58-24Z`

jennyziyi-xu:
> In testing this we've also found the `test_reset_minio_credentials_allows_admins_to_delete` to be faulty as it cannot be run independently of the other tests in the file. The needed buckets don't get created, so the policy can't be applied. This test was introduced in #690. It may be because the fixture parameters are in the wrong order?

The issue is fixed by deleting `minio_admin.set_policy()` in  `test_users.py`. As pointed out in `users.py` -> `def reset_minio_credentials`, MinIO requires a user to exist to create the group. So the policy can only be set after creating a user.

kevinlul:
Please also freeze the MinIO version to `RELEASE.2021-10-08T23-58-24Z` in all Compose files so the tests pass to address the recent regression.

flask/app/participants.py
```diff
@@ -247,6 +248,65 @@ def list_participants(page: int, limit: int) -> Response:
...
+                    models.TissueSample.datasets
+                ),
+            ),
+            user,
+        )
+
+    try:
+        participant = query.one()
+    except NoResultFound:
+        abort(403)
```
kevinlul, line 284: If you look at `get_dataset` (`GET /api/datasets/:id`), you shouldn't be performing two queries here against the participant table. You should only do one. The only return code should be a 404, regardless of whether the user has permission or not to view a participant, as we don't want to reveal its existence by a different status code.

The family and institution relationships should be loaded by a joinedload, since a participant only has one of each.

flask/tests/test_participants.py
```diff
@@ -111,6 +111,47 @@ def test_list_participants_user_from_admin(test_database, client, login_as):
...
+    assert client.get(f"/api/participants/2?user=1").status_code == 200
+
+    assert client.get(f"/api/participants/1?user=2").status_code == 200
+    assert client.get(f"/api/participants/1?user=4").status_code == 403
+
+
+def test_get_participants_user(test_database, client, login_as):
+    login_as("user_c")
+
+    # Test if the user can only see datatsets that belong to the same group the user belongs to.
```
kevinlul, line 139: typo in "datatsets"

flask/app/participants.py
jennyziyi-xu: I made the corresponding changes! I am still using `query.one()` instead of `query.first_or_404()` because `first()` applies a LIMIT of 1 *within* the generated SQL, so that only one primary entity row is generated on the server side. This behaviour is not desired since we want to obtain all the datasets from the subquery.

jennyziyi-xu: The main entity is the participant, but there is a subquery performed in `filter_datasets_by_user_groups`. This [post](https://stackoverflow.com/questions/29161730/what-is-the-difference-between-one-and-first) is relevant in terms of the difference between `first()` and `one()` in this context. 

Also, the [official documentation](https://docs.sqlalchemy.org/en/14/orm/query.html#sqlalchemy.orm.Query.first) says:

" first() applies a limit of one within the generated SQL, so that only one primary entity row is generated on the server side (note this may consist of multiple result rows if join-loaded collections are present)."

This is not the case for `one()`. I am not totally sure regarding my current implementation. (I also think this scenario is different than `get_dataset()`).

kevinlul:
Re: the difference between `.one_or_none()` and `.first_or_404()`, it was previously encountered during #184

PR #871, opened 2021-10-12
React Doc & .gitattribute update

Lakshya2610:
- React.md: Added missing step in getting started guide
- .gitattribute: Force LF line ending for bash scripts to avoid errors for windows users

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,4 @@
 * text=auto
+
+# force LF line ending for these files
+*.sh text eol=lf
--- a/docs/react.md
+++ b/docs/react.md
@@ -18,15 +18,16 @@ docker-compose up -d --build
 ```
 
 1. Switch to the `react` directory.
-1. Create a `.env` with these parameters that you can change as needed.
+2. Create a `.env` with these parameters that you can change as needed.
 
    ```
    REACT_APP_NAME=Stager
    REACT_APP_MINIO_URL=http://localhost:9000/
    ```
 
-1. Install dependencies with `yarn`.
-1. Start a development server in watch mode with `yarn start`.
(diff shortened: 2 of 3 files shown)


.gitattributes
```diff
@@ -1 +1,4 @@
 * text=auto
+
+# force LF line ending for these files
+*.sh text eol=lf
```
kevinlul, line 4: I think you can put the `eol=lf` directive above for all files so we don't get caught off-guard again.

docs/react.md
```diff
@@ -18,15 +18,16 @@ docker-compose up -d --build
...
+2. Create a `.env` with these parameters that you can change as needed.
 
    ```
    REACT_APP_NAME=Stager
    REACT_APP_MINIO_URL=http://localhost:9000/
    ```
 
-1. Install dependencies with `yarn`.
-1. Start a development server in watch mode with `yarn start`.
+3. Set ```ENABLE_OIDC=``` to blank (empty).
```
kevinlul, line 28: This step applies to the steps in `docker.md` as it concerns the backend. You also do not need to actually number in order for Markdown ordered lists.

kevinlul (changes requested):
In the future, for our PR branches, we generally name them in the form `name/feature` or `username/feature`

docs/react.md
Lakshya2610: I can add it there as well. But I think this should be here as well. The current doc says that if the user does all the given steps, react app should work which isn't true. What do think?

kevinlul: The docs here are scoped specifically to the frontend. It links to `docker.md` before the steps for how to set that up. I think the problem is that as we discovered at the end of last week, the docs each only describe a component, and there's no "Getting started" section in the main README to elucidate how to read them.

PR #881, opened 2021-10-15
Flask: Added request info to logs for 5xx error codes

### LINKED ISSUE #880: Flask: Add request information to 5xx error traces
Our 500-level stack traces should also print selected information about the request. For now this should include the request, query string parameters and, for Post/Put/Patch, the request body. Pretty format would be great too :slightly_smiling_face:

Lakshya2610:
* For GET requests, the full url+method & query parameters are printed
* For other requests, the body is also printed (including the form data)
* Query/body is pretty printed with proper indentation

Sample log:
```
<Request 'http://localhost:5000/api/_bulk?groups=c4r' [POST]>
Query Params:
{
    "groups": "c4r"
}
Body:
[
    {
        "affected": null,
        "condition": "Control",
        "dataset_type": "CES",
        "family_codename": "test",
        "participant_codename": "test",
        "participant_type": null,
        "sequencing_date": "2021-10-13",
        "sex": "Unknown",
        "solved": false,
        "tissue_sample_type": "Blood"
    }
]
```
	
Fixes #880

### DIFF
--- a/flask/app/error_handler.py
+++ b/flask/app/error_handler.py
@@ -21,6 +21,11 @@ def get_request_info() -> str:
     return info
 
 
+def dump_error_context() -> None:
+    app.logger.error(get_request_info())
+    app.logger.error(traceback.format_exc())
+
+
 @error_blueprint.app_errorhandler(Exception)
 def handle_error(error: Exception):
 
@@ -35,13 +40,14 @@ def handle_error(error: Exception):
     code = 500  # defaults to 500 for non HTTP exceptions
     msg = error
 
-    # this is useful for non-http exceptions as well
-    app.logger.error(get_request_info())
-
     if isinstance(error, HTTPException):
         code = error.code
         msg = error.description
         # 500 codes are handled as they all exist as a subclass of HTTPException
         if code in werkzeug_500_codes:
-            app.logger.error(traceback.format_exc())
+            dump_error_context()
+    else:
+        # this is useful for non-http exceptions as well
+        dump_error_context()
+
     return jsonify(error=str(msg)), code


flask/app/error_handler.py
```diff
@@ -1,11 +1,21 @@
...
 import traceback
 from sqlalchemy import exc
 from .extensions import db
 
 error_blueprint = Blueprint("error_handler", __name__)
 
+# dump relavant request info into a string
+def get_request_info() -> str:
+    info: str = "\n"
+    info += f"{str(request)}\n"
```
kevinlul, line 12: I'm not sure if this is helpful because this will be reflected in a later server log line, e.g. `"GET /api HTTP/1.1" 503`

flask/app/error_handler.py
```diff
@@ -1,11 +1,21 @@
...
+    info += f"Query Params:\n{json.dumps(request.args.to_dict(), indent=4)}\n"
+    if request.method != "GET":
+        body = request.get_data(parse_form_data=True)
+        info += f"Body:\n{json.dumps(json.loads(body), indent=4)}\n"
```
kevinlul, line 16: We are reading the entire HTTP payload and treating it as JSON conditional only on the request method not being GET! Payloads could be attached to GET requests too (unconventional), but more importantly, if the payload is not meant to be JSON or is badly formed, this will error, and we cannot have an error handler raise an exception itself.

Lakshya2610:
> I'm not so sure this will capture everything. If you edit `/api` (routes.py::version) to raise an exception, unless it's specifically a werkzeug 500 type, this will not capture it at all, but this would be how exceptions bubble out from SQLAlchemy.

@cklamann Suggested to dump request info **only** for 5xx type errors

cklamann:
> If you edit /api (routes.py::version) to raise an exception, unless it's specifically a werkzeug 500 type, this will not capture it at all, but this would be how exceptions bubble out from SQLAlchemy.

This is a good point. Would it be more accurate to say that we'd like to print traces of all exceptions except 4xx?  Another option might be to break our [generic handler](https://flask.palletsprojects.com/en/2.0.x/errorhandling/#generic-exception-handlers) into a few more specific handlers, however I'm not sure we'd gain much from doing that if we want to handle 5xx and non-http exceptions the same way.

Lakshya2610:
Oh, I see what you mean. That makes sense. Yeah, I guess I could just move it out of the `if isintance(..` to catch all exceptions, not just `HTTPException`

flask/app/error_handler.py
```diff
@@ -1,11 +1,25 @@
...
 from .extensions import db
 
 error_blueprint = Blueprint("error_handler", __name__)
 
+# dump relavant request info into a string
+def get_request_info() -> str:
+    info: str = "\n"
+    info += f"{str(request)}\n"
+    info += f"Query Params:\n{json.dumps(request.args.to_dict(), indent=4)}\n"
+    if "json" in request.headers.get("Content-Type", ""):
```
kevinlul, line 14:
```suggestion
    if request.mimetype == "application/json":
```

PR #890, opened 2021-10-18
React: Added Linked Files Dropdown

hannahle:
- Added the `FileLinkingComponent` to `DatasetInfoDialog`. The functionality here should be the same as the `FileLinkingComponent` when the user edits the row. 
- The check button for saving changes seemed a bit out of place. I replaced with with a button so the design looks more uniform.

### DIFF
--- a/react/src/Datasets/components/DatasetInfoDialog.tsx
+++ b/react/src/Datasets/components/DatasetInfoDialog.tsx
@@ -90,9 +90,12 @@ export default function DatasetInfoDialog({ dataset_id, onClose, open }: DialogP
             },
             {
                 onSuccess: receiveDataset => {
-                    enqueueSnackbar(`Dataset ID ${newData.dataset_id} updated successfully`, {
-                        variant: "success",
-                    });
+                    enqueueSnackbar(
+                        `Dataset ID ${receiveDataset.dataset_id} updated successfully`,
+                        {
+                            variant: "success",
+                        }
+                    );
                 },
                 onError: response => {
                     console.error(
--- a/react/src/Datasets/components/DatasetTable.tsx
+++ b/react/src/Datasets/components/DatasetTable.tsx
@@ -60,7 +60,6 @@ const customFileFilterAndSearch = (filter: string, rowData: Dataset) => {
 const EditFilesComponent = (props: EditComponentProps<Dataset>) => {
     const filesQuery = useUnlinkedFilesQuery();
     const files = filesQuery.data || [];
-    console.log(props.rowData.linked_files);
     return (
         <FileLinkingComponent
             values={props.rowData.linked_files}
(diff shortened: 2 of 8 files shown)


react/src/Datasets/components/DatasetInfoDialog.tsx
```diff
@@ -68,15 +67,6 @@ export default function DatasetInfoDialog({ dataset_id, onClose, open }: DialogP
         const newData = fields
             .map(field => {
                 if (field.fieldName && !field.disableEdit) {
-                    if (
```
jennyziyi-xu, line 71: This section of code is responsible for reformatting the value in `Library Prep Date` as a string that matches a certain regex such that the backend can interpret correctly the request as a Date object. Without this reformatting step, the backend would just interpret the request as a string and is unable to convert it to a Date object. 

Right now, since this section is deleted and not added anywhere else, we would observe a 500 INTERNAL SERVER error when changes to `Library Prep Date` are made.

jennyziyi-xu (changes requested):

react/src/Datasets/components/DatasetInfoDialog.tsx
```diff
@@ -68,15 +67,6 @@ export default function DatasetInfoDialog({ dataset_id, onClose, open }: DialogP
...
-                        field.fieldName === "library_prep_date" &&
-                        field.value != null &&
-                        typeof field.value === "string" &&
-                        /[A-Z]/.test(field.value[0])
-                    ) {
-                        const datePart = field.value.substring(field.value.indexOf(",") + 2);
-                        field.value = dayjs(datePart, "MMMM D, YYYY h:mm A").format("YYYY-MM-D");
-                    }
```
jennyziyi-xu, line 79: This section of code is responsible for reformatting the value in Library Prep Date as a string that matches a certain regex such that the backend can interpret correctly the request as a Date object. Without this reformatting step, the backend would just interpret the request as a string and is unable to convert it to a Date object.

Right now, since this section is deleted and not added anywhere else, we would observe a 500 INTERNAL SERVER error when changes to Library Prep Date are made.

jennyziyi-xu: In the PR: "Refactored updateDataset such that the date formatting is moved inside the formatFieldValue function. This ensures that the date has the appropriate format when passing back to the backend."

According to what I found earlier, the formatting needs to satisfy a certain regex so Python can interpret it correctly. There may be different ways to achieve this actually. What I came up earlier is only one way. I found this which can help: https://www.tutorialsteacher.com/articles/convert-string-to-datetime-in-python

hannahle: Oh interesting! This shouldn't be the case because I added a code block to `formatFieldValue` that should handle the reformatting when the data is passed to the backend.

jennyziyi-xu: Do you also observe a 500 INTERNAL SERVER error when the `library_prep_date` is not null?

jennyziyi-xu: Hannah and I found out that 500 INTERNAL SERVER happens when the second time we save an edit, the `library_prep_date` is left unchanged. This is exactly what I addressed in an earlier ticket: https://github.com/ccmbioinfo/stager/issues/814

Issue #908, opened 2021-10-26
React: Configure login screen and dynamic endpoint selection to support multiple backends


cklamann:
- add dropdown to login page to select lab if multiple labs are supported (CHEO front-end, for example, should not have a dropdown)
- selected API endpoint should be passed to app context and used to configure ReactQuery (this includes login call, since user databases will vary by backend)
- lab-to-endpoint configurations should be environment-specific, so that local and build environments can provide their own mappings.

from #875

PR #925, opened 2021-11-02
React: Dropdown for lab selection

Lakshya2610:
- Added a dropdown component to allow for lab selection
- On lab selection, the custom endpoint is saved in `APIInfo`
- Default value for endpoint in `APIInfo` is null which implies the app should use the default endpoint

Note: Still need to do some refactoring to make use of this custom endpoint value

#908

### DIFF
--- a/react/src/App.tsx
+++ b/react/src/App.tsx
@@ -86,9 +86,6 @@ function BaseApp(props: { darkMode: boolean; toggleDarkMode: () => void }) {
             const apiInfoResult = await fetch("/api");
             if (apiInfoResult.ok) {
                 let apiInfo = await apiInfoResult.json();
-                if (!("endpoint" in apiInfo)) {
-                    apiInfo = { ...apiInfo, endpoint: null }; // null endpoint value -> use the default endpoint
-                }
                 setApiInfo(apiInfo as APIInfo);
             }
         })();
--- a/react/src/App.tsx
+++ b/react/src/App.tsx
@@ -85,7 +85,10 @@ function BaseApp(props: { darkMode: boolean; toggleDarkMode: () => void }) {
 
             const apiInfoResult = await fetch("/api");
             if (apiInfoResult.ok) {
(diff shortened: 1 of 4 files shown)


cklamann (changes requested):
I can see the lab selection control on the login screen, though I don't see any options and the login text controls are hidden. I think that if we want to merge this in as a WIP, it would be best to hide the dropdown, probably just conditioned on the number of endpoint options available (which right now would be 0).

Lakshya2610:
I added the default "Select a lab" option at the very end which broke the conditional rendering. Fixed it now

react/src/App.tsx
```diff
@@ -85,11 +85,20 @@ function BaseApp(props: { darkMode: boolean; toggleDarkMode: () => void }) {
 
             const apiInfoResult = await fetch("/api");
             if (apiInfoResult.ok) {
-                const apiInfo = await apiInfoResult.json();
+                let apiInfo = await apiInfoResult.json();
+                if (!("endpoint" in apiInfo)) {
+                    apiInfo = { ...apiInfo, endpoint: null }; // null endpoint value -> use the default endpoint
+                }
```
kevinlul, line 91: Are you supposing that the correct endpoint (domain or prefix) for this tenant will be in the response from the `/api` endpoint on the current domain?

Lakshya2610: This was an assumption initially that @cklamann cleared up. That statement doesn't need to be wrapped in an if condition.

kevinlul: No, the problem is that you can't make a request to the backend to figure out what backend to make requests to!

Lakshya2610: Yeah, this code was initially done with the incorrect assumption. I'm gonna remove this for now as it'll be fixed in the second PR for this feature. This PR is for purely UI stuff

kevinlul:
The result after adding items to `getLabs()`
image

PR #927, opened 2021-11-02
Flask: separate MinIO upload bucket from results bucket

### LINKED ISSUE #874: Flask: separate MinIO upload bucket from results bucket
Each permission group can currently access one bucket of the same name. Extend this so now two can be accessed: the original `GROUP`, now designated for uploads only, and `results-GROUP` for holding pipeline outputs, manual or automatic. The behaviour of linked files should not change, but it will no longer be cluttered by outputs that can never actually be inputs. Furthermore, `results-` shall be an illegal prefix for a group code.

kevinlul:
Update policy to include 'results-CODENAME'
Deny group codes starting with 'results-'
Add Click command 'migrate-minio-policies' to retroactively update existing groups
Objects will have to be moved to the correct bucket after the migration
Fix order of input validation for POST and PATCH endpoints

Closes #874

Missing
- [x] update tests
- [x] retroactively create buckets?
- [x] create the result bucket at the time of group creation?

### DIFF
--- a/.github/workflows/flask.yml
+++ b/.github/workflows/flask.yml
@@ -136,7 +136,7 @@ jobs:
           MINIO_ENDPOINT: ${{ secrets.MINIO_ENDPOINT }}
           MINIO_ACCESS_KEY: ${{ secrets.MINIO_ACCESS_KEY }}
           MINIO_SECRET_KEY: ${{ secrets.MINIO_SECRET_KEY }}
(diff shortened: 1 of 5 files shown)


delvinso (approved):
LGTM, works on my end!

kevinlul:
The migration has been performed in prod.

PR #929, opened 2021-11-05
Flask: Fix updated timestamp in the modals.

jennyziyi-xu:
When we save changes in the modals and if there are no changes in the field values, the `updated` timestamp would still change. This is caused by either a nonnull `library_prep_date` field or a nonnull `month_of_birth` field. When the backend receives the JSON request, the values for these fields are of type String. However, the values of these fields are stored in the database as `DateTime.Date()` object. Thus, when we commit the session, SQLAlchemy will think that there's a change in data values and update the timestamp. 

The solution presented here is to convert the values from `String` type to `DateTime.Date()` type in the backend.


kevinlul (approved):
LGTM!

PR #930, opened 2021-11-05
React: Modal loading and circular progress icon display. 

jennyziyi-xu:
For the modals (info dialog) in `Participant` `Analysis` `Dataset` tables, a backdrop and a circular progress icon are implemented. The new feature is effective when a user saves changes in the modal and when a user opens up the modal.


cklamann (approved):
LGTM!

PR #931, opened 2021-11-05
Docker: disable MinIO update check at startup, forward Console port, freeze version

kevinlul:
The combination of MINIO_UPDATE=off AND --quiet is required to prevent calling home in this instance
This means we lose all startup messages and logs

Forward the Console port for MinIO releases since 20210617
Freeze version due to regression in 20211105 (no release notes) discovered by Jenny


cklamann (approved):
LGTM!

PR #932, opened 2021-11-05
React: use no-referrer globally and for all external links

kevinlul:
Please check if I missed any. To observe the behaviour change, open the developer tools, disable cache, and click on outbound links. There should be no `Referer:` header. This should also be true for the two font requests on a hard reload.

Documentation:
https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types/noreferrer
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer
https://developer.mozilla.org/en-US/docs/Web/Security/Referer_header:_privacy_and_security_concerns
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/link#attr-referrerpolicy

### DIFF
--- a/react/public/index.html
+++ b/react/public/index.html
@@ -2,6 +2,7 @@
 <html lang="en">
     <head>
         <meta charset="utf-8" />
+        <meta name="referrer" content="no-referrer" />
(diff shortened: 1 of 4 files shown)


cklamann (approved):
LGTM!

PR #938, opened 2021-11-10
React: fix dataset types going out of sync between the Dataset and Participant modals

jennyziyi-xu:
Fixed bug: if a dataset's type is updated in the Dataset table or Dataset modal, the corresponding change is not reflected in the Participant table or Participant modal if we don't refresh the Participant page.

### DIFF
--- a/flask/app/participants.py
+++ b/flask/app/participants.py
@@ -206,6 +206,11 @@ def list_participants(page: int, limit: int) -> Response:
     results = [
         {
             **asdict(participant),
+            "dataset_types": [
+                dataset.dataset_type
+                for tissue_sample in participant.tissue_samples
+                for dataset in tissue_sample.datasets
+            ],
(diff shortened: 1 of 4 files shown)


cklamann (commented):
The update/refresh behavior seems to be working as expected locally for me! Question though: is it absolutely necessary for the participant payloads to have a top-level `dataset_types` property when the front end could potentially derive the list from `partipant.tissue_samples.dataset`?

jennyziyi-xu:
I am not sure whether `dataset_types` would be better derived in the frontend or backend. Since `dataset_types` is not an attribute for class `Participant` in `models.py`, it makes sense that it should be derived from the frontend. 

As a side note, I am leaving duplicate dataset types in the modal, but I can remove the duplicates if that's needed.

kevinlul (approved):
LGTM!

PR #940, opened 2021-11-12
React: Dynamic Endpoint Selection

### LINKED ISSUE #908: React: Configure login screen and dynamic endpoint selection to support multiple backends

Lakshya2610:
- Added a new React context: `FetchContext` to store & access active endpoint
- Modified all fetch calls & fetch wrappers to use the active endpoint
- Fetch wrappers in `utils` access the active endpoint from `localStorage`

2/2 for #908 

Fixes #908


react/public/dummy_labs.json
```diff
@@ -0,0 +1,10 @@
+[
+  {
+    "name": "Lab 0",
+    "endpoint": "http://lab0.stager.ca"
+  },
+  {
+    "name": "Lab 1",
+    "endpoint": "http://lab1.stager.ca"
+  }
+]
```
kevinlul, line 10: The parent domain here is real. Please don't use it. Use our real one, localhost, or example domains that are guaranteed to not be in use.

cklamann: What do you think in terms of a plan for rolling out this code? If we merge it in while there is still just one STAGER instance (CHEO), will `labs.json` exist? Will `dummy_labs.json` be committed to source? And then once we have the CCM deployment, is `labs.json` the file that will have the ccm-specific endpoint configuration?

react/src/AddDatasets/components/UploadDialog.tsx
```diff
@@ -52,7 +54,7 @@ export default function UploadDialog({ open, onClose, groups }: UploadDialogProp
         if (file !== null) {
             // Upload
             const groupsParam = new URLSearchParams({ groups: groups.join(",") });
-            const response = await fetch("/api/_bulk?" + groupsParam.toString(), {
+            const response = await fetch(`${fetchContext}/api/_bulk?` + groupsParam.toString(), {
```
cklamann, line 57: I like storing the endpoint in a Context. What do you think about going a step farther and writing a hook that wraps `fetch` itself, either something like `useFetch` (user passes in verb, base endpoint already bound) or `usePost`, `useGet`, etc.? It would take some refactoring of whatever fetch calls aren't currently part of the React ecosystem, but it would save the trouble (now and down the line) of maintaining endpoint configurations for each and every `fetch`.

Lakshya2610: Yeah, that good idea. I'll start looking into that then, & we can hold this PR off for now

react/public/dummy_labs.json
Lakshya2610: I put the `dummy_labs.json` for it to be committed to the source so that we have a reference json. As for deployment for single endpoints, we could either not have the json or have it be an empty array, the app behavior would be identical. The app treats missing json as `[]`.

react/src/AddDatasets/components/UploadDialog.tsx
Lakshya2610: So, I ended up getting rid of the fetchContext completely. The useFetch accesses the active endpoint from `localStorage`.

react/src/App.tsx
```diff
@@ -72,24 +73,46 @@ function BaseApp(props: { darkMode: boolean; toggleDarkMode: () => void }) {
...
+            const apiInfoResult = await fetch(`/api`);
             if (apiInfoResult.ok) {
-                const apiInfo = await apiInfoResult.json();
+                let apiInfo = await apiInfoResult.json();
                 setApiInfo(apiInfo as APIInfo);
             }
         })();
     }, []);
+
+    const setEndpoint = async (newEndpoint: string) => {
```
cklamann, line 108: I think this might need to be renamed to something like `setApiInfo`, since the endpoint is actually being set in LabDropdownSelect.tsx. It also might be worth using this function in place of the code lines 100-104 if `customFetch` can be called at this point (since the endpoint has already been set in `LabDropdownSelect`).

react/src/hooks/utils.tsx
```diff
@@ -14,6 +14,19 @@ import { Query, SetDataOptions } from "react-query/types/core/query";
...
 
+function getActiveEndpoint(): string {
+    const activeEndpoint = localStorage.getItem("endpoint");
+    return activeEndpoint === null ? "" : activeEndpoint;
+}
+
+/**
+ * Wrapper for fetch - injects currently selected endpoint url to the outgoing request
+ */
+export function customFetch(url: string, ...args: any[]): Promise<Response> {
```
cklamann, line 25: Is there a more expressive name for this function? `apiFetch`? Maybe the real issue is our proliferation of `fetch`s-- `changeFetch`, `basicFetch`, `customFetch`. Is it possible to merge `basicFetch` and `customFetch`?

cklamann (changes requested):
By entering my (identical) local endpoints into `dummy_labs.json` I was able to test the basic functionality and everything seems to be working as intended with the exception of the call to the `_bulk` endpoint on the Add Datasets page, which seems now to be making a `GET` instead of `POST` request.

react/public/dummy_labs.json
```diff
@@ -0,0 +1,10 @@
+[
+  {
+    "name": "Lab 0",
+    "endpoint": "http://lab0.endpoint.com"
+  },
+  {
+    "name": "Lab 1",
+    "endpoint": "http://lab1.endpoint.com"
+  }
+]
```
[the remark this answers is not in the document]
Lakshya2610: I left that so the person that creates the actual config for deployment has a reference for the formatting. The name of the file for the deployment should always be `labs.json`

react/src/App.tsx
Lakshya2610: Fixed it

react/src/hooks/utils.tsx
Lakshya2610: Renamed `customFetch` to `apiFetch`

cklamann (approved):

PR #943, opened 2021-11-18
React/Flask: Add condition to dataset_type editability, fix Analyses table part I. 

jennyziyi-xu:
1. In `Dataset` modal and table, disable a dataset's ` dataset_type` editing if there's an analysis associated with it. Now the user can't change a `dataset_type` once there's an analysis associated with it.

2. Display all analyses, including the ones that don't have an assignee.

3. Fix the table nonrefreshing issue when a re-analysis is requested through the dialogue.

4. Fix the table nonrefreshing issue when an analysis is cancelled.

Other issues to be addressed in another ticket: 

- When the user selects an analysis and performs any of the following actions: run, complete, error, set assignee, the status is not reflected immediately.  For some of these actions, the snackbar would not even show up. My preliminary investigation for the action `Run Analysis` reveals that this is due to the asynchronous nature of `setActiveRows()` such that `changeAnalysisState` is not supplied with the new `activeRows`.

### DIFF
--- a/flask/app/analyses.py
+++ b/flask/app/analyses.py
@@ -519,32 +519,6 @@ def create_reanalysis(id: int):
         abort(404, description="Some datasets were not found")
 
     # TODO: Do we need to check for compatibility if the previous analysis already exists?
-    app.logger.info(
-        f"Checking if datasets are still compatible with pipeline {analysis.pipeline_id}..."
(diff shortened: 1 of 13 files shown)


cklamann (approved):
LGTM!

PR #945, opened 2021-11-18
Docker: fix up configurations

kevinlul:
Fix nonfunctional healthcheck
--quiet not required to disable MinIO update check
Implicitly refer to envvars of the same name
Bump to 3.9
Set defaults for HOST_PORT vars
Remove unhelpful UIDGID configuration
Update docs


cklamann (approved):
LGTM!

PR #946, opened 2021-11-19
React: Fix actions in Analysis Table.

jennyziyi-xu:
When the user selects an analysis and performs any of the following actions: run, complete, error, set assignee, the status is not reflected immediately. For some of these actions, the snackbar would not show up.

This is fixed by passing `activeRows` or newly selected rows to the mutation functions directly. Otherwise, the mutation functions are not supplied with fresh data for `activeRows`. 

Note that some other issues in Analysis Table were addressed in #943


react/src/Analyses/index.tsx
```diff
@@ -186,8 +186,9 @@ export default function Analyses() {
         "analysisTableSortOrder"
     );
 
-    function changeAnalysisState(newState: PipelineStatus) {
-        return _changeStateForSelectedRows(activeRows, analysisUpdateMutation, newState);
+    function changeAnalysisState(newState: PipelineStatus, activeRowsParam?: Analysis[]) {
+        const rows = activeRowsParam || activeRows;
```
kevinlul, line 190: I think you can use a default argument here.
```suggestion
    function changeAnalysisState(newState: PipelineStatus, rows = activeRows) {
```

kevinlul (changes requested):

cklamann (approved):
LGTM!

Issue #935, opened 2021-11-09
Flask: email alerts for analysis requests


cklamann:
Research and configure a method for sending automated emails from the Flask app to alert users that an analysis has been requested. Options include:
- SendGrid or similar
- Setting up our own SMTP solution

kevinlul:
stager.genomics4rd.ca verified SendGrid

PR #947, opened 2021-11-19
Flask: implement email notifications

hannahle:
- We used SendGrid Email API to send emails to appropriate users when an analysis is requested.
- The `send_email` function will send from whichever user email we specify in the `from_email` field. The email domains in `from_email` would need to be verified for some email providers (like SickKids) to accept it and show it in the end user's inbox. So far, we've verified the `ccmdev.ca` domain, and the test `from_email` is `test@ccmdev.ca`.
- In the email, there would also be an Unsubscribe option, where users can choose to not receive future notification emails. 
- The code also checks for whether we have exceeded the email limit in the day, and if so, queue the email the next day. Otherwise, it sends the email 1 minute after an analysis has been successfully requested.

**To test the code, you need to:**

1. Create a SendGrid account and an API key and add it to `SENDGRID_API_KEY` in `.env`.
2. Follow these steps in [Single Sender Verification](https://docs.sendgrid.com/ui/sending-email/sender-verification) to verify yourself as a sender. If you are not verified as a sender, you will not be able to send any email.Hence, the recipient (which is also you in this case, because you're testing) won't receive any email. 
_Note:_ Single Verification is only used for fast, local testing and not recommended in production. As mentioned above, we are going to verify the email sender domains, similar to `ccmdev.ca`, in production.
3. Add the email you have verified in step 2 to the `from_email` argument in `email.send_email`
4. Add your personal (non-SickKids) email to the `to` field. The reason why you can't use your SickKids email here is because SickKids email server doesn't trust emails sent via sendgrid as a proxy. This problem is solved once we verify the domain. 
5. Import [this](https://drive.google.com/file/d/1qgtVTJbpqtqpjSJoUOlybzjBemeyXHzp/view?usp=sharing) custom email HTML template to the **[Dynamic Templates](https://mc.sendgrid.com/dynamic-templates)** section in your SendGrid dashboard.
6. Copy-paste the template ID and add it to `.env`.
7. Create an Unsubscribe group in your dashboard [here](https://app.sendgrid.com/suppressions/advanced_suppression_manager). Add your unsubscribe group ID to `.env`.
5. Try requesting an analysis and you should see an email to your personal inbox!

**Note:**

This code can be merged into master with email notification disabled by simplying setting the `SENDGRID_API_KEY` to empty until we've properly confirmed all collaborators that we will be sending them email.


docker-compose.yaml
```diff
@@ -34,9 +34,17 @@ services:
...
+      MINIO_SECRET_KEY: "${MINIO_SECRET_KEY}"
+      MINIO_REGION_NAME: "${MINIO_REGION_NAME}"
+      ENABLE_OIDC: "${ENABLE_OIDC:-}"
+      OIDC_CLIENT_ID: "${OIDC_CLIENT_ID}"
+      OIDC_CLIENT_SECRET: "${OIDC_CLIENT_SECRET}"
+      OIDC_WELL_KNOWN: "${OIDC_WELL_KNOWN}"
+      OIDC_PROVIDER: "${OIDC_PROVIDER}"
+      SENDGRID_API_KEY: "${SENDGRID_API_KEY}"
+      SENDGRID_UNSUBSCRIBE_GROUP_ID: "${SENDGRID_UNSUBSCRIBE_GROUP_ID}"
+      SENDGRID_EMAIL_TEMPLATE_ID: "${SENDGRID_EMAIL_TEMPLATE_ID}"
```
kevinlul, line 47: Looks like this merge went the wrong way. The result should be
```suggestion
      MINIO_ACCESS_KEY:
      MINIO_SECRET_KEY:
      MINIO_REGION_NAME: 
      SENDGRID_API_KEY:
      SENDGRID_UNSUBSCRIBE_GROUP_ID:
      SENDGRID_EMAIL_TEMPLATE_ID:
```

docker-compose.test.yaml
```diff
@@ -27,6 +27,9 @@ services:
       TEST_MINIO_ACCESS_KEY: "${TEST_MINIO_ACCESS_KEY}"
       TEST_MINIO_SECRET_KEY: "${TEST_MINIO_SECRET_KEY}"
       MINIO_REGION_NAME:
+      SENDGRID_API_KEY: "${SENDGRID_API_KEY}"
+      SENDGRID_UNSUBSCRIBE_GROUP_ID: "${SENDGRID_UNSUBSCRIBE_GROUP_ID}"
+      SENDGRID_EMAIL_TEMPLATE_ID: "${SENDGRID_EMAIL_TEMPLATE_ID}"
```
kevinlul, line 32:
```suggestion
```
Shouldn't be sending emails when running tests

react/src/hooks/utils.tsx
```diff
@@ -10,10 +10,12 @@ import {
     UseQueryOptions,
     UseQueryResult,
 } from "react-query";
-import { Query, SetDataOptions } from "react-query";
+import { Query } from "react-query";
 import { stringToBoolean } from "../functions";
 import { QueryWithSearchOptions as MTQueryWithSearchOptions } from "../typings";
 
+type SetDataOptions = NonNullable<Parameters<QueryClient["setQueryData"]>[2]>;
+
```
kevinlul, line 18: What's going on here in the frontend? I think you might have an old version of react-query installed. Try running `yarn` again to update.

flask/app/email.py
```diff
@@ -0,0 +1,91 @@
...
+                send_at = send_at + timedelta(days=1, minutes=1)
+
+    return send_at.timestamp()
+
+
+# Utils
+
+
+def stringify_date(date):
+    return datetime.strftime(date, "%Y-%m-%d")
```
kevinlul, line 91: This is only used once. Every other time you're directly calling `strftime`

flask/app/email.py
```diff
@@ -0,0 +1,91 @@
...
+    To,
+    From,
+    Subject,
+)
+from flask import current_app as app
+from pytz import timezone
+import math
+
+
+sg = SendGridAPIClient(os.environ.get("SENDGRID_API_KEY"))
```
kevinlul, line 20: We use `getenv` elsewhere in the codebase, though they are totally aliases.

flask/app/email.py
```diff
@@ -0,0 +1,91 @@
...
+
+tz = timezone("EST")
+
+
+def send_email(from_email, to_emails, subject, dynamic_template_object):
```
kevinlul, line 25: Add type hints

flask/app/analyses.py
```diff
@@ -454,6 +454,45 @@ def create_analysis():
     db.session.add(analysis)
     transaction_or_abort(db.session.commit)
 
+    email.send_email(
```
kevinlul, line 457: It doesn't look like this is conditioned on having a SendGrid API key?

kevinlul (changes requested):
So I'm not quite seeing how email batching is implemented here, just a delay. I also don't see how the metadata dictionary becomes the email body. Are these both done on SendGrid?

react/src/hooks/utils.tsx
hannahle: Cool! Will try installing yarn again.

flask/app/analyses.py
hannahle: When the `SENDGRID_API_KEY` in `.env` is set to empty, the `send_email` would immediately throw an 403 Unauthorized error and the email would fail to be sent.

flask/app/email.py
hannahle: > So I'm not quite seeing how email batching is implemented here, just a delay. I also don't see how the metadata dictionary becomes the email body. Are these both done on SendGrid?

Ah, the email batching (i.e. content from 5 analysis requests merged into 1 email) is not implemented yet. One email is sent per analysis request for now. I'll look into this!

Regarding the email body, this is done via SendGrid. What's happening is you have an HTML file that represents the email body and you set that as a template inside SendGrid. To tell your code to use that template, you can see that I have added a template ID.

hannahle:
> So I'm not quite seeing how email batching is implemented here, just a delay. I also don't see how the metadata dictionary becomes the email body. Are these both done on SendGrid?

Ah, the email batching (i.e. content from 5 analysis requests merged into 1 email) is not implemented yet. One email is sent per analysis request for now. I'll look into this!

Regarding the email body, this is done via SendGrid. What's happening is you have an HTML file that represents the email body and you set that as a template inside SendGrid. To tell your code to use that template, you can see that I have added a template ID.

flask/app/analyses.py
kevinlul: It's important to have the conditional before that, because you're sending out a meaningless request to another server and that adds to the response time of this endpoint.

hannahle: Cool! I'll add it in.

kevinlul:
#935

flask/app/analyses.py
```diff
@@ -454,6 +454,44 @@ def create_analysis():
...
+                "updated_by": dataset.tissue_sample.updated_by.username,
+                "created_by": dataset.tissue_sample.created_by.username,
+                "participant_notes": dataset.tissue_sample.participant.notes,
+            }
+            for dataset in found_datasets
+        ],
+    }
+
+    # Add email to cache
+    current_cache = cache.get("analyses_emails")
```
kevinlul, line 489:
```suggestion
    current_cache = cache.get("analyses_emails") or []
```
With this you don't need to have the if/else after with nearly identical bodies

flask/app/config.py
```diff
@@ -37,3 +37,5 @@ class Config(object):
     )
 
     MSTEAMS_WEBHOOK_URL = os.getenv("MSTEAMS_WEBHOOK_URL")
+
+    CACHE_TYPE = "SimpleCache"
```
kevinlul, line 41: Is this actually used?

flask/app/__init__.py
```diff
@@ -34,10 +40,27 @@ def create_app(config):
     register_extensions(app)
     manage.register_commands(app)
     register_blueprints(app)
+    register_schedulers(app)
 
     return app
 
 
+def register_schedulers(app):
```
kevinlul, line 48: We shouldn't be registering this altogether if we don't have an API key. We also definitely need an out-of-process cache as there are multiple Gunicorn workers, and we can't send out a batch of emails on exit because that would happen for every update we deploy and potentially spuriously since the server process could be restarted for whatever reason.

kevinlul (changes requested):
It looks like there are still changes to the `react/` subdirectory. You can revert them with `git checkout origin/master react`

I'm not seeing where the cache is emptied after the email is sent.

flask/app/email.py
```diff
@@ -0,0 +1,92 @@
...
+
+sg = SendGridAPIClient(os.getenv("SENDGRID_API_KEY"))
+
+tz = timezone("EST")
+
+
+def send_email(
+    from_email: str, to_emails: str, dynamic_template_object: Dict[str, Any]
+) -> None:
+    """Sends a transanctional email
```
kevinlul, line 25: Typo, suggest "sends an email based on a template stored in SendGrid"

flask/app/email.py
```diff
@@ -0,0 +1,92 @@
...
+
+    message = Mail()
+
+    message.to = To(to_emails)
+    message.from_email = From(from_email, "Stager Team")
+    message.send_at = SendAt(math.ceil(scheduled_time))
+    message.dynamic_template_data = dynamic_template_object
+    message.template_id = os.getenv("SENDGRID_EMAIL_TEMPLATE_ID")
+
+    app.logger.info(os.getenv("SENDGRID_API_KEY"))
```
kevinlul, line 46: Don't dump this to logs

flask/app/extensions.py
```diff
@@ -1,3 +1,4 @@
+from flask_caching import Cache
```
kevinlul, line 1: We're no longer using this, so make sure to remove it from dependencies!

react/src/functions.tsx
```diff
@@ -228,7 +228,7 @@ export function getDatasetFields(dataset: Dataset): Field[] {
         },
         {
             title: "Permission Groups",
-            value: dataset.group_code.join(", "),
+            value: dataset?.group_code?.join(", "),
```
kevinlul, line 231: Please make sure there are no changes to the `react/` subdirectory (unless needed for features of this PR). You can revert them with `git checkout origin/master react`

flask/app/email.py
```diff
@@ -0,0 +1,92 @@
...
+from sendgrid import SendGridAPIClient
+from sendgrid.helpers.mail import From, Mail, ReplyTo, SendAt, To
+
+from flask import current_app as app
+
+from typing import Any, Dict
+
+sg = SendGridAPIClient(os.getenv("SENDGRID_API_KEY"))
+
+tz = timezone("EST")
```
kevinlul, line 19: I'm going to have a look at whether setting the timezone in the container causes a problem with our APIs that return timestamps.

kevinlul (changes requested):

flask/app/email.py
kevinlul: It will because our datetime objects are naive, so it's out of scope for this PR. I think this should be set to `America/Toronto` or it might not account for daylight saving?

kevinlul (approved):
Thanks for working on this big feature. Looks good to me! I'll wait until Monday to merge though since Conor isn't here.

PR #948, opened 2021-11-22
Flask: Check for duplicate datasets

jennyziyi-xu:
1. Abort the transaction when either 1) one of the newly uploaded datasets is a duplicate of an existing dataset or 2) there are duplicates among the newly uploaded datasets. The combination of unique fields is: `participant_id`-`family_id`-`tissue_sample_type`-`dataset_type`-`sequencing_date`.
2. Added unit tests. Changed a few existing unit tests to avoid errors raised by duplicate datasets. 
3. Changed line 472 in `routes.py` to `db.session.flush`

Observation: `sequencing_date` of a dataset is not shown in the dataset table nor in the dataset modal. Maybe the reason is that this is a piece of information that the user should not have access to?

### DIFF
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -546,7 +546,8 @@ def bulk_update():
             app.logger.error("\tThe dataset is a duplicate of an existing dataset.")
             db.session.rollback()
             abort(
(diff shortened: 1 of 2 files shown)


cklamann (approved):
LGTM!

PR #952, opened 2021-11-24
Flask: Remove related objects when a dataset is deleted. 

### LINKED ISSUE #941: Flask: remove linked_files from dataset before deleting
Currently, the dataset `delete` endpoint does not remove the dataset's `linked_files` before deleting it (the intermediate record is removed, however). If the dataset being deleted is the only one associated with a file, the file should be removed from the database so it's available later. This could possibly be accomplished by calling the function `update_dataset_linked_files()` with an empty list as as the second argument in the delete controller.

UPDATE: we should do the same from participants and families. The participant should be removed if it is not linked to another tissue sample (edited by Jenny), and the family should be removed if it has no more participants after current participant is deleted. Failing to do so will leave participants/families orphaned, which may prevent the codenames from being reused (if, for instance, someone tries to re-input an orphaned participant from a deleted dataset with a new family codename, it would come back as a dupe).

jennyziyi-xu:
Addresses issue #941. 

As a side note, this is how we handle `tissue_sample` when a dataset is deleted, in the master branch:[ link ](https://github.com/ccmbioinfo/stager/blob/e0bbcdf717e25fba568e0b8674f02031e5a2db6e/flask/app/datasets.py#L378)

This is how I handle it in this PR:[ link ](https://github.com/ccmbioinfo/stager/blob/9012059f07a199c8684fb07bb0eb3550ac089f11/flask/app/datasets.py#L378)

You can see that my approach is different in the sense that I don’t check `dataset_id` when looping through tissue_sample’s datasets. This is because I get rid of joinedLoad.

I prefer my approach because it's clearer as to what to expect, in the sense of SQLAlchemy queries. Feel free to let me know if the old approach is what we want. 

Closes #941

### DIFF
--- a/flask/app/datasets.py
+++ b/flask/app/datasets.py
@@ -389,9 +389,6 @@ def delete_dataset(id: int):
     participant = tissue_sample.participant
     family = participant.family
 
-    # Remove permission groups with the corresponding dataset ID in groups_datasets table
-    dataset.groups = []
-
     if not dataset.analyses:
         try:
             db.session.delete(dataset)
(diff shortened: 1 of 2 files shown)


kevinlul (approved):
LGTM!

PR #955, opened 2021-11-25
Docker: use single-threaded Gunicorn workers scaled to the number of cores

kevinlul:
Interim experimental solution to the server process choking out until we determine root cause


cklamann (approved):
LGTM!

PR #957, opened 2021-11-29
React: disable fetch to labs.json

cklamann:
If `labs.json` does not exist, the index page will be returned instead, which the processing code can't handle.


kevinlul (approved):
Emergency approved!

Lakshya2610 (approved):

PR #959, opened 2021-11-29
SQLAlchemy: Upgrade to version 1.4.27

jennyziyi-xu:
A new migration file is added to make `discriminator` non-nullable for class `Dataset`. Otherwise, querying the class `Dataset` would cause errors. The migration file also populates the `rnaseq_dataset` table accordingly. 

Other information: 

1, Before, we access a subquery's column directly as `subquery.column_name`. By upgrading to version 1.4.27, the column can no longer be found this way and will lead to errors. I have modified it to `subquery.c.column_name`.

Supporting argument: https://docs.sqlalchemy.org/en/14/orm/tutorial.html#returning-lists-and-scalars 

"Once we have our statement [subquery], it behaves like a Table construct, such as the one we created for users at the start of this tutorial. The columns on the statement are accessible through an attribute called c."

2, I have added `select` construct explicitly in certain places to suppress warning messages.


delvinso:
After debugging, the warning can be confirmed to be i) from L390 in `variants.py` and ii) only for the 'gene' search type. When the search type is 'gene', ie. using the ENSG ID to filter the variants, there is a filter placed on `gene.ensembl_id`. It looks like the problem is from the lack of an explicit join between the `Variant` and `Gene` tables, yielding the following SQL statement. 

```
SELECT count(DISTINCT variant.variant_id) AS count_1 
FROM variant, gene 
WHERE gene.ensembl_id IN (__[POSTCOMPILE_ensembl_id_1])

# apparently the comma in the FROM statement results in a cartesian product, hence the warning?: https://docs.sqlalchemy.org/en/14/changelog/migration_14.html#built-in-from-linting-will-warn-for-any-potential-cartesian-products-in-a-select-statement
# thanks to Jenny for the link
```
This can be fixed by using a join between`Variant` and `Gene` (because of the search strategy we employ, there is no explicit relationship between these two tables). This results in the following statement:

```
SELECT count(DISTINCT variant.variant_id) AS count_1 
FROM gene JOIN variant ON gene.chromosome = variant.chromosome AND gene.start <= variant.position AND variant.position <= gene."end" 
WHERE gene.ensembl_id IN (__[POSTCOMPILE_ensembl_id_1])

```

jennyziyi-xu:
Nice! It was confusing which line is causing the problem. I was focusing on the wrong place.

flask/app/analyses.py
```diff
@@ -278,6 +284,8 @@ def list_analyses(page: int, limit: int) -> Response:
         for analysis in analyses
     ]
 
+    app.logger.info(results)
```
cklamann, line 287: Stray logging statement?

cklamann (changes requested):
Delvin noticed that with this upgrade, queries will fail when they try to select records with a `null` `discriminator` column. Since most `dataset.discriminator` columns are `null`, we will need to update them with the appropriate values before this can be merged and probably make the column itself non-nullable. Changelog [here](https://github.com/sqlalchemy/sqlalchemy/blob/main/doc/build/changelog/changelog_14.rst). Relevant excerpt:
```
.. change::
        :tags: bug, orm
        :tickets: 4836

        An exception is now raised if the ORM loads a row for a polymorphic
        instance that has a primary key but the discriminator column is NULL, as
        discriminator columns should not be null.

delvinso (approved):
LGTM - works as expected!

cklamann (approved):

PR #971, opened 2021-12-01
React: reinstate lab fetch after routing fix

cklamann:


cklamann:
Closing for now -- will reopen/merge when the fix is in place for the dev instance as well.

PR #974, opened 2021-12-03
Flask: add marshmallow and a few schemas for input validation 

### LINKED ISSUE #218: Input validation
Right now if we do weird stuff like strings or arrays where ints are expected we have undesirable behaviour that's inconsistent between endpoints

Requires some research

delvinso:
part of #218 

- adds `flask-marshmallow` and `marshmallow-sqlalchemy` to requirements 
- schemas are stored in `schemas.py`
- for a first pass, schemas have been automagically constructed for `Dataset` and `Analysis` using the sqlalchemy models with custom validation depending on the endpoint. I tried using a decorator to automatically construct a schema when the models are created (the schema would be an attribute) but was getting some errors related to relationship definitions. Regardless it may not be the best idea anyways to use a decorator to construct the schema given that we do have quite a bit of custom endpoint logic (and therefore custom validation).
- as an added bonus enum validation seems to be taken care of since the schemas are constructed using the models via `SQLAlchemyAutoSchema` 

- All errors are now returned at once

```
curl -X POST localhost:5000/api/analyses \
  -H 'Content-Type: application/json' \
  -d '{"datasets": ["haha"], "pipeline_id": 1, "priority": "NOPE"}'
{
  "error": "{'datasets': {0: ['Not a valid integer.']}, 'priority': ['Must be one of: Clinical, Research.']}"
}
```

- missing schemas
- [ ] users
- [ ] groups
- [ ] _bulk


kevinlul:
Neat, this isn't as complicated as I expected!

jennyziyi-xu (changes requested):
A minor request: `validate_enums` doesn't need to be imported anymore in most python files. 

Is there a need to use marshmallow for input validation in the `bulk` POST endpoint?

flask/app/schemas.py
```diff
@@ -0,0 +1,84 @@
+from marshmallow_sqlalchemy import SQLAlchemyAutoSchema
+from marshmallow import fields, Schema
```
jennyziyi-xu, line 2:
```suggestion
from marshmallow import fields
```

jennyziyi-xu (changes requested):

delvinso:
> A minor request: `validate_enums` doesn't need to be imported anymore in most python files.
> 
> Is there a need to use marshmallow for input validation in the `bulk` POST endpoint?

I think the decision was to split the input validation into 2 PRs since the `_bulk`, `users` and `groups` are more complicated due to the custom validation we use.

jennyziyi-xu (approved):

PR #975, opened 2021-12-06
Flask/React: update /api/unlinked to query by prefix only, debounce requests on frontend

hannahle:
- We've seen that when the unlinked file payload is very large (>2MB), it takes a long time for all files to be returned, leading to a laggy user experience. 
- I modified the `get_unlinked_files` function to only return results based on prefixes. This is per Hannah Driver's feedback since she usually already knows what files she has to pass in, and uses the prefix in stager as a way to type in the right file faster and more accurately.
- The frontend was also modified to only return results based on the prefixes that user types.

### DIFF
--- a/flask/app/buckets.py
+++ b/flask/app/buckets.py
@@ -16,59 +16,56 @@ bucket_blueprint = Blueprint(
 @bucket_blueprint.route("/api/unlinked/<prefix>", methods=["GET"])
 @login_required
 def get_unlinked_files(prefix: str):
-
-    app.logger.debug("this is prefix backend", prefix)
-
-    minio_client = get_minio_client()
-
-    app.logger.debug("Getting all minio bucket names..")
-    all_bucket_names = [bucket.name for bucket in minio_client.list_buckets()]
-    app.logger.debug("All bucket names", all_bucket_names)
-
-    # Remove buckets the current user does not have access to assuming group_code.lower() == bucket name
-    app.logger.debug("Getting all buckets that user has access to..")
-    user = (
-        models.User.query.filter_by(user_id=current_user.user_id)
-        .options(joinedload(models.User.groups))
-        .first()
-    )
-
-    valid_bucket_names = []
-    for group in user.groups:
-        code = group.group_code.lower()
-        if code in all_bucket_names:
-            valid_bucket_names.append(code)
-
-    app.logger.debug("Getting all files in valid minio buckets..")
-    all_files = []
-    for bucket in valid_bucket_names:
-        objs = minio_client.list_objects(bucket, prefix=prefix, recursive=True)
-        for obj in objs:
-            all_files.append(bucket + "/" + obj.object_name)
-
-    app.logger.debug("Getting all linked files..")
-
     files = []
 
-    linked_files = {
-        f.path: True
-        for f in models.File.query.filter(
-            or_(models.File.multiplexed == None,
-                models.File.multiplexed == False)
-        ).all()
-    }
+    if prefix != "None":
 
-    linkable_files = {
-        f.path: True
-        for f in models.File.query.filter(models.File.multiplexed == True).all()
-    }
+        minio_client = get_minio_client()
 
-    for file_name in all_files:
-        if linkable_files.get(file_name):
-            files.append({"path": file_name, "multiplexed": True})
-        elif not linked_files.get(file_name):
-            files.append({"path": file_name, "multiplexed": False})
+        app.logger.debug("Getting all minio bucket names..")
+        all_bucket_names = [bucket.name for bucket in minio_client.list_buckets()]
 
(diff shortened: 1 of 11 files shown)


react/src/components/FileLinkingComponent.tsx
```diff
@@ -47,11 +45,18 @@ const useAutocompleteStyles = makeStyles(() => ({
...
+    setFiles?: React.Dispatch<React.SetStateAction<UnlinkedFile[]>>;
     disabled?: boolean;
     disableTooltip?: boolean;
-}> = ({ values, options, onEdit, disabled, disableTooltip }) => {
+    inputValue?: string;
+    onInputChange?: (newInputvalue: string) => void;
+}> = ({ values, onEdit, disabled, disableTooltip, inputValue, onInputChange }) => {
+    const debouncedSearchQuery = useDebounce(inputValue || "None", 600);
+
+    const files = useUnlinkedFilesQuery(debouncedSearchQuery);
```
kevinlul, line 57: So for debouncing here, you changed both the frontend and the backend so that the frontend will still send a request, just to `/api/unlinked/None`, which the backend will ignore?

hannahle: Yes, that's right.

hannahle: This is mainly because we can't use the custom hook inside a conditional statement on the frontend, so a request would still need to be sent, but the backend would simply ignore it.

react/src/hooks/useDebounce.ts
```diff
@@ -0,0 +1,18 @@
+import React from "react";
+
+export function useDebounce(value: string, delay: number = 500) {
```
kevinlul, line 3: Was this copied from somewhere? The default argument isn't used.

flask/app/buckets.py
```diff
@@ -17,54 +17,80 @@
...
-
-    app.logger.debug("Getting all files in valid minio buckets..")
-    all_files = []
-    for bucket in valid_bucket_names:
-        objs = minio_client.list_objects(bucket, recursive=True)
-        for obj in objs:
-            all_files.append(bucket + "/" + obj.object_name)
-
-    app.logger.debug("Getting all linked files..")
+    path = request.args.get("path", type=str)
```
kevinlul, line 20: Please set a default value here or you'll error out on a plain `/api/unlinked`. Otherwise input validation doesn't have to be in scope of this PR since it wasn't a thing here originally.

flask/app/buckets.py
```diff
@@ -17,54 +17,80 @@
...
-        elif not linked_files.get(file_name):
-            files.append({"path": file_name, "multiplexed": False})
-
-    app.logger.debug("Returning JSON array..")
+    if path != "None":
+        filepath = path.split("/", 1)
+        if len(filepath) < 2:
+            abort(
+                400,
+                description="Please make sure your file path contains both folder and file name, such as c4r/folder1/folder2/filename",
```
kevinlul, line 29: Please call the buckets and prefixes for what they are, otherwise it could lead to confusion.

kevinlul (changes requested):
Okay, I think I understand the debouncing now. It's a separate state that only updates if no changes occur within the specified interval, and then that will actually trigger in the request. I don't think the `None` case is needed at all — you can just default to blank, that request only gets sent out if you type something and then delete it.

hannahle:
> Okay, I think I understand the debouncing now. It's a separate state that only updates if no changes occur within the specified interval, and then that will actually trigger in the request. I don't think the `None` case is needed at all — you can just default to blank, that request only gets sent out if you type something and then delete it.

Makes sense. I'll add this in.

react/src/AddDatasets/components/utils.tsx
```diff
@@ -153,8 +145,8 @@ export function getOptions(
         case "solved":
             return booleans.map(b => toOption(b, "Is Solved"));
 
-        case "linked_files":
-            return files;
+        // case "linked_files":
+        //     return files;
```
kevinlul, line 149: We can just delete since we're on Git.

flask/app/buckets.py
kevinlul: This comment wasn't addressed!

kevinlul (changes requested):

flask/app/buckets.py
hannahle: To clarify, I changed the variable names "bucket" to "bucket_name" and "prefix" to "file_name". Was this not what you meant?

kevinlul: I was looking at the error message here.

hannahle: Ok, cool. Just changed the error message!

kevinlul (approved):
LGTM!

PR #979, opened 2021-12-08
Flask/Minio: add MINIO_TLS env var to docker-compose.ccm

cklamann:


docker-compose.ccm.yaml
```diff
@@ -29,6 +29,7 @@ services:
       MINIO_ENDPOINT: "${HIRAKI_MINIO_ENDPOINT}"
       MINIO_ACCESS_KEY: "${HIRAKI_MINIO_ACCESS_KEY}"
       MINIO_SECRET_KEY: "${HIRAKI_MINIO_SECRET_KEY}"
+      MINIO_TLS: true
```
kevinlul, line 32: This has to be quoted or numeric, or it gets interpreted by the YAML parser as a boolean, which isn't really a thing for a environment variables.

kevinlul (approved):
🙏

Issue #950, opened 2021-11-24
Flask/DevOps: Intermittent outages


cklamann:
Users have observed intermittent outages of the STAGER backend, with the server returning a `504`. Users report that these outages started a few weeks ago and weren't observed previously. Anecdotally, it seems that concurrency might be a factor, but we'll probably need additional logging to know more.

kevinlul:
#955

kevinlul:
Also somewhat related is #754, if query efficiency is a major contributor

PR #983, opened 2021-12-09
Flask: add Prometheus endpoint

kevinlul:
This is to help gain insights for #950.
Leverage [prometheus-flask-exporter](https://github.com/rycus86/prometheus_flask_exporter), a high-level client atop Prometheus' Python client designed for Flask and Gunicorn.
Use `/tmp` for the multiprocess directory to temporarily write metrics for the manager process, which is always writable and reset when the container is recreated, as required.
Add alternate gunicorn configuration for testing this metrics setup without going straight into production.

Integrate entrypoints into Dockerfile, now that stager-dev is no longer needed.
Disable pip cache when building images to reduce image size.


kevinlul:
Merge conflict due to #974 resolved

cklamann (approved):
LGTM!

PR #1005, opened 2022-01-04
Docker: switch to Traefik for multitenant deployment

### LINKED ISSUE #976: Switch to Traefik in the multitenant deployment
Follow-up to #909 / #973

The deployment configuration is a little unwieldly if we need to add tenants: secrets, compose, and nginx config all need to be updated. This should be an ideal use of Traefik, which uses container labels to route.

This should also phase out the repo needing to be cloned to the deployment host at /opt/stager

kevinlul:
This is already working on the real site. Please check if the configuration makes sense and if additional documentation is needed beyond [Traefik's own](https://doc.traefik.io/traefik/).

Closes #976

### DIFF
--- a/docker-compose.ccm.yaml
+++ b/docker-compose.ccm.yaml
@@ -30,14 +30,15 @@ services:
       MINIO_SECRET_KEY: "${HIRAKI_MINIO_SECRET_KEY}"
     labels:
       - traefik.enable=true
-      - traefik.http.routers.app.rule=Host(`stager-hiraki.ccm.sickkids.ca`) # && PathPrefix(`/api`)
-      - traefik.http.routers.app.entrypoints=https
-      # Assume this works for now, need to confirm with domains
+      - traefik.http.routers.app-hiraki.rule=Host(`stager-hiraki.ccm.sickkids.ca`) # && PathPrefix(`/api`)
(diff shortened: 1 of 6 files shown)


docker-compose.ccm.yaml
```diff
@@ -37,12 +48,33 @@ services:
   #     MINIO_ENDPOINT: "${HAWKINS_MINIO_ENDPOINT}"
   #     MINIO_ACCESS_KEY: "${HAWKINS_MINIO_ACCESS_KEY}"
   #     MINIO_SECRET_KEY: "${HAWKINS_MINIO_SECRET_KEY}"
-  nginx:
-    image: nginx:alpine
+  #   labels:
+  #     - traefik.enable=true
+  #     - traefik.http.routers.app-hawkins.rule=Host(`stager-hawkins.ccm.sickkids.ca`) # && PathPrefix(`/api`)
```
delvinso, line 53: I assume these should stay in anticipation for the Hawkins tenancy in the future?

delvinso (approved):
LGTM! Traefik documentation seems self-explanatory

docker-compose.ccm.yaml
kevinlul: Yes, and as an example of how to add another tenant.

PR #1010, opened 2022-01-07
React: Handle NetworkError and raised exceptions. 

jennyziyi-xu:
I simulated NetworkError by hardcoding the endpoint to be `localhost:5000` in `apiFetch()`.  I am open to any feedback. 

1. Handle NetworkError raised by `apiFetch()`

- `useEffect()` in `App.tsx`: When encountering a NetworkError, pass a prop to `LoginPage` component to display an error message and to disable the Login button. 
- `signout()` in `App.tsx`: Show an error message in the navigation bar. 
- `OIDCRedirectHandler()` in `Login.tsx`: `setError()`
- `authenticate()` in `Login.tsx` : `setError()`
- `sendFile()`: `enqueueSnackbar()`

2. Handle exceptions raised by `basicFetch()`, called by the following functions 

- `useEnumsQuery.tsx` : Not handling as of now.  
- ` useInsitutionsQuery.tsx `: enqueueSnackbar
- `useMetadatasetTypesQuery.tsx ` : enqueueSnackbar
- `usePipelinesQuery.tsx ` : enqueueSnackbar 
- `useAnalysesQuery.tsx ` : enqueueSnackbar 
- `useAnalysisQuery.tsx ` : enqueueSnackbar 
- `useBulkCreateMutation.tsx` : enqueueSnackbar in `AddDatasets/index.tsx`
- `useDatasetQueries.tsx ` : enqueueSnackbar 
- `useDatasetQuery.tsx`: enqueueSnackbar 
- `useDatasetsQuery.tsx`: useDatasetsQuery is not called anywhere, deleted this file.  
- `useFamiliesQuery.tsx`: enqueueSnackbar  
- `useGenesQuery.ts`: enqueueSnackbar  
- `useGroupQuery.tsx`: enqueueSnackbar in  `Groups/index.tsx`
- `useGroupsQuery.tsx`: enqueueSnackbar in `CreateUserModal.tsx` and  `Groups/index.tsx`, not handled in `UserDetails.tsx`
- `useParticipantQuery.tsx`: enqueueSnackbar  
- `useUnlinkedFilesQuery.tsx` : not handled because onError is already specified in `FileLinkingComponent.tsx `
- `useUserQuery.tsx`: enqueueSnackbar in  `Settings/index.tsx`, not handled in `UserDetails.tsx` to not be redundant with `useUsersQuery`
- `useUsersQuery.tsx`: enqueueSnackbar in  `UserList.tsx `


react/src/hooks/users/useUserQuery.tsx
```diff
@@ -12,7 +12,9 @@ async function fetchUser(username: string) {
  * That is, return all details about a particular user,
  * including MinIO credentials.
  */
-export function useUserQuery(username: string) {
-    const result = useQuery<User, Response>(["users", username], () => fetchUser(username));
+export function useUserQuery(username: string, options: UseQueryOptions<User, Response> = {}) {
+    const result = useQuery<User, Response>(["users", username], () => fetchUser(username), {
+        ...options,
+    });
```
kevinlul, line 18: You should be able to pass `options` through, no need for a spread copy.

react/src/hooks/groups/useGroupsQuery.tsx
```diff
@@ -11,8 +11,8 @@ async function fetchGroups() {
  *
  * That is, return an array of all permission groups as name-code pairs.
  */
-export function useGroupsQuery() {
+export function useGroupsQuery(options: UseQueryOptions<Group[], Response> = {}) {
     // all groups
-    const result = useQuery<Group[], Response>("groups", fetchGroups);
+    const result = useQuery<Group[], Response>("groups", fetchGroups, { ...options });
```
kevinlul, line 16: ditto

kevinlul (changes requested):
Instead of displaying `Network Error!` to users, we can choose a more helpful sentence, like `The backend server seems to be unavailable. Please try again later.`

react/src/hooks/users/useUserQuery.tsx
jennyziyi-xu: I am passing `options` here because `useUserQuery` is called in `Settings/index.tsx` and `UserDetails.tsx`. I would like to handle the error in `Settings/index.tsx` but not in `UserDetails.tsx` because `useUsersQuery` is already called in `UserDetails.tsx`. I am handling network error caused by `useUsersQuery`. 

What do you think about this?

kevinlul: I mean that you should be able to directly pass in `options` as an argument. You shouldn't need to make a copy with `{ ...options }` as you aren't modifying any properties.

jennyziyi-xu: Sorry, I get what you mean.

kevinlul:
image

kevinlul (commented):
This is better than we had before, however, if the login endpoint returns an error 500, the login page is still blanked out. (Reproduce by starting the React app without a backend server up. The dev server proxy will be unable to proxy and returns a 500.)

jennyziyi-xu:
> This is better than we had before, however, if the login endpoint returns an error 500, the login page is still blanked out. (Reproduce by starting the React app without a backend server up. The dev server proxy will be unable to proxy and returns a 500.)

You are right, thanks! I have added that `if not (response.ok)`, show the error message and disable the user from signing in. According to [Fetch API ](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API), "The Promise returned from fetch() won’t reject on HTTP error status even if the response is an HTTP 404 or 500. Instead, it will resolve normally (with ok status set to false), and it will only reject on network failure or if anything prevented the request from completing."

kevinlul (approved):

kevinlul:
Reverted in #1018, we'll need to look at this again.

PR #1012, opened 2022-01-10
Flask: fix admins not having MinIO delete permissions

kevinlul:
Deny overrides Allow, so change the per-group policy to use explicit Allows. All policies: https://docs.min.io/minio/baremetal/security/minio-identity-management/policy-based-access-control.html

I will run `migrate-minio-policies` after merging.


delvinso (approved):
LGTM!

PR #1014, opened 2022-01-11
Flask: fix /api/unlinked to allow for prefix searches

hannahle:
Before, we could only search if a file exists by full file path. 

Now, if user types in a bucket name followed by a file prefix, they should be able to see a list of files under that bucket. For example, typing in "c4r/exome" should show results such as "c4r/exome/testfile1.bam", "c4r/exome/testfile2.bam", etc.


kevinlul (approved):
LGTM!

PR #1016, opened 2022-01-14
Flask: Input validation for _bulk, create_user and create_group endpoints 

jennyziyi-xu:
Marshmallow input validation is added for these endpoints

- [ ] ` _bulk [POST]`
- [ ] `users [POST]`
- [ ] `groups [POST]`

- `unknown=EXCLUDE` is added to the schemas so that in the `_bulk` endpoint, unknown fields are ignored by the input validation. Documentation for this schema option: [link](https://marshmallow.readthedocs.io/en/stable/api_reference.html#marshmallow.Schema). 
- For the email validation in users, marshmallow has [built-in email validation functionality](https://marshmallow.readthedocs.io/en/stable/_modules/marshmallow/validate.html#Email) that we might consider.
- Once Marshmallow input validation is added for the PATCH endpoints, we can delete `verify_email` in `users.py` because it would be handled in `schemas .py`.
- Input validation for PATCH endpoints will be handled in another issue. 

[Edit Jan 21st] 

- `unknown=EXCLUDE` is removed. 
- Before performing input validation, the payload is processed by first checking if the fields in the payload are model attributes. In addition, some fields are filtered out because they are a collection of items created by `db.relationship()` so that they are not generated by Marshmallow schemas. If these fields were not filtered out, Marshmallow validation will raise errors because these are unknown fields.


jennyziyi-xu:
Thank you, Delvin for pointing out that using `unknown=EXCLUDE` is not explicit, and it’s not clear what we are validating if we just ignore unknown fields. There are a few alternative choices I can think of: 

1. Remove `unknown=EXCLUDE` for all the schemas. Define a schema for the `bulk` endpoint alone. This schema wouldn't be autogenerated, meaning that each Field's specification would be hardcoded. If we want the schema to be autogenerated, we would need to create a Python class first, which we can do but I don't know if this is too much work just for the `bulk` endpoint. 
2. Remove `unknown=EXCLUDE` for all the schemas. In the `bulk` endpoint, filter out the irrelevant fields for each model and then use marshmallow schemas for validation. This means that the DatasetSchema must work for both the POST endpoint for Dataset and the `bulk` endpoint, excluding the option for customization. 

What are everyone's thoughts?

flask/tests/test_groups.py
```diff
@@ -210,7 +210,7 @@ def test_create_group(test_database, client, login_as, minio_admin):
             "/api/groups",
             json={"group_code": "sk", "group_name": "Invalid s3 bucket name"},
         ).status_code
-        == 422
+        == 400
```
kevinlul, line 213: Why did these two status codes change?

flask/app/routes.py
```diff
@@ -431,28 +408,8 @@ def bulk_update():
             app.logger.debug(
                 "\tFamily ID '%s' already exists and won't be created.", family_id
             )
+        row["family_id"] = family_id
```
kevinlul, line 411: Do we need to keep adding these in to make sure subsequent schemas in the hierarchy validate?

jennyziyi-xu: As defined in `models.py`, `family_id` is a non-nullable field for the model `Participant`. When the schema is auto-generated for `Participant`, `family_id` thus becomes a `required` field. That's why I added `row["family_id"] = family_id` in order for input validation to be performed for  `Participant`. 

The same goes for the other insertions into the `row` dictionary.

flask/tests/test_groups.py
jennyziyi-xu: It is specified in `schemas.py` that `group_name` must satisfy a regex that doesn't contain spaces:[ link](https://github.com/ccmbioinfo/stager/blob/659a93a6745f0a7f5dd734f7e9950e3b0900a742/flask/app/schemas.py#L92). 

Marshmallow input validation catches the error and throws 400 code here: [link.](https://github.com/ccmbioinfo/stager/blob/659a93a6745f0a7f5dd734f7e9950e3b0900a742/flask/app/groups.py#L183)

jennyziyi-xu:
`unknown=EXCLUDE` is removed.
Before performing input validation, the payload is processed by first checking if the fields in the payload are model attributes. In addition, some fields are filtered out because they are a collection of items created by `db.relationship()` so that they are not generated by Marshmallow schemas. If these fields were not filtered out, Marshmallow validation will raise errors because these are unknown fields.

jennyziyi-xu:
This [CodeQL error](https://github.com/ccmbioinfo/stager/security/code-scanning/22?query=ref%3Arefs%2F[…]ull%2F1016%2Fhead+ref%3Arefs%2Fheads%2Fjenny%2Finput-validation) appeared earlier today with my commits. This CodeQL error also appeared 7 hours ago on the master branch. 

My latest commit fixed this error according to CodeQL's suggestion. I don't know if my fix is enough, and whether this security error will affect production.

kevinlul:
madmin regularly gets flagged by CodeQL, but they are false positives in our codebase because of the very limited ways that we use that codepath. The user input is the group code, but it is already validated to be S3-compliant and thus cannot contain anything problematic when used as a command-line argument or a filesystem path. In production, the backend is also running as a non-root user in a container and cannot modify its own code either.

jennyziyi-xu:
Ok got it, yes I agree that adding this check is not that useful. I will revert back.

flask/app/schemas.py
```diff
@@ -36,30 +39,37 @@ class TissueSampleSchema(SQLAlchemyAutoSchema):
...
 
 
-class DatasetSchema(SQLAlchemyAutoSchema):
+class RNASeqDatasetSchema(SQLAlchemyAutoSchema):
     """
     POST /api/datasets
-    - a valid tissue_sample_id is required for the dataset to be linked
-    - linked_files must be a valid list
     - created_by_id and updated_by_id is not given as part of the request body and should therefore be excluded from validation
+    - discriminator field is populated automatically when a new dataset is inserted into the database.
```
delvinso, line 51: I think it would be helpful to clarify why `RNASeqDataset` (both model and name of the class) is used here instead of `Dataset` for transparency

delvinso (approved):

PR #1017, opened 2022-01-14
CI: deploy Aleixo Muise lab tenant

kevinlul:
Halve the number of Gunicorn workers as to divide the number of processes (ish) between tenants.

### DIFF
--- a/docker-compose.ccm.yaml
+++ b/docker-compose.ccm.yaml
@@ -39,6 +39,20 @@ services:
       - traefik.http.middlewares.app.headers.accessControlAllowCredentials=true
       - traefik.http.middlewares.app.headers.frameDeny=true
       - traefik.http.middlewares.app.headers.browserXssFilter=true
+  app-muise:
+    <<: *app
+    environment:
+      <<: *env
+      ST_SECRET_KEY: "${MUISE_ST_SECRET_KEY}"
+      ST_DATABASE_URI: "mysql+pymysql://${MUISE_MYSQL_CONNECTION_STRING}"
+      MINIO_ENDPOINT: "${MUISE_MINIO_ENDPOINT}"
+      MINIO_ACCESS_KEY: "${MUISE_MINIO_ACCESS_KEY}"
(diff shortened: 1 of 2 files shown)


kevinlul:
working on adding the secrets atm, should be able to deploy before end of day

kevinlul:
Okay, should be good to go! Merge this for backend, and then #1010 or $1015 for frontend.

docker-compose.ccm.yaml
```diff
@@ -39,6 +39,20 @@ services:
...
+    environment:
+      <<: *env
+      ST_SECRET_KEY: "${MUISE_ST_SECRET_KEY}"
+      ST_DATABASE_URI: "mysql+pymysql://${MUISE_MYSQL_CONNECTION_STRING}"
+      MINIO_ENDPOINT: "${MUISE_MINIO_ENDPOINT}"
+      MINIO_ACCESS_KEY: "${MUISE_MINIO_ACCESS_KEY}"
+      MINIO_SECRET_KEY: "${MUISE_MINIO_SECRET_KEY}"
+    labels:
+      - traefik.enable=true
+      - traefik.http.routers.app-hawkins.rule=Host(`stager-muise.ccm.sickkids.ca`) # && PathPrefix(`/api`)
```
hannahle, line 53: For these 3 lines, I think you meant to put `app-muise` as opposed to `app-hawkins`.

kevinlul: Oops, good catch.

docker-compose.ccm.yaml
```diff
@@ -39,6 +39,20 @@ services:
       - traefik.http.middlewares.app.headers.accessControlAllowCredentials=true
       - traefik.http.middlewares.app.headers.frameDeny=true
       - traefik.http.middlewares.app.headers.browserXssFilter=true
```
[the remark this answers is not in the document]
kevinlul: The container labels are equivalent to declaring these in separate YAML files like in `traefik/`
Since the `app` middleware is already declared here, we can reuse it elsewhere as the configuration elements aren't actually scoped to the container but merged into one.

PR #1026, opened 2022-01-27
React: fix security warnings

kevinlul:
All resolved by re-resolving CRA dependencies (yarn upgrade react-scripts), though the vulnerabilities aren't actually on the live codepath.

Can confirm with `yarn audit` that there are no complaints now.

### DIFF
--- a/react/yarn.lock
+++ b/react/yarn.lock
@@ -18,25 +18,25 @@
   dependencies:
     "@babel/highlight" "^7.16.7"
 
-"@babel/compat-data@^7.13.11", "@babel/compat-data@^7.16.4":
-  version "7.16.4"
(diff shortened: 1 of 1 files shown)


jennyziyi-xu:
We are not supposed to see any vulnerabilities right? 
Screen Shot 2022-01-28 at 1 40 34 PM

kevinlul:
I guess more got flagged since. Again, they aren't really on used codepaths. The first needs CRA to update SVGR per facebook/create-react-app#11780. The latter needs CRA to update [resolve-url-loader](https://github.com/bholloway/resolve-url-loader/releases/tag/5.0.0) one major version as well.

jennyziyi-xu (approved):
LGTM.

PR #1027, opened 2022-01-27
Flask: make most initializations of MinIO client SDKs consistent

kevinlul:
Fixes some click commands not working because MINIO_TLS wasn't considered


jennyziyi-xu (approved):
LGTM!

PR #1028, opened 2022-01-27
Flask: upgrade to Python 3.8 

jennyziyi-xu:
The use of `@patch()` from `unittest.mock` module for mocking request context does not raise any problem when using Python3.7. By switching to Python 3.8, we will get "RuntimeError: Working outside of request context." 

Potential causes: 
1, Python 3.8 has added support for asynchronous code testing. This runtime error might be due to the added support for mocking asynchronous code in Python 3.8 as explained here: [link](https://github.com/pallets/flask/issues/3637#issuecomment-655372502).
2, `@patch` namespace resolution. 

The solution in this PR: 
Use [Flask testing approaches](https://flask.palletsprojects.com/en/2.0.x/testing/#other-testing-tricks) for `test_must_pass_in_user_argument_if_login_disabled`, `test_validate_json_throws_415_on_non_json_header` and `test_validate_json_passes_on_json_header`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Production image. Runs a Gunicorn WSGI server.
 FROM minio/mc:RELEASE.2021-12-10T00-14-28Z AS mc
-FROM python:3.7-slim
+FROM python:3.8-slim
 ARG GIT_SHA
 LABEL org.opencontainers.image.title Stager production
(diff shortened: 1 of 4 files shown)


kevinlul:
There's an additional Dockerfile in the repository root used to build the prod image. That should be bumped as well to match.

kevinlul (approved):
LGTM!

PR #1029, opened 2022-01-27
Flask: made unlinked file query non-recursive to reduce payload

hannahle:
The goal of this PR is to reduce the `unlinked` file's endpoint payload. 

The main changes in this PR are:
- Set `recursive` property to `false`
- Added a check to see if an option is a file or folder. If it's a folder, autocomplete the folder path and sends off a query. If not, then the file is selected as an unlinked file.

### DIFF
--- a/flask/app/buckets.py
+++ b/flask/app/buckets.py
@@ -36,9 +36,11 @@ def get_unlinked_files():
 
         # Check if the bucket in file path exists
         app.logger.debug("Getting all minio bucket names..")
-        all_bucket_names = [bucket.name for bucket in minio_client.list_buckets()]
+        all_bucket_names = [
+            bucket.name for bucket in minio_client.list_buckets()]
         if bucket_name not in all_bucket_names:
-            abort(400, description=f"Bucket name must be one of {all_bucket_names}")
+            abort(
+                400, description=f"Bucket name must be one of {all_bucket_names}")
 
         app.logger.debug("Getting all buckets that user has access to..")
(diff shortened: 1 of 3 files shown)


jennyziyi-xu:
The user cannot start the input path with a "/". If the user inputs `/cheo/`, the error message will be displayed: "Bucket name must be one of ...". In this case, the user might not know where is the problem because "cheo" is among the list of suggested bucket names. It might be better to allow the flexibility of this leading slash, although this is a minor issue.

kevinlul:
We didn't support the leading slash to begin with; it's syntactically incorrect.

kevinlul:
This approach seems effective, but behaviour feels a little funky when just using the keyboard to navigate. Suggested text in the textbox and the options sometimes disappear when moving around with arrow keys.

jennyziyi-xu:
For me, when just using the keyboard to navigate (left and right arrow keys), the options disappear but would come back if I press up or down arrow keys.

kevinlul:
I'm not expecting the dropdown to change. I thought it might be because of the altered stale time but that doesn't seem to be it.

hannahle:
> I'm not expecting the dropdown to change. I thought it might be because of the altered stale time but that doesn't seem to be it.

Ah nope, the stale time just makes the data "stale" after 1s and allows for a re-query.

kevinlul (approved):
Okayed by collabs 🙏 🎉 ⌛

PR #1042, opened 2022-02-03
Flask: Upgrade to python 3.8 (again)

jennyziyi-xu:
This PR is a continuation of #1028 . In production, `wsgi` import raised an error because the module import `typing_extensions` in `schemas.py` can't be found. With this PR, simulating the production environment with `docker-compose up app_gunicorn` raised no errors.


kevinlul (approved):
Let's give this another shot 🤞

kevinlul:
🎉

PR #1047, opened 2022-02-07
Traefik: switch internal API, metrics to HTTPS/8443, enable HSTS

kevinlul:
Reverse proxying to MinIO forwarded HSTS headers, so we'll implement HTTPS and HSTS for every service as it affects the entire primary domain already.

Refactor all HTTP security-related headers into the `security` middleware to reuse across all routers.

Configure the Traefik entrypoint to use port 8443 instead of 8080 and enable TLS. This will serve the main Traefik API and dashboard on the main domain, and proxy to the Prometheus endpoint of each backend on the corresponding backend domains. To implement this, an additional router is configured per service for the 8080 Gunicorn metrics port, since Traefik only automatically configures one service (port) per container.

### DIFF
--- a/docs/production.md
+++ b/docs/production.md
@@ -1,13 +1,12 @@
 # Production
 
-Stager is deployed at https://sampletrackerdev.ccm.sickkids.ca via Compose.
+Stager is deployed via Compose at https://stager.genomics4rd.ca to serve Care4Rare as part of Genomics4RD.
+A separate multitenant deployment serves other collaborators at SickKids at https://stager.ccm.sickkids.ca.
 
-## [nginx](https://hub.docker.com/_/nginx)
-
-An nginx container is used to serve the static frontend files and as a reverse proxy to the
-application server and MinIO. It is also solely responsible for HTTPS/TLS. Before starting up,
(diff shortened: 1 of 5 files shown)


kevinlul:
Need to confirm that the app container-specific changes are effective

docs/production.md
```diff
@@ -1,50 +1,80 @@
...
-```bash
-docker-compose -f docker-compose.prod.yaml ps
-```
+The Compose configuration is at [docker-compose.ccm.yaml](/docker-compose.ccm.yaml). For the time being,
+we must clone the repository to the production host in order to have the Traefik dynamic configuration and
+TLS certificates mounted in `traefik/certs`.
 
-Currently, a typical redeployment scenario involves manually performing these steps with some
-downtime. Updated images may need to be pulled and the stale database bind mount reset.
+Traefik is also serves a separate dashboard and its own metrics on a custom port. It is configured
```
delvinso, line 79: 'is' - typo?

kevinlul: Good catch

delvinso (approved):
LGTM

PR #1067, opened 2022-03-02
React: Automatically display candidate genes 

hannahle:
When uploading an RNAseq dataset to Stager (i.e. datasets type 'RRS'), an additional field 'Vcf available appears'. 

However, to see the 'Candidate genes' field you have to click on the show/hide fields icon and select that field. 

This PR makes it it possible to have this field shown automatically when the RRS dataset type is selected, just like the 'Vcf available' field.

### DIFF
--- a/react/src/AddDatasets/index.tsx
+++ b/react/src/AddDatasets/index.tsx
@@ -344,7 +344,9 @@ export default function AddDatasets() {
             currCols,
             RNA_ONLY_FIELDS.map(field => ({
                 field,
-                hidden: OPTIONAL_RNA_FIELDS.includes(field) && field !== "vcf_available",
+                hidden:
+                    OPTIONAL_RNA_FIELDS.includes(field) &&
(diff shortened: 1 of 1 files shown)


react/src/AddDatasets/index.tsx
```diff
@@ -344,7 +344,9 @@ export default function AddDatasets() {
             currCols,
             RNA_ONLY_FIELDS.map(field => ({
                 field,
-                hidden: OPTIONAL_RNA_FIELDS.includes(field) && field !== "vcf_available",
+                hidden:
+                    OPTIONAL_RNA_FIELDS.includes(field) &&
+                    !["vcf_available", "candidate_genes"].includes(field),
```
kevinlul, line 349: I think we can actually just leave this off as `OPTIONAL_RNA_FIELDS = ["vcf_available", "candidate_genes"]`

kevinlul (changes requested):

kevinlul (approved):

PR #1068, opened 2022-03-03
React: Expose button to delete analysis on the front-end

### LINKED ISSUE #1051: Expose delete button for analyses to admins
We already have the `DELETE /api/analyses/:id` endpoint for admins only, but it's not exposed in the frontend. Add another button on the analysis page or modal visible to admins for this action. It should give an appropriate warning like the dataset delete button.

hannahle:
Closes #1051

### DIFF
--- a/react/src/Analyses/index.tsx
+++ b/react/src/Analyses/index.tsx
@@ -433,6 +433,7 @@ export default function Analyses() {
                     onClose={() => {
                         setDetail(false);
                     }}
+                    refreshTable={tableRef.current.onQueryChange}
                 />
             )}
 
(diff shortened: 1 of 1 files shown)


react/src/Analyses/index.tsx
```diff
@@ -621,6 +669,14 @@ export default function Analyses() {
...
                             },
                         },
+                        {
+                            icon: Delete,
+                            tooltip: "Delete Analysis",
+                            onClick: (event, rowData) => {
+                                setActiveRows(rowData as Analysis[]);
+                                setConfirmDelete(true);
+                            },
+                        },
```
kevinlul, line 679: Can you hide this if the user is not an admin?

kevinlul (changes requested):

react/src/Analyses/index.tsx
hannahle: On it!

kevinlul: Sorry, I meant to physically hide the button, which is done on the datasets page with its delete button.

kevinlul (approved):
LGTM 🚀

PR #1076, opened 2022-03-17
Flask: Fix Marshmallow validation error logging 

jennyziyi-xu:
Error messages from Marshmallow validation are jsonified for logging. They should not be jsonified to output a human-readable error message in logging.


kevinlul (approved):
Thanks for catching this!

PR #1080, opened 2022-03-24
CI: add workflow to validate Docker Compose files

kevinlul:
Fix bad indentation for one container in #1078 and add a check so this doesn't happen again.


hannahle (approved):
Looks good to me!

hannahle:
This extension is also useful for syntax highlighting/validation in VSCode: https://marketplace.visualstudio.com/items?itemName=redhat.vscode-yaml

PR #1083, opened 2022-03-24
Docker: fix sidecar volume permissions

kevinlul:
Volumes behave like the `mount` command, with the local filesystem permissions at the mount path by default. If the path does not already exist, then it gets created inside the container owned by root:root. Switch to /var/tmp, which is writable by www-data. Alternatively, we could mkdir and chown an appropriate path in the Dockerfile. Also switch busybox to the same UID rather than root.

### DIFF
--- a/docker-compose.ccm.yaml
+++ b/docker-compose.ccm.yaml
@@ -22,6 +22,7 @@ x-env: &env
 x-sidecar: &sidecar
   image: busybox
   init: true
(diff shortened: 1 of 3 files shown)


hannahle (approved):
LGTM!

kevinlul:
working for CHEO-RI

kevinlul:
working for CCM, both backends

PR #1108, opened 2022-04-13
React: Fix the bug in dataset modal to get options for Dataset types

### LINKED ISSUE #1100: Bug in dataset modal for datasets of type RES
Using seed data, on the Datasets page, when we toggle the edit mode in the Dataset modal with a dataset of type `RES`, the web interface will go blank and the following error will be shown: 
Screen Shot 2022-04-12 at 11 51 12 AM
Screen Shot 2022-04-12 at 11 51 38 AM
Screen Shot 2022-04-12 at 11 51 52 AM

These errors do not occur with datasets of other types. 

- [ ] Look at the raw JSON representation of the affected datasets.
- [ ] Maybe there's a missing field.

jennyziyi-xu:
Closes #1100

### DIFF
--- a/react/src/components/FieldDisplayEditable.tsx
+++ b/react/src/components/FieldDisplayEditable.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useMemo, useState } from "react";
 import { Box, Fade, makeStyles, MenuItem, TextField, TextFieldProps } from "@material-ui/core";
 import { useAPIInfoContext } from "../contexts";
 import { formatDisplayValue } from "../functions";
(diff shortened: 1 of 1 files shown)


react/src/components/FieldDisplayEditable.tsx
```diff
@@ -87,8 +93,9 @@ function EnhancedTextField({
             textFieldProps.onChange = (e: TextFieldEvent) => {
                 onEdit(field.fieldName, e.target.value === "" ? null : e.target.value);
             };
+
             children = [
-                ...enums[getFieldInEnums(field.fieldName)]?.map((option: string) => (
+                enums[getFieldInEnums(field.fieldName)]?.map((option: string) => (
```
kevinlul, line 98: Dropping the spread operator means we get an array with one element that is another array, are you sure? I don't recall why the `[...thing]` pattern was used here. `.map` does already return an array that isn't a reference to an existing one.

react/src/components/FieldDisplayEditable.tsx
```diff
@@ -50,7 +50,13 @@ function EnhancedTextField({
     onEdit: (fieldName: string, value: boolean | string | null | UnlinkedFile[]) => void;
 }) {
     const [filePrefix, setFilePrefix] = useState<string>("");
-    const enums = useAPIInfoContext()?.enums;
+    const apiInfo = useAPIInfoContext() ?? undefined;
+    var enums;
+    if (apiInfo) {
+        enums = apiInfo.enums;
+        enums["DatasetType"] = Object.keys(apiInfo.dataset_types);
+    }
```
kevinlul, line 58: You could copy over the `hackedTogetherEnums` thing so this is memoized.

kevinlul (changes requested):

kevinlul (approved):
LGTM!

PR #1113, opened 2022-04-18
Bug: fix email not sending error

hannahle:


kevinlul (commented):
Did a SQLAlchemy upgrade change something?

hannahle:
> Did a SQLAlchemy upgrade change something?

From the production logs it seems like it requires you to explicitly cast the type, otherwise it would throw an error. Hence the `text(dataset_id)` in code

flask/app/tasks.py
```diff
@@ -3,22 +3,22 @@
...
         app.logger.info("Sending email notification...")
 
-        yesterday = (datetime.now() - timedelta(1)).strftime("%Y-%m-%d")
-        today = datetime.now().strftime("%Y-%m-%d")
+        yesterday = (datetime.now() - timedelta(5)).strftime("%Y-%m-%d")
 
         analyses = db.session.query(models.Analysis).filter(
             models.Analysis.requested >= yesterday,
-            models.Analysis.requested < today,
         )
```
kevinlul, line 19: This changes "yesterday" to 5 days ago

kevinlul (changes requested):
Looking at this again, there is some weird stuff... so before, only the last analysis wasbeing appended due to indentation? Why are we querying for the datasets per analysis separately anyway, can't we do a join along the `datasets` relationship backref?

kevinlul:
By the way, did you have any trouble analyzing the logs to troubleshoot this?

hannahle:
> By the way, did you have any trouble analyzing the logs to troubleshoot this?

Not really, it was pretty straight forward. I Cmd+F for `send_email_notification` and was able to find the 2 main bugs that way.

flask/app/tasks.py
```diff
@@ -1,50 +1,34 @@
...
-            )
-
-            dynamic_object["analyses"].append(
+            email_object["analyses"].append(
                 {
+                    "analysis_id": analysis.analysis_id,
                     "requested": analysis.requested.strftime("%Y-%m-%d"),
                     "requester": analysis.requester.username,
-                    "pipeline": analysis.pipeline.pipeline_name,
+                    "pipeline": analysis.kind,
```
kevinlul, line 31: Oh you know, this might also be the part that broke it. I must have missed this.

flask/app/tasks.py
```diff
@@ -1,50 +1,34 @@
...
+        datasets = (
+            db.session.query(models.Dataset)
+            .join(models.Dataset.analyses)
+            .filter(models.Analysis.requested >= yesterday)
+            .all()
         )
 
-        app.logger.info("Available analyses...")
-        app.logger.debug(analyses)
+        analyses = list(chain(*[d.analyses for d in datasets]))
```
kevinlul, line 21: I think `chain.from_iterable` will flatten the list of lists without this notation? Like
```python
for analysis in chain.from_iterable(d.analyses for d in datasets):
```
This is only needed if the query is performed on the dataset rather than analysis table though.

flask/app/tasks.py
```diff
@@ -1,50 +1,34 @@
...
 
-        dynamic_object = {"analyses": []}
+        email_object = {"analyses": []}
```
kevinlul, line 23: You can just start with a list here for improved logging and pass `dynamic_template_object={"analyses": analyses}` to the email call at the end.

flask/app/tasks.py
```diff
@@ -1,50 +1,34 @@
...
-
-        yesterday = (datetime.now() - timedelta(5)).strftime("%Y-%m-%d")
-
-        analyses = db.session.query(models.Analysis).filter(
-            models.Analysis.requested >= yesterday,
+        datasets = (
+            db.session.query(models.Dataset)
+            .join(models.Dataset.analyses)
+            .filter(models.Analysis.requested >= yesterday)
+            .all()
```
kevinlul, line 18: You should be able to do the query from the Analysis side like `models.Analysis.query.options(joinedload(models.Analysis.datasets)).filter(...)`

This query does not do what you want as it will filter based on the analysis requested date, but it won't route those analyses to the relationship, so the next line will requery for those analyses for each dataset individually, without regard for the requested date condition above, so you could end up emailing out old analyses.

kevinlul (changes requested):

kevinlul:
I realized another issue when thinking about polling Slurm job status. The BackgroundScheduler is started once per worker process under Gunicorn. We actually need to run it on the master process under Gunicorn, like the Prometheus endpoint.

hannahle:
> https://enqueuezero.com/projects/apscheduler/gunicorn.html

Ahh good catch. Ok I'll make it run on the master process.

flask/app/tasks.py
```diff
@@ -1,84 +1,66 @@
...
-        app.logger.info("Sending email notification...")
-
         yesterday = (datetime.now() - timedelta(1)).strftime("%Y-%m-%d")
-        today = datetime.now().strftime("%Y-%m-%d")
 
-        analyses = db.session.query(models.Analysis).filter(
-            models.Analysis.requested >= yesterday,
-            models.Analysis.requested < today,
+        analyses = (
+            models.Analysis.query.options(joinedload(models.Analysis.datasets))
```
kevinlul, line 13: It looks like later, the dictionary construction is incurring all sorts of additional queries. We could add all the joinedloads for each relationship here.

flask/app/tasks.py
```diff
@@ -13,6 +13,9 @@ def send_email_notification(app):
         yesterday = (datetime.now() - timedelta(1)).strftime("%Y-%m-%d")
         analyses = (
             models.Analysis.query.options(
+                joinedload(models.Analysis.datasets).joinedload(
+                    models.Dataset.linked_files
+                ),
```
kevinlul, line 18: Does this do it with two `joinedload(models.Analysis.datasets)` options. I had imaged that the remaining `joinedload(models.Dataset.tissue_sample)...` would be part of parameters to an `.options` call on this joinedload object along with linked_files. Let me check the final query...

kevinlul: Yup, this looks good to me!

kevinlul (approved):

kevinlul:
Scheduler issue pushed to #1135

PR #1118, opened 2022-04-20
Flask: automate trio exomic analysis

### LINKED ISSUE #1055: The long-coveted Pipeline automation
- PR the change that rips out all database tables from dataset type through to pipeline and see if that's excessive
- take out /api/pipelines and the frontend radio buttons and let the backend decide what pipeline to run

Follow-up to #985

kevinlul:
Fix missing SQLAlchemy log handler in development (disable with `config.py::SQLALCHEMY_LOG = False` if too noisy as in the past)
Remove some extraneous logs that were not informative when developing this

To test against the SickKids HPF, create a dummy pipeline entrypoint in your HPF home directory:
```bash
#!/usr/bin/env bash
set -euo pipefail
echo "$@"
```
Name it `dnaseq_slurm_api.sh` and don't forget to `chmod +x dnaseq_slurm_api.sh` to make it executable.

Accepted job submissions should have standard output captured in your home directory, printing out the passed-in parameters, which should match the format expected by @Madelinehazel: analysis id, family codename, and a JSON of participant codenames to arrays of linked file paths.

Sanitation of inlined parameters in the shell script:
- the analysis id is a MySQL-generated integer. No sanitation needed
- the family codename is a user-input arbitrary string. Use single quotes to prevent shell interpretation and escape any single quotes in the value.
- the remaining arguments are serialized as JSON, which is then put in single quotes to prevent shell interpretation.
If this is still insufficient and passing inline arguments in a shell script is always unsafe, then as noted we should switch the pipeline entrypoint to take environment variables for its parameters. Or we can use the [`argv`](https://slurm.schedmd.com/rest_api.html#v0.0.37_job_properties) job property.

See also ccmbioinfo/crg2#136
Closes #1055


kevinlul:
This is a draft only because I have not implemented the configuration passing to make this immediately live in production.

kevinlul:
Confirmed this works in CHEO-RI with a Python script that reproduces the submission for a real analysis. We will wait until the Slurm cluster is upgraded and provisioned to merge.

PR #1122, opened 2022-04-21
React: fix candidate genes dropdown

### LINKED ISSUE #1075: Add Datasets: candidate genes cannot be selected from dropdown
Regression seems to have been introduced in #1029 as the AutocompleteMultiselect component is also used by CandidateGeneCell

kevinlul:
Fixes #1075

Regression seems to have been introduced in https://github.com/ccmbioinfo/stager/pull/1029 as the AutocompleteMultiselect component is also used by CandidateGeneCell


jennyziyi-xu (approved):
LGTM!

PR #1134, opened 2022-05-02
Flask: set column types explicitly when reading in a csv for bulk upload

### LINKED ISSUE #1124: ERROR [bulk_update, line 330]: {'family_codename': ['Not a valid string.']}
A collaborator is trying to upload some additional datasets and they are being rejected with an error 400: `{'family_codename': ['Not a valid string.']}`.

Server logs show `ERROR [bulk_update, line 330]: {'family_codename': ['Not a valid string.']}`.

I believe this could be a bug with Marshmallow schema validation, as the submitted family codenames all happen to be able to be parsed as whole numbers. I'm not sure how this didn't come up before though, so perhaps #1117 broke something?

delvinso:
- closes #1124

### DIFF
--- a/flask/app/routes.py
+++ b/flask/app/routes.py
@@ -215,7 +215,12 @@ def bulk_update():
         app.logger.debug("Content type is csv, using pandas to read in.")
         try:
             app.logger.debug("Reading in csv and converted to a dictionary..")
-            dat = pd.read_csv(StringIO(request.data.decode("utf-8")))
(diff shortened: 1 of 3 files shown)


delvinso:
> Should we use `object` here or the more specific `str`?

good catch, didn't realize the explicit string dtype was supported now ((https://pandas.pydata.org/docs/user_guide/text.html) !

kevinlul (approved):
LGTM!

PR #1148, opened 2022-06-07
Flask: add /api/analyses/:id/run to retroactively run backlogged analyses

kevinlul:
There is no frontend button for this but I'll call this from the developer console to run the current Care4Rare backlog of ~92 analyses.


delvinso (approved):
LGTM! My previous comment is just a suggestion

PR #1150, opened 2022-06-09
Flask: refactor for clarity, then regularly poll slurm for job status

### LINKED ISSUE #1135: Read job status from Slurm, move BackgroundScheduler to gunicorn master process
The BackgroundScheduler used for sending emails should be moved to the gunicorn master process in production. Otherwise, one scheduler will be started per worker, and they operate independent.

Then, this scheduler can also be used for a task that polls the Slurm jobs API for the state of jobs that Stager submitted, updating the analysis table accordingly. The Slurm API clients should be constructed globally (master and workers), as they create a thread pool and other scaffolding as a full API client, and these resources should be reused instead of being discarded with every request and task.

## See also

https://enqueuezero.com/projects/apscheduler/gunicorn.html
https://apscheduler.readthedocs.io/en/stable/faq.html#how-do-i-share-a-single-job-store-among-one-or-more-worker-processes
https://docs.gunicorn.org/en/stable/settings.html?highlight=on_starting#server-hooks
https://docs.python.org/3/library/multiprocessing.html#multiprocessing.pool.AsyncResult

kevinlul:
Closes #1135


kevinlul:
Tomorrow, this will be broken up into several PRs:
- timezone + configuration refactor
- blueprints subdirectory
- StagerLoginManager
- Stager Flask app subclass
- Mailer class
- Scheduler initialization + poll_slurm implementation

PR #1152, opened 2022-06-09
Flask: move blueprints to directory, routes -> misc, buckets -> unlinked

kevinlul:
Separated from https://github.com/ccmbioinfo/stager/pull/1150

PR #1153, opened 2022-06-10
Flask: refactor app into subclasses of Flask, LoginManager, initialize scheduler in master

kevinlul:
Using classes here helps keep dependencies straight and be clear on initialization order, since we're dealing with specifics of when objects are initialized and in what process.

This assumes #1152 already and has the contents of that PR merged in already, so review that one first!

#1135


kevinlul:
#1152 is merged, rebased onto master

flask/migrations/versions/6f448fc94a2d_many_to_many_dataset_files.py
```diff
@@ -8,7 +8,8 @@
 from alembic import op
 import sqlalchemy as sa
 
-from app.extensions import db
+# NOTE: DO NOT DO THIS, STATIC MIGRATIONS CANNOT DEPEND ON A CHANGING MODEL FILE
```
frewmack, line 11: I take it that this will be left as a future issue?

frewmack (approved):
LGTM!

flask/migrations/versions/6f448fc94a2d_many_to_many_dataset_files.py
kevinlul: Yeah, I actually noticed this way back when I removed the metadatasets table and it broke a migration that was added while I was away. It's a _fait accompli_ at this point, but we do need to keep note of this as in development environments, we are running all the migrations from the very first one, instead of just any new ones that were added.

PR #1154, opened 2022-06-10
Flask: poll Slurm API for job status to update analysis status

### LINKED ISSUE #1135: Read job status from Slurm, move BackgroundScheduler to gunicorn master process

kevinlul:
Initialize Slurm clients at top-level instead of per request to retain the thread pool and avoid the reconstruction cost. Threads are not spawned until first request, so they will be kept per process instead of disappearing across forks.

Closes #1135


kevinlul:
Rebased after #1153 merged

frewmack (approved):
Looks good to me!

kevinlul:
Thanks! 🙏

PR #1157, opened 2022-06-20
Flask: automatically set analysis result path when run

kevinlul:
Add family and participant codename check constraints to prevent illegal filenames and arbitrary paths. Participant codenames allow periods because some old ones have them, but no family codename has them.


flask/app/models.py
```diff
@@ -106,7 +111,14 @@ class ParticipantType(str, Enum):
 class Participant(db.Model):
     participant_id: int = db.Column(db.Integer, primary_key=True)
     family_id = db.Column(db.Integer, db.ForeignKey("family.family_id"), nullable=False)
-    participant_codename: str = db.Column(db.String(50), nullable=False, unique=True)
+    participant_codename: str = db.Column(
+        db.String(50),
+        CheckConstraint(
+            "participant_codename REGEXP '^[.,[:alnum:],-,_]+$' AND participant_codename NOT REGEXP '^[.]+$'"
```
frewmack, line 117: Need to escape the minus sign in order to match for it in a bracket expression. Also unless we're matching for commas, the characters don't need to be comma-delimited.
```suggestion
            r"participant_codename REGEXP '^[.[:alnum:]\-_]+$' AND participant_codename NOT REGEXP '^[.]+$'"
```

flask/app/models.py
```diff
@@ -70,7 +70,12 @@ class Group(db.Model):
 @dataclass
 class Family(db.Model):
     family_id: int = db.Column(db.Integer, primary_key=True)
-    family_codename: str = db.Column(db.String(50), nullable=False, unique=True)
+    family_codename: str = db.Column(
+        db.String(50),
+        CheckConstraint("family_codename REGEXP '^[[:alnum:],-,_]+$'"),
```
frewmack, line 75:
```suggestion
        CheckConstraint(r"family_codename REGEXP '^[[:alnum:]\-_]+$'"),
```

flask/migrations/versions/690e68b426ca_add_codename_check_constraints.py
```diff
@@ -0,0 +1,29 @@
...
+# revision identifiers, used by Alembic.
+revision = "690e68b426ca"
+down_revision = "c137bdb34421"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.execute(
+        "ALTER TABLE family ADD CONSTRAINT family_chk_codename CHECK (family_codename REGEXP '^[[:alnum:],-,_]+$')"
```
frewmack, line 20:
```suggestion
        r"ALTER TABLE family ADD CONSTRAINT family_chk_codename CHECK (family_codename REGEXP '^[[:alnum:]\-_]+$')"
```

flask/migrations/versions/690e68b426ca_add_codename_check_constraints.py
```diff
@@ -0,0 +1,29 @@
...
+    )
+    op.execute(
+        "ALTER TABLE participant ADD CONSTRAINT participant_chk_codename CHECK (participant_codename REGEXP '^[.,[:alnum:],-,_]+$' AND participant_codename NOT REGEXP '^[.]+$')"
```
frewmack, line 23:
```suggestion
        r"ALTER TABLE participant ADD CONSTRAINT participant_chk_codename CHECK (participant_codename REGEXP '^[.[:alnum:]\-_]+$' AND participant_codename NOT REGEXP '^[.]+$')"
```

frewmack (changes requested):

PR #1159, opened 2022-06-23
Docker: bump MinIO for security

kevinlul:
Three high-severity vulnerabilities with real consequences:

- CVE-2021-43858  / GHSA-j6jc-jqqc-p6cx
- CVE-2022-24842 / GHSA-2j69-jjmg-534q
- CVE-2022-31028 / GHSA-qrpr-r3pw-f636

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Production image. Runs a Gunicorn WSGI server.
-FROM minio/mc:RELEASE.2021-12-10T00-14-28Z AS mc
+FROM minio/mc:RELEASE.2022-06-17T02-52-50Z AS mc
 FROM python:3.9-slim
 ARG GIT_SHA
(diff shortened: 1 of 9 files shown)


frewmack (approved):
LGTM! 🚀

kevinlul:
I've confirmed that our use case still works, especially moving files in on the backend after the server has started, but please note that now MinIO will break if you try to start it with existing content in its directory that did not come from a prior MinIO version (missing `.minio.sys`): https://docs.min.io/minio/baremetal/installation/deploy-minio-single-node-single-drive.html#minio-snsd-pre-existing-data

The workaround is to either move the files in after the server has started on an empty directory, or start a pre-2022-06-02 version of MinIO on the directory with existing contents and then upgrade to a post-2022-06-02 version of MinIO.

PR #1203, opened 2024-12-02
Set CHEO worker timeout to 10 min from 30 sec

frewmack:
Should resolve a bug with large downloads timing out on the CHEO instance


frewmack:
Github Actions uses a new version of Docker where "docker-compose" is no longer a valid command, and "docker compose" is. Our VM uses "docker-compose", so updating the script to pass the validate action will cause the script to fail in our VMs. 

We need to revisit this issue in the future by either 
- updating our VMs to a new version of Docker that uses "docker compose" and updating the scripts accordingly, or
- updating the Github Actions to install the specific version of Docker that we use in our VMs.

